PR #7, opened 2018-08-01
Content for Ecosystem Section

ranjanan:


ranjanan:
cc: @ViralBShah

ChrisRackauckas:
Oh, and the Lorenz animation would be great for the scientific computing tab. The visualization tab should get something more data visualization related.

ViralBShah:
Yes, everything is pretty haphazardly chosen. I will make one more pass over it. There will be no significant changes before it goes live except for a few changes for choices of organizations and text edits.

The website will go live after 0.7, but before 1.0.

ViralBShah:
The main thing is that all the tabs need to be as inclusive as can be without cluttering things, and advertise well maintained packages.

PR #8, opened 2018-08-01
Modernize example code

ararslan:
The code as written doesn't work at all on 0.7 and is deprecated on 0.6.


ararslan:
This code seems to have been removed from the site.

PR #13, opened 2018-08-01
Add a new research tab

ranjanan:
This will have the all the publications and potentially another other interesting research done on and in Julia


ranjanan:
Again, I don't know how to get the alignment right

ranjanan:
Managed to add the new styling to it too

PR #36, opened 2018-08-03
Fix <p> tag that was breaking homepage

djsegal:
This was messing up your whole landing page

// shakin my head :P  

screen shot 2018-08-02 at 10 48 49 pm

----

(also, the `Donate` button in the footer is funky on mobile. but this redesigns pretty sweet!)

### DIFF
--- a/_includes/jumbotron.html
+++ b/_includes/jumbotron.html
@@ -26,7 +26,7 @@
         </br/>
         </div>
         <div class="col-3 col-md-1 col-sm-1"></div>
-      </p>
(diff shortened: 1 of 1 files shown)


djsegal:
To prevent this kind of thing happening in the future, you should probably add a testing suite that takes pictures of the pages and does a spot check?

ViralBShah:
Not sure I can tell the difference.

djsegal:
Everything is inside the jumbotron on the left

Issue #18, opened 2018-08-02
Font is too bold


iamed2:
![too bold](https://user-images.githubusercontent.com/3131001/43597299-51ee1312-9647-11e8-9bc7-bdff88a5b5e7.png)

The bold font weight pushes letters close together and makes them harder to distinguish. Perhaps "Futura LT Medium" would be a better choice than "Futura LT Bold"?

ViralBShah:
PRs appreciated.

ViralBShah:
@essenciary I often felt that all the bold fonts could be made non bold at least to try things out. Thoughts?

PR #40, opened 2018-08-03
Change bold weight font to medium weight font

iamed2:
Closes https://github.com/JuliaLang/www.julialang.org/issues/18

I made the changes live in Firefox and checked it out and I think it looks good.


ViralBShah:
Let's merge and find out. :-)

ViralBShah:
Does look nice.

PR #52, opened 2018-08-03
Change ecosystem image for ML

Evizero:
changes the static image in the ecosystems tab for ML to this gif:

rnd_cartpole

EDIT: a side note: i created it with julia and Plots.jl a while back


Evizero:
it distorts the aspect ratio for some reason, which didn't happen when I edited and tried it locally. unsure what is going on

ViralBShah:
Let's leave it like this for a bit and see if someone knows what to do. @essenciary or @djsegal ?

essenciary:
Yes, I noticed that too. 

It's because the inline style of the image I put as the example was copied without updating the dimensions of the new image. 

Sidenote: we can use only one of the 2 dimensions (only the width, if we want all the tabs to have the same image width; or just the height, if we want to constrain/control the hight of the tabs). The other will be kept proportional. 

image

PR #56, opened 2018-08-03
Add ecosystem pulse button to Packages section

ranjanan:
image


ViralBShah:
It would be nice if the Ecosystem pulse and JuliaObserver could all become one site and the new pkg.julialang.or. Probably a lot of work. 

cc @djsegal

PR #57, opened 2018-08-03
Update Diffeq GSoCs

ChrisRackauckas:
A lot of the newer changes were lost in the transition. I just want to make sure this is kept up-to-date so for the next GSoC it's easy to get started. 

BTW, we should discuss when we plan on opening up the GSoC tag again. Students last year started messaging in early December about projects.


ViralBShah:
Well, then we should open it up in December.

Issue #54, opened 2018-08-03
Main page is very large (~17MB)


jebej:
Here is a snip from the chrome web tools, sorted by size:

![image](https://user-images.githubusercontent.com/11141558/43644351-6da73790-96fc-11e8-88ce-02e3a02f6518.png)

jebej:
Maybe some of these plots could be re-rendered with Plotly or similar.

ViralBShah:
Maybe we should just use static images at least on mobiles?

ViralBShah:
I also don't know how this affects our bandwidth bill with the CDN. I assume much of this just gets cached in the user's browser.

mbauman:
What if we deferred loading the gifs until the tab is clicked?  We could even start with a static image and then dynamically swap it out once the gif arrives.  That should also help mitigate the bandwidth bill.

mbauman:
Well after #60 we're down to 6.2MB, of which half is waves.gif.  Better, but still probably worth investigating deferred downloading or perhaps a more gif-optimization-friendly demo for the visualization tab.

ViralBShah:
Or pick a different thing to show there. I just picked what I found on the Plots.jl pages.

ranjanan:
Has #71 helped?

KristofferC:
It is dominated by waves.gif but #72 at least took away a MB from it

mbauman:
Here's after #72:

![screen shot 2018-08-03 at 5 19 59 pm](https://user-images.githubusercontent.com/154641/43668255-89db6946-9741-11e8-97d9-6c986ff55f90.png)

Why am I downloading some of the same scripts/css files three times?

KristofferC:
17MB -> 5MB in 9 hours is good progress :)

mbauman:
Ah, the triplets are coming from the three embedded YouTube videos.

mbauman:
This looks promising to shave off ~1.5MB: https://www.labnol.org/internet/light-youtube-embeds/27941/

ranjanan:
Nice. Will make a PR and test it

essenciary:
I haven't tested it, but in theory, we could convert waves.gif to a video using something like http://gif-2-mp4.com

ViralBShah:
I think we are now at 5MB. So probably ok to close this.

KristofferC:
Emulating a slow 3g connection the website takes over 1 minute to load.

Fast 3g is 20 seconds.

[...]

PR #60, opened 2018-08-03
Reduce GIF filesizes

### LINKED ISSUE #19: Text on data science plot is rendered poorly for the size of image
This is what it looks like on my computer:

data science image screenshot

I think perhaps if the image was re-rendered at the size it will be displayed at, the text would be clearer?

mbauman:
Resize all ecosystem gifs to 400px -- the size they appear on the homepage. Also optimize their transparency between frames. Fixes #19 by avoiding some web-browsers' shoddy resize methods, mitigates #54.

Before:

screen shot 2018-08-03 at 11 02 31 am

After:

screen shot 2018-08-03 at 11 02 07 am


ViralBShah:
Not sure if this was supposed to fix it, but the ML gif is still stretching the page. Perhaps the gifs need to be constrained by height instead of width.

PR #65, opened 2018-08-03
Add vim, emacs and sublime to IDE list 

ranjanan:


ViralBShah:
Can we remove sublime?

KristofferC:
I don't think we should include the sublime plugin since it ihas beem unmaitained for many years.

ranjanan:
No, it's as well if not better maintained than julia-vim.

ViralBShah:
I thought @quinnj said he doesn't maintain it at all. The emacs plugin has many users and maintainers.

ViralBShah:
Let's see how this looks.

ViralBShah:
I see we need a 6th one for symmetry.

ViralBShah:
@mbauman Can we have your magic here to reduce file sizes?

PR #66, opened 2018-08-03
Shrink IDE icon sizes down, fixup aspect ratios

mbauman:


ranjanan:
Thanks! :)

ViralBShah:
Things don't looks all the same height now. The Vi looks bigger.

mbauman:
Probably constant height would be better.

PR #67, opened 2018-08-03
Add Julia 0.7.0-rc2 binaries

ararslan:


ViralBShah:
Probably needs updating in the footer too.

ViralBShah:
Maybe we should not mention versions in the footer.

ararslan:
I did update the footer, actually. Though it's not entirely obvious that a specific version would be listed there, so I agree that it may be worthwhile not to list versions there.

PR #70, opened 2018-08-03
Add all orgs from previous site

### LINKED ISSUE #34: Big List of GithubOrgs, links to chats etc
New site is pretty.

Not quiet 100% of information from
https://github.com/JuliaLang/www_old.julialang.org/blob/master/ecosystems/index.md
has made it across.

For example that big list had links to all the Chatrooms, forums, facebooks, etc etc  for all the localization efforts.
New localization page only links to github orgs, or websites.
Most of those websites do then have links to the other communication channels so maybe that is OK.
On the other hand @ViralBShah's [comment here](https://discourse.julialang.org/t/chinese-discourse-is-set-up/12829/2?u=oxinabox)
about making a PR to this repo, to link to the Chinese discourse, is not possible.
(or at least would look out of place without the others.)

The other thing the ecosystems page had was a big list of all the orgs,
and all there chat rooms etc.
Some of that has made it into the new ecosystems pages.
But not all of it.

The github orgs are a big way of guiding people to packages -- it is a proxy for stability and compatibility.
Having that would help address this concern [raised on discourse](https://discourse.julialang.org/t/new-website/12844/4?u=oxinabox)

On the other hand the orgs list does become out-dated, chatrooms can become disused etc.
And the huge list is huge.

I still think it has some value that we want preserve.

Maybe there could be a "More Info" link behind the new fancy pretty ecosystems page?

ranjanan:
Closes #34 

This PR adds all orgs from the previous website. I guess that's the information that didn't make it to the new site. I could add more maybe. 
Suggestions, @oxinabox, @ViralBShah?

### DIFF
--- a/community/index.html
+++ b/community/index.html
@@ -126,6 +126,7 @@ title:  Julia Community
           <li><a href="https://github.com/JuliaLangEs">JuliaLangEs</a> (Spanish)
           <li><a href="https://github.com/JuliaLangPt">JuliaLangPt</a> (Portuguese)
           <li><a href="https://juliakorea.github.io/ko/">JuliaKorea</a> (Korean)
+          <li><a href="https://github.com/JuliaLangGerman">JuliaGerman</a> (German)
         </ul>
       </p>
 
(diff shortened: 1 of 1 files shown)


ViralBShah:
Not sure this is the right place, but good starting point.

oxinabox:
Community is where the list was originally. Before we decided it was too long and moved it to ecosystems.
Community has dome advantages because it localises all the chat links

PR #72, opened 2018-08-03
optimize some gifs and lossy compress waves.gif

KristofferC:
All gifs losslessly compressed except waves.gif which I used LossyGif to compress from 3MB to 2MB. I can't see a difference just looking at the gif.


mbauman:
3MB down to 2MB on waves.gif is a nice win.

ViralBShah:
Nice

PR #78, opened 2018-08-04
add github ribbon; make download green

Evizero:
- changes the download button to green
- adds a red github ribbon linking to the julia repo

Before | After
---|---
old | new

Note: I know little about jekyll, so I just winged it concerning how i included the ribbon in its own file and such (monkey see, monkey do). It works with and without the alert though, so thats something

### DIFF
--- a/_includes/jumbotron.html
+++ b/_includes/jumbotron.html
@@ -11,8 +11,9 @@
       <div class="row">
         <div class="col-lg-3 col-md-1 col-sm-1"></div>
         <div class="col-lg-3 col-md-5 col-sm-5">
-          <a class="btn btn-danger btn-lg btn-block" href="/downloads" role="button">Download {{ page.julia_version }}</a>
-          <a class="link extra-link" href="/benchmarks">&nbsp;Benchmarks</a>
+          <a class="btn btn-success btn-lg btn-block" href="/downloads" role="button">Download {{ page.julia_version }}</a>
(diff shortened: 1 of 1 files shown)


ViralBShah:
I am not a big fan of that ribbon. It distracts. I would rather just say it in the footer.

I also do like the current colors. Any particular reason for making it green?

Evizero:
just a suggestion. Really, I'd say it boils down to differences in taste at this point. If I had to come up with a logical reason I'd say that a red button isn't as inviting as a green one (the class is called `btn-danger` after all).

Either way, I don't have a strong opinion on this

ViralBShah:
Yes, Red does suggest to stop, whereas green to move forward - in almost all cultures too. But I love it how the red jumps out of the page and catches your attention. Let's leave this open and see how we feel. But let's not do the ribbon.

mbauman:
I like the green, but not the ribbon.

How about instead, we add another link below the download button (aside "Benchmarks") for "Source" that goes to the GitHub repo?  That'd add a bit more symmetry with the documentation button.  It'd also make it a bit more obvious that Julia is open source.

Evizero:
screen shot 2018-08-06 at 11 11 40

KristofferC:
FWIW, I like the green more, it is a bit more inviting and calm :)

Evizero:
(i renamed the commit message to reflect the update)

PR #79, opened 2018-08-04
RFC: use three buttons in jumbotron

Evizero:
So this is more of a inspiration/discussion than a PR I think. Essentially I tried to do three things

- make the download button the primary eye catcher
- make the Benchmarks more discoverable
- mimic the Julia logo with the button layout

screen shot 2018-08-04 at 13 58 46

Thoughts?

### DIFF
--- a/_includes/jumbotron.html
+++ b/_includes/jumbotron.html
@@ -9,17 +9,24 @@
         Julia is a highly productive language that runs fast
       </p>
       <div class="row">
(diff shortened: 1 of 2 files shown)


ViralBShah:
I like the idea but it doesn't look nice. I am going to close this one.

Evizero:
Well my main hope here was that someone with actual webdev skills looks at this and goes: "oh I like the idea and I can actually make it look nice"

Evizero:
That said I share your opinion

PR #82, opened 2018-08-06
Update packages.html

ahclaster:


ViralBShah:
Just checked that actual number of downloads is 2.3M (includes all CI and attempted downloads).

PR #86, opened 2018-08-06
replace `---` with the correct en-dash

vtjnash:


ararslan:
Shouldn't that be — rather than –?

ViralBShah:
I am not an expert on dashes, but pasting @ararslan's dash into the PR did not change how it looked. If the dash overlords want to correct this further, I won't stop them.

vtjnash:
Alex is right, "work—and" might be the more common style, over "work – and" or "work — and" (
http://en.wikipedia.org/wiki/Dash#En_dash_versus_em_dash)

ararslan:
Jameson is dash overlord confirmed

PR #91, opened 2018-08-07
Make column smaller for blog posts on large screens

haampie:
Blog posts have long lines on large screens, making it very unpleasant to read imo.

This PR reduces the width to 2/3 of what it was and centers the column.

Before
Before

After
After

### DIFF
--- a/_layouts/post.html
+++ b/_layouts/post.html
@@ -9,7 +9,7 @@ layout: homepage
 <div class="container">
 
   <div class="row">
-    <div class="col-12">
+    <div class="col-12 col-lg-8 offset-lg-2">
       <h1>{{ title }}</h1>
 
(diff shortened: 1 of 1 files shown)


ViralBShah:
I believe it used to be like that and we made it wider (but can't be sure I remember the details accurately). I guess people can always make their browser windows narrower.

KristofferC:
> I guess people can always make their browser windows narrower.

That seems kinda sucky :(

(FWIW the JuliaComputing blog seems to have reasonable width)

haampie:
But is there an argument for the wider layout? An argument against is that I don't have to resize my window ^^.

ViralBShah:
Let's merge and see if we get complaints.

PR #107, opened 2018-08-09
Update the website for 1.0

ararslan:
Changes include:

* Setting download links on downloads/index.html to point to 1.0.0
* Moving 0.7.0 to the older releases page
* Setting the big button on the front page to v1.0
* Changing the footer to point to 1.0 and "older" rather than "Julia v"
  and "Julia v0.7.0-rc2"


Keno:
Merging, we can do fine tuning afterwards.

Issue #102, opened 2018-08-08
Blog Formatting


MikeInnes:
Has come out pretty badly, [e.g.](https://julialang.org/blog/2017/12/ml&pl).

A bunch of these posts still get a lot of views, so it'd be worth having them look nice.

mbauman:
We just need to move some of the CSS from the old 
[css/screen.css](https://github.com/JuliaLang/www.julialang.org/blob/master/css/screen.css) (which is no longer included) to the new [v2/css/app.css](https://github.com/JuliaLang/www.julialang.org/blob/master/v2/css/app.css).   Let's also delete the old one once we make that transition because it's confusing to folks who don't know which is the right one :)

ViralBShah:
We need a PR.

jekbradbury:
We also need `syntax.css` I think?

PR #116, opened 2018-08-09
Minimal blog layout fixes

jekbradbury:
Fixes the giant images and poorly-formatted code (#102); the other big thing still missing from blog formatting is `syntax.css` (syntax highlighting). I'm not sure why the img max-width was commented out before, so this might break something else.

### DIFF
--- a/v2/css/app.css
+++ b/v2/css/app.css
@@ -403,7 +403,7 @@ img.ecosystem-image {
 }
 
 img {
-  /* max-width: 100%; */
+  max-width: 100%;
 }
 
 .page-heading .main-heading {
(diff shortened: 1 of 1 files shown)


ViralBShah:
Should we merge this?

ViralBShah:
Or does it need to wait on the other fixes?

jekbradbury:
I think it should be OK to merge, but somebody presumably commented out the `max-width` intentionally, so there’s a chance it breaks another part of the site.

ViralBShah:
@essenciary Could you review this?

essenciary:
Indeed, I think it was me commenting out `max-width`. The selector was too generic and was breaking something else. I suggest defining it to be more specific, maybe `.blog.img`.

ViralBShah:
Maybe can be improved later. Merging and seeing if this fixes some existing issues.

PR #123, opened 2018-08-10
Create 2018-08-08-one-point-zero-es.md

miguelraz:


ViralBShah:
I think instead of linking these as separate posts, perhaps we should link the translations from the english version? That might also increase the discoverability.

PR #139, opened 2018-08-14
New benchmark plot for julia-1.0.0

johnfgibson:
New microbenchmark results for julia-1.0.0. 

I added language versions in text following plot, since the HTML table no longer appears. I think the table should be put back in, as it's clearer to read specific timings from that than by staring at the colored dots. But inclusion of _includes/benchmarks.html from benchmarks/index.md looks pretty terrible in the new web framework. I'lll file an issue on that.

Additional language Rust is thanks to @Enet4. 

@ViralBShah: changing Julia's optimization level to -O 3 makes no discernable difference in any of the timings, changes in only the third digit. 

Geometric mean of Julia is 1.17 relative to C, LuaJIT is 1.09, Rust is 1.10. Those are degradations from roughly 1.05 in all three cases from the runs for julia-0.6, IIRC. There is clearly a regression in the Julia parse_int, now nearly a factor of 2 slower than C versus 1.2 or 1.3 for julia-0.6. But the relative changes in LuaJIT and Rust suggest some of Julia's relative performance regression is due to better C compilation by gcc-7.3.1 than gcc-4.3.5. If it helps to have hard numbers on this I'll do a head-to-head comparison of julia-0.6.4 and 1.0.0.


ViralBShah:
Good to merge?

johnfgibson:
Yes, good to merge, IMHO.

PR #146, opened 2018-08-18
Add ThinkJulia

cormullion:
Don't know if this was lost in the move or removed deliberately?


ViralBShah:
I dropped a bunch at some point that I wasn't sure were updated.

PR #147, opened 2018-08-18
Fix link of Julia 1.0 zh_tw

Roger-luo:


Roger-luo:
Oops, I mixed with another commit

PR #158, opened 2018-08-20
Rename 2017-12-20-ml&pl-cn.md to 2017-12-20-ml-pl-zh_cn.md

annainfo:
1) IANA language tag for simplified Chinese (https://tools.ietf.org/html/bcp47 ), based on ISO 639-1 and ISO 3166-1 is zh-CN.

2) Replace "&" which is a RFC 3986 reserved character (https://tools.ietf.org/html/rfc3986#section-2.2 ) and RFC 3986 percent encoding (https://tools.ietf.org/html/rfc3986#section-2.1) can cause problems in communication, e.g. broken hyperlinks.


ViralBShah:
We won't be breaking these URLs. If you really want to do this, please also provide a redirect for the old one.

annainfo:
Happy to provide redirects. How are they made? Yesterday you "broke" a URL https://github.com/JuliaLang/www.julialang.org/pull/151 - probably I could understand how redirects are made if you point me to the redirect for that one.

Evizero:
> Yesterday you "broke" a URL #151 - probably I could understand how redirects are made if you point me to the redirect for that one.

I don't think that PR you link to broke any links.

annainfo:
> I don't think that PR you link to broke any links.
Can you provide verifiable information that explains why you think so?

Evizero:
The permalink that jekyll created for the blogpost remained the same before and after the PR. Notice how the "day" portion of the date is not configured to be part of the perma link jekyl creates (this can be set in the _config.yml, see https://github.com/JuliaLang/www.julialang.org/blob/master/_config.yml#L1). Another simple check that the link still works, is to just click on the link that is included in the description of the referenced PR.

You know, you made the confident claim that the referenced PR broke a link also without any supporting evidence. Yet you set a high standard for anyone voicing doubt about your unproven claims. There is a certain tone and double standard here that is not inviting collaboration.

annainfo:
Ignoring comments on users, I am find with providing evidence
https://github.com/JuliaLang/www.julialang.org/blob/master/blog/_posts/2017-09-1-gsoc-fenics.md
returns a page with a 404 label for me.

annainfo:
Whatever, could someone address the issue and explain how redirects are created?

StefanKarpinski:
If you grep for the word `redirect` in this repo there a bunch of examples of redirects.

PR #166, opened 2018-08-23
Update doc-link in footer from stable to v1.

fredrikekre:
ref https://github.com/JuliaLang/julia/pull/28792 and https://github.com/JuliaLang/julia/issues/26825

cc @mortenpi


fredrikekre:
Will merge tomorrow unless someone reports that https://docs.julialang.org/en/v1/ does not work as expecte.

PR #167, opened 2018-08-24
No politics in Julia please.

SalchiPapa:
Reference:

* https://julialang.slack.com/archives/C680MM7D4/p1535088776000100

> Please remember folks that this Slack is for Julia. Side discussions about this or that are fine and natural in any real-time discussion
> venue, but some topics are just flatly out of scope because, like it or not, they can be polarizing and end up alienating people, which we
> absolutely do not want, regardless of the reason.

> People have their own beliefs about things, which they're entitled to, and it should be irrelevant here. We're here for a common interest
> in Julia.

> Please direct discussions about such matters to non-Julia forums

@ararslan

### DIFF
--- a/community/standards/index.html
+++ b/community/standards/index.html
@@ -19,7 +19,7 @@ title:  Julia Community
       <h3>Be respectful and inclusive</h3>
       <p>
         Please do not use overtly sexual language or imagery, and do not attack anyone based on any aspect of personal identity,
-        including gender, sexuality, religion, ethnicity, race, age or ability. Keep in mind that what you write in public forums is
+        including gender, sexuality, politics, religion, ethnicity, race, age, ability, etc. Keep in mind that what you write in public forums is
         read by many people who don't know you personally, so please refrain from making prejudiced or sexual jokes and comments –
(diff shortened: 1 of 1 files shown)


community/standards/index.html
```diff
@@ -19,7 +19,7 @@
       <h3>Be respectful and inclusive</h3>
       <p>
         Please do not use overtly sexual language or imagery, and do not attack anyone based on any aspect of personal identity,
-        including gender, sexuality, religion, ethnicity, race, age or ability. Keep in mind that what you write in public forums is
+        including gender, sexuality, politics, religion, ethnicity, race, age, ability, etc. Keep in mind that what you write in public forums is
```
ararslan: "etc" doesn't seem quite direct enough, IMO; I'd say "or anything else." Does that seem reasonable?

ararslan:
Seems like a good change to me, but since this is a change to the language in the community standards, I'll let a steward review and merge.

ararslan:
Thanks Viral, and thanks Ismael for taking the initiative to add this,

PR #173, opened 2018-08-29
Fix usage: enthusiastic vs enthused

apjanke:
Minor prose usage fix: "enthused" is the past-tense verb form of "enthuse"; "enthusiastic" is the adjective.


ararslan:
Thanks, this is better.

PR #183, opened 2018-09-19
Update index.md

olavmars:
Sorry, was a mix up with following entry...
is this update now correct?
The review looks perfect with link and next entry working....
Regards
Olav


olamar6:
Hi ViralBShah, still not correct, sorry, I think i updated index 12 days ago..?!

PR #200, opened 2018-10-09
added link to a past Intro to Julia

VictorelPetrovich:
to Intro to Julia, by Jane Herriman (August 31, 2018)

### DIFF
--- a/learning/index.html
+++ b/learning/index.html
@@ -38,7 +38,7 @@ title:  Learning Julia
         <h3>Past </h3>
 
         <ul>
-            <li>Intro to Julia, by Jane Herriman (August 31, 2018) </li>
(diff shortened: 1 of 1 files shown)


xorJane:
Hey @VictorelPetrovich, thanks for the edit! As Viral mentioned, we intentionally don't include links for the repeat tutorials where content hasn't changed much. Our next tutorial on October 26th will be the first intro tutorial on 1.0, and so we'll include a link for that. :)

xorJane:
>Or else, maybe you should just delete those items of past tutorials without links (what's their point?)

That's a valid question. In my mind, the point is to show that tutorials are regularly occurring, but we may simply be creating visual clutter without adding real content. @ahclaster Do you see any reason to keep a history of all past intro to Julia tutorials posted @ https://julialang.org/learning/ ?

PR #225, opened 2018-11-12
Add GSoC Link

ChrisRackauckas:


aviks:
`/soc` redirects to `https://docs.julialang.org/en/latest/soc/` and then throws a `404`

ChrisRackauckas:
What's the correct link?

ChrisRackauckas:
Found it.

PR #235, opened 2018-11-28
Update images.md

zygmuntszpak:
Adds stereo matching project description

### DIFF
--- a/soc/projects/images.md
+++ b/soc/projects/images.md
@@ -28,3 +28,20 @@ OpenCV is one of the pre-eminent image-processing frameworks, and there is an ex
 **Expected Results:** A rejuvinated OpenCV package that integrates more effectively into the ecosystem of JuliaImages.
 
 **Mentors:** [Maximiliano Suster](https://github.com/maxruby) and [Tim Holy](https://github.com/timholy/).
+
+## Contributions to a Stereo Matching Package
+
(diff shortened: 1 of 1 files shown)


soc/projects/images.md
```diff
@@ -28,3 +28,20 @@ OpenCV is one of the pre-eminent image-processing frameworks, and there is an ex
...
+
+This project will implement several stereo matching algorithms. Emphasis will be placed on *efficient* implementations which leverage all of Julia's features for writing fast code. 
+
+Example algorithms: 
+  1. Bleyer, Michael, Christoph Rhemann, and Carsten Rother. "PatchMatch Stereo-Stereo Matching with Slanted Support Windows." Bmvc. Vol. 11. 2011.
+  2. Hirschmuller, Heiko. "Accurate and efficient stereo processing by semi-global matching and mutual information." Computer Vision and Pattern Recognition, 2005. CVPR 2005. IEEE Computer Society Conference on. Vol. 2. IEEE, 2005.
+  3. Gehrig, Stefan K., and Clemens Rabe. "Real-time semi-global matching on the CPU." Computer Vision and Pattern Recognition Workshops (CVPRW), 2010 IEEE Computer Society Conference on. IEEE, 2010.
+
+**Recommended skills:** Sufficient familiarity with the Julia language to be able to write *performant* code in Julia.
```
ChrisRackauckas: Writing performant code can't be learned on the job?

zygmuntszpak: My intention is to signal that it would be great if the participant already had some familiarity with typical performance traps and that they are aware of how to benchmark code, reduce allocations, use StaticArrays and watch out for type-instability etc. 

Would you like me to rephrase the statement to something along the lines I have written in this response?

ChrisRackauckas: Yup. I think it's good to be clear that you want them to know. Saying that they need to know how to write performant code seemed like it would turn away some potential applicants who would still be qualified, or at least it wasn't clear how deep into code performance they should already be familiar with.

PR #237, opened 2018-11-28
Shameless self-promotion

timholy:
This adds two projects I've been involved with to the home page. The first, JuliaImages, seems pretty clear: by number of stars, development activity, and number of mentions on discourse it seems to be at least as central as quite a few of the organizations that are already listed in the "Ecosystem" section. (I'm not quite sure why it wasn't included already, but maybe it's because no one ever submitted a PR.) I included a link to the docs rather than the GitHub page.

The second is to add Revise to the development tools. I know Gallium was once listed, so there is precedent for non-editors, and overall I think quite a few people find Revise to be a tool that is worth knowing about (but hard to learn about unless you use discourse).

I am not sure I've added these correctly, but they look OK locally (the Revise icon seems a little on the big size, but not sure how to tweak that).


ViralBShah:
Merging. Let's fix things if they look a bit off...

Issue #135, opened 2018-08-13
Add important descriptors on www.julialang.org


xorJane:
Under "Julia in a nutshell", we say that Julia is fast, dynamic, optionally typed, general, technical, and composable.

Two important descriptors we should add and highlight are "easy  to learn" and "open source". These sections might read

> **Easy to learn**
Julia has high level syntax, making it an accessible language for programmers from any background or experience level.

>  **Open source**
Julia is free for everyone to use, and all source code is publicly viewable on GitHub.

@ahclaster, does that wording sound reasonable?

ararslan:
I'd reword the former a bit:
> **Easy to learn**
> Julia has high level syntax, making it an accessible language for programmers from any background or experience level.

Also, I don't find "you can avoid licensing" particularly clear.

xorJane:
I like your reword and comment! Updated above. :)

ViralBShah:
May need a bit of a redesign also - maybe 4x2?

ViralBShah:
I think "Technical" and "Composable" can be replaced with "Easy to Learn" and "Open Source".

PR #240, opened 2018-12-03
Update descriptors on main page

xorJane:
Add "Easy to use" and "Open source" as descriptors for Julia. As suggested by @ViralBShah in issue #135, use these new descriptors to replace "Technical" and "Composable".

Thanks!


ViralBShah:
Merging now. We can keep wordsmithing further. Thanks @xorJane

PR #242, opened 2018-12-05
Add more JuliaImages projects

timholy:
CC @zygmuntszpak; I didn't want to assign myself as a potential mentor for too many projects, so I just deleted the "mentor" section. Let me know if you object.

### DIFF
--- a/soc/projects/images.md
+++ b/soc/projects/images.md
@@ -7,41 +7,43 @@ title:  JuliaImages Projects – Summer of Code
 
 {% include toc.html %}
 
-[JuliaImages](https://github.com/JuliaImages) has recently undergone a significant revamp, with many improvements in the core infrastructure, project organization, and documentation. For the future, a major push is on expanding the functionality. Please see the [documentation](http://juliaimages.github.io/latest/) for information about the current state; perhaps the [feature comparison matrix](http://juliaimages.github.io/latest/api_comparison.html) may be particularly useful for generating further ideas about possible ways to contribute.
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
We need to have somebody listed as the mentor.

timholy:
Would it be reasonable to just list a group of people for all the projects, or do you need a project-by-project breakdown?

ChrisRackauckas:
Listing a group of possible mentors for imaging projects? Yeah, that seems reasonable to me. We just need to make sure that we (and the students) know who to contact when someone shows interest. If it's any/all people from the list, that's fine.

zygmuntszpak:
I was using the mentors section to signal what I am willing/able to mentor.

I don't want to participate on anything related to opencv or opencv integration. I also have no experience with GUI toolkits/GLK so, while I think creating something like an ImageJ GUI would be fantastic, I won't have time during the GSOC to learn enough to stay ahead of the student and be helpful debugging compilation issues etc. 

I've put up a project that I can certainly mentor and I am open to other ideas as well. I don't mind if we list a group of people and remove the mentor section but I don't want to create the impression that I can help with all of the projects that are listed.

timholy:
OK, see if this is OK. @zygmuntszpak, my intention definitely was not to sign you up for more than you bargained for; I was just trying to avoid giving the impression that I am willing to mentor 4 students simultaneously.

ChrisRackauckas:
Note, because there's no listed mentors on the page I am sending students to this PR since this is only place I know that documents who is willing to mentor these projects.

timholy:
> Note, because there's no listed mentors on the page

:confused:  https://julialang.org/soc/projects/images.html#where-to-go-for-discussion-and-to-find-mentors

ChrisRackauckas:
Aha I didn't find that. Thanks I'll link that.

PR #244, opened 2018-12-07
Add using Pkg to plotting examples.

fredrikekre:
Prompted by https://stackoverflow.com/questions/53661230/pkg-not-defined

### DIFF
--- a/downloads/plotting.html
+++ b/downloads/plotting.html
@@ -24,6 +24,7 @@ title:  Plotting in Julia
 
       <pre>
         <code class="julia">
+    using Pkg
     Pkg.add("Plots")
     using Plots
     plotly() # Choose the Plotly.jl backend for web interactivity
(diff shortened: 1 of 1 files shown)


downloads/plotting.html
```diff
@@ -24,7 +24,7 @@ <h2>Plots</h2>
 
       <pre>
         <code class="julia">
-    Pkg.add("Plots")
+    using Pkg; Pkg.add("Plots")
```
ararslan: It'd be nice to have the `using` on a separate line, which is more typical

downloads/plotting.html
```diff
@@ -24,7 +24,8 @@ <h2>Plots</h2>
 
       <pre>
         <code class="julia">
-    Pkg.add("Plots")
+    using Pkg
+       Pkg.add("Plots")
```
scls19fr: Maybe it should be indented the same as `using Pkg`

PR #247, opened 2018-12-07
Update the iteration blog post for Julia 1.1

timholy:
Now that Julia 1.1 is presumably just around the corner, now seems like a good time to update the iteration blog post. See

- https://discourse.julialang.org/t/psa-replacement-of-ind2sub-sub2ind-in-julia-0-7/14666/6
- https://discourse.julialang.org/t/what-is-the-proper-way-to-construct-cartesianindices-from-cartesianindex-in-julia-1-0/15579
- https://discourse.julialang.org/t/cartesianrange-julia-1-0/18403

CC @mauro3.

### DIFF
--- a/blog/_posts/2016-02-01-iteration.md
+++ b/blog/_posts/2016-02-01-iteration.md
@@ -4,33 +4,34 @@ title:  Multidimensional algorithms and iteration
 author: <a href="http://holylab.wustl.edu">Tim Holy</a>
 ---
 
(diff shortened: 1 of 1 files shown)


fredrikekre:
Maybe it is better to publish a new version instead?

ViralBShah:
I think it is fine to update it (and it clearly says so). Please merge away.

PR #248, opened 2018-12-13
Add GSOC project for generic linear algebra

simonbyrne:


JeffreySarnoff:
Its going to be <in part> about what demands does the GenericLinearAlgebra place upon all the incognito yet very specific numeric types which would be calling the generic linear algebra functions provided?
There is all sorts of minutiae that I have had to cover within the scope of each floating point type just to approach comity with some LinearAlgebra operations (Generic or Otherwise).

Two examples (true with v0.x and maybe still true):

one must provide `flipsign(fptype, Union{fptype, IEEEFloat, Signed})`  where it should suffice to provide `signbit` and `unary minus` (`sign` is also provided)  and this remains so with `copysign` provided

its not enough that Julia can form `Matrix{fptype}` (e.g. using provided `zeros(fpsign, rowcount, colcount)`) for some linear algebra ops -- within the computation something other than `+ - * /` and getindex/setindex occurs and whatever that may be (e.g. an expectation about the availability of rounding either to an integer value or to some number of fractional digits/bits) has been dropped back onto the fptype. 

There will be some specific minimal operational support required of any fptype that would enter the generic linear algebra through its exports or op signature extensions / abstractions.  It is necessary that the designer / gsocer know what that is precisely so the creator of a next floating point or real number type knows what must be done and then can decide what is to be done therebeyond.

JeffreySarnoff:
I could provide a bare bones floating point type that would allow the gsjocer to see the quiet assumptions she had made (and then either make it work absent that assumption holding or require that assumption hold).  With any assumption that becomes a requirement of the floating point or other number type, adding that capability to the type would be easy _maybe just require adding the appropriate tiny source file already written to the public package_.

simonbyrne:
@JeffreySarnoff That's a good point. Perhaps part of the project will be exploring and documenting what interfaces number types should provide to make them work?

PR #249, opened 2018-12-14
remove references to pkg.julialang.org which is no longer maintained

StefanKarpinski:


PallHaraldsson:
For those who still go there, could "v0.6.3 (current release) — v0.7.0-DEV (unstable)" be updated?

StefanKarpinski:
I don't know what you're referring to. Is it relevant to this PR in any way?

ViralBShah:
@PallHaraldsson is referring to pkg.julialang.org. 

No, we can't update it. But what we should do is say that this website is unmaintained. Or even better, just take it down.

Issue #255, opened 2018-12-21
Linux generic-binary installation instructions are confusing


natschil:
The current linux generic binary installation instructions give the example of how to link /usr/local/bin/julia to /bin/julia. This is weird as nobody downloads julia to /usr/local/, and also confusing as the next line then says talks about /usr/local/bin maybe not being in a user's $PATH. 

Is it enough to link <Downloaded_folder>/bin/julia to /usr/bin/julia, or is it necessary to also link to the lib/ folders? Perhaps instead of telling people how `ln -s` works in general is to provide a list of commands that will work on ubuntu/fedora and to assume everyone using something more exotic knows what they are doing anyways?

ViralBShah:
Just sufficient to link the binary and nothing else.

ViralBShah:
Submit a PR? What you say sounds right.

natschil:
#262

PR #262, opened 2019-01-08
Describe installation of generic linux binaries in more detail

natschil:
Previous instructions were confusing, this explains to how to extract
tarball and create symbolic link


natschil:
#255

natschil:
Oh nevermind, looks like this is already fixed in master.

PR #263, opened 2019-01-09
Add a github org (JuliaReach)

mforets:


mforets:
This is confusing. Seems like i wrote in [https://julialang.org/ecosystems/](https://julialang.org/ecosystems/), but it should have been in [https://julialang.org/community/](https://julialang.org/community/) (or in both?). Sorry for the confusion.

PR #264, opened 2019-01-10
Make GSoC pages use the same design as other pages.

aviks:
they no longer look like they are from another planet.

### DIFF
--- a/soc/ideas-page.md
+++ b/soc/ideas-page.md
@@ -1,5 +1,5 @@
 ---
-layout: default
+layout: insidepage
 title:  Summer of Code Ideas Page
 redirect_from:
   - /jsoc.html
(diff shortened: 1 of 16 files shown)


ChrisRackauckas:
Thanks so much!

soc/ideas-page.md
```diff
@@ -31,3 +31,5 @@ We have our project ideas organised roughly into the skillsets required:
 * [Graphics](projects/graphics.html) - projects ranging from low level OpenGL rendering to high level plotting
 
 You can see previous Summer of Code projects at the [Archive page](archive.html).
+
+</div
```
ChrisRackauckas: Looks like this slipped through.

PR #269, opened 2019-01-13
Fix copyright year

hycakir:
This should fix `©2018`.


ViralBShah:
Should be 2018-19, I think

hycakir:
It seems to be just a matter of taste. Many uses only the last year on the notice, while some uses `YYYY-20XX` style. I thought you have chosen the former. I can fix it however the way you want.

PR #271, opened 2019-01-13
add POS tagging to NLP list

oxinabox:
@aviks if doing any of the others POS tagging makes sense to do too.

IIRC NLTK pulls it off with a perception,
so it's not hard.


aviks:
Yes, thanks!

PR #274, opened 2019-01-18
DiffEqFlux Blog Post

ChrisRackauckas:
Two discussion points from the private channels which are still un-addressed:

- diffeq_rd is explained long before its use. Example should probably come first?
- 100 epochs of training without data is admittedly odd.

### DIFF
--- a/blog/_posts/2019-01-18-fluxdiffeq.md
+++ b/blog/_posts/2019-01-18-fluxdiffeq.md
@@ -159,7 +159,7 @@ and provide an initial condition `u0`, and a timespan `tspan` to solve over, and
 specify the parameters `p`.
 
 For example, the
-[Lotka-Volerra equations describe the dynamics of the population of  rabbits and wolves](https://en.wikipedia.org/wiki/Lotka%E2%80%93Volterra_equations).
+[Lotka-Volerra equations describe the dynamics of the population of rabbits and wolves](https://en.wikipedia.org/wiki/Lotka%E2%80%93Volterra_equations).
 They can be written as:
 
 $$
--- a/blog/_posts/2019-01-18-fluxdiffeq.md
+++ b/blog/_posts/2019-01-18-fluxdiffeq.md
@@ -10,12 +10,13 @@ robustly add differential equation solvers to neural networks in Julia.
 ![Flux ODE Training Animation](https://user-images.githubusercontent.com/1814174/51399500-1f4dd080-1b14-11e9-8c9d-144f93b6eac2.gif)
 
 The [Neural Ordinary Differential Equations](https://arxiv.org/abs/1806.07366)
-paper has been a hot topic even before it made a splash as Best Paper of NeurIPS
-2018. Among other demonstrations, combining these two disparate fields enabled
-the authors to reduce the number of parameters and reduce memory costs in
-comparison to ResNet. But this is only the beginning: neural networks and
-differential equations were born to be together. This blog post will explain
-why, and start to give a sense of what's possible with state-of-the-art tools.
+paper has been a hot topic even before it made a splash as Best Paper of
+NeurIPS 2018. By mixing ordinary differential equations and neural networks
+they were able to reduce the number of parameters and reduce memory costs in
+comparison to ResNet. Now with the floodgates opened causing a merge of
+differential equations with machine learning, the purpose of this blog post is
+to introduce the reader to differential equations from a data science
+perspective and show how to mix these tools with neural nets.
 
 The advantages of the Julia
 [DifferentialEquations.jl](https://github.com/JuliaDiffEq/DifferentialEquations.jl) library for numerically solving differential equations have been
@@ -27,7 +28,7 @@ it includes other modern features such as
 [distributed (multi-node) parallelism](http://docs.juliadiffeq.org/latest/features/monte_carlo.html),
 and [sophisticated event handling](http://docs.juliadiffeq.org/latest/features/callback_functions.html).
 Recently, these native Julia differential equation solvers have successfully been embedded
-into the [Flux](https://github.com/FluxML/Flux.jl/) deep learning package, to allow the use of a full suite of
+into the [Flux.jl](https://github.com/FluxML/Flux.jl/) deep learning package, to allow the use of a full suite of
 highly tested and optimized DiffEq methods within neural networks. Using the new package
 [DiffEqFlux.jl](https://github.com/JuliaDiffEq/DiffEqFlux.jl/),
 we will show the reader how to easily add differential equation
@@ -35,13 +36,14 @@ layers to neural networks using a range of differential equations models, includ
 differential equations, stochastic differential equations, delay differential
 equations, and hybrid (discontinuous) differential equations.
 
-This is the first toolbox to combine a fully-featured differential equations
-solver library and neural networks seamlessly together. The blog post will also
-show why the flexibility of a full differential equation solver suite is
-necessary. With the ability to fuse neural networks with ODEs, SDEs, DAEs, DDEs,
-stiff equations, and different methods for adjoint sensitivity calculations,
-this is a large generalization of the neural ODEs work and will allow
-researchers to better explore the problem domain.
+This is the first user-friendly toolbox to combine a fully-featured differential
+equations solver library and neural networks seamlessly together. The blog post will also show
+why the flexibility of a full differential equation solver suite is necessary
+for doing this properly. With the ability to fuse neural networks
+seamlessly with ODEs, SDEs, DAEs, DDEs, stiff equations, and different methods
+for adjoint sensitivity calculations, this is a large generalization of the
+neural ODEs work and will allow researchers to better explore the problem
+domain.
 
 (Note: If you are interested in this work and are an undergraduate or graduate
 student, we have [Google Summer of Code projects available in this area](https://julialang.org/soc). This
@@ -73,7 +75,7 @@ about it differently, this is actually just a nonlinear transformation $y=ML(x)$
 The reason $ML$ is interesting is because its form is basic but adapts to the
 data itself. For example, a simple neural network (in design matrix form) with
 sigmoid activation functions is simply matrix multiplications followed
-by application of sigmoid functions. Specifically, $ML(x)=\sigma(W_{3}\cdot\sigma(W_{2}\cdot\sigma(W_{1}\cdot x)))$ is a three-layer deep
+by application of sigmoid functions. Specifically, meaning $ML(x)=\sigma(W_{3}\cdot\sigma(W_{2}\cdot\sigma(W_{1}\cdot x)))$ is a three-layer deep
 neural network, where $W=(W_1,W_2,W_3)$ are learnable parameters.
 You then choose $W$ such that $ML(x)=y$ reasonably fits the function you wanted it to fit.
 The theory and practice of machine learning confirms that this is a good way to learn nonlinearities.
@@ -137,16 +139,16 @@ The way the Euler method works is based on the fact that $y'(x) = \frac{dy}{dx}$
 
 $$\Delta y = (y_\text{next} - y_\text{prev}) = \Delta x\cdot ML(x) \text{ which implies that } y_{i+1} = y_{i} + \Delta x\cdot ML(x_{i}).$$
 
-This looks similar in structure to a ResNet, one of the most successful image
-processing models. The insight of the the Neural ODEs paper was that
-increasingly deep and powerful ResNet-like models effectively approximate a kind
-of "infinitely deep" model as each layer tends to zero. Rather than adding more
-layers, we can just model the differential equation directly and then solve it
-using a purpose-built ODE solver. Numerical ODE solvers are a science that goes
-all the way back to the first computers, and modern ones can adaptively choose
-step sizes $\Delta x$ and use high order approximations to dratically reduce the
-number of actual steps required. And as it turns out, this works well in
-practice, too.
+Looking familiar now? This is the basis of a residual network (ResNet) which is a type of recurrent neural network (RNN),
+and is the foundation of many of the best AI systems known to date.
+The genius of the neural ordinary differential equation paper was to notice this and say,
+let's just model the differential equation directly and then solve it using a much better numerical ODE solver methods.
+The creation of numerical ODE solvers is a science that goes
+all the way back to the first computers, and can adaptively choose step sizes $\Delta x$
+and use high order approximations to reduce the number of steps required to get
+sufficiently small error. By swapping out the Euler discretization with the LSODE
+numerical ODE solver, their results were good enough to make everyone think twice
+about reflexively using an RNN.
 
 ## How do you solve an ODE?
 
@@ -196,6 +198,40 @@ plot(sol)
 
 ![LV Solution Plot](https://user-images.githubusercontent.com/1814174/51388169-9a07f300-1af6-11e9-8c6c-83c41e81d11c.png)
 
+Optional arguments can be passed to control the ODE solver to customise most aspects of the numerical integration.
+For example, if you've some familiarity with the area, a good choice for
+this equation is to use the 5th order Runge-Kutta method `Tsit5()` (related to the `dopri` method,
+or `ode45` in MATLAB). We can also say that we want this solved with a relative error
+tolerance of `1e-8`, and that we want the solution saved at every `0.1` time points. This looks like:
+
+```julia
+sol = solve(prob,Tsit5(),saveat=0.1,reltol=1e-8)
+```
+
+There are a lot more arguments you can pass in to customise most aspects of the ODE solver,
+as
+[specified in the documentation](http://docs.juliadiffeq.org/latest/basics/common_solver_opts.html),
+but these will suffice for our demonstration.
+
+Note that the solution of a differential equation is a continuous object, so
+for example
+
+```julia
+sol(5.24)
+```
+
+gives the solution at time `5.24`. It also acts as an array, where
+
+```julia
+sol[2,5]
+sol.t[5]
+```
+
+is the solution of the 2nd variable (`y`) at the 5th time point. Notice that
+the 5th time point is `0.4` since we set `saveat=0.1`. If we did not set `saveat`,
+this would be adaptively chosen. More details on handling the solution type
+[can be found in the DifferentialEquations.jl documentation](http://docs.juliadiffeq.org/latest/basics/solution.html).
+
 One last thing to note is that we can make our initial condition (`u0`) and time spans (`tspans`)
 to be functions of the parameters (the elements of `p`). For example, we can define the `ODEProblem`:
 
@@ -209,33 +245,20 @@ prob = ODEProblem(lotka_volterra,u0_f,tspan_f,p)
 In this form, everything about the problem is determined by the parameter vector (`p`, referred to
 as `θ` in associated literature). The utility of this will be seen later.
 
-DifferentialEquations.jl has many powerful options for customising things like
-accuracy, tolerances, solver methods, events and more; check out [the
-docs](http://docs.juliadiffeq.org/latest/) for more details on how to use it in
-more advanced ways.
-
 ## Let's Put an ODE Into a Neural Net Framework!
 
-To understand embedding an ODE into a neural network, let's look at what a
-neural network layer actually is. A layer is really just a *differentiable
-function* which takes in a vector of size `n` and spits out a new vector of size
-`m`. That's it! Layers have traditionally been simple functions like matrix
-multiply, but in the spirit of [differentiable
-programming](https://julialang.org/blog/2018/12/ml-language-compiler) people are
-increasingly experimenting with much more complex functions, such as ray tracers and
-physics engines.
+To understand embedding an ODE into a neural network, let's recall what a layer
+of a neural network actually is. A layer is a transformation 𝐑ⁿ→𝐑ᵐ which
+takes in a vector of size `n` and spits out a new vector of size `m`. That's it!
+So a differential equation layer is simply a layer that takes in a parameter
+vector `p` (of size `n`) and spits out a new vector (of size `m`). You can do this
+manually using the
+[`remake` function in DifferentialEquations.jl](http://docs.juliadiffeq.org/latest/basics/problem.html#Modification-of-problem-types-1),
+or you can use the Flux.jl layers defined in DiffEqFlux.jl. Let's give those a spin.
 
-Turns out that differential equations solvers fit this framework, too: A solve
-takes in some vector `p` (which might include parameters like the initial
-starting point), and outputs some new vector, the solution. Moreover it's
-differentiable, which means we can put it straight into a larger differentiable
-program. This larger program can happily include neural networks, and we can keep
-using standard optimisation techniques like ADAM to optimise their weights.
-
-DiffEqFlux.jl makes it convenient to do just this; let's take it for a spin.
-
-The most basic differential equation layer is `diffeq_rd`. The way to understand this
-function is that it does the following. It takes in parameters for the integrand, `p`,
+The most basic differential equation layer is `diffeq_rd`. The function is of
+the form `diffeq_rd(p,reduction,prob,args...;kwargs...)`. The way to understand this
+function is that it does the following. It takes in the parameter vector `p`,
 it puts it in the differential equation defined by `prob`, solves it with
 the chosen arguments (solver, tolerance, etc), and then gets a vector out through
 some function `reduction`. For example, in our Lotka-Volerra problem, let's say we wanted
@@ -265,7 +288,7 @@ our solution handling be in the function `reduction` used above in `diffeq_rd(p,
 Thus this code is equivalent to:
 
 ```julia
-using Flux, DiffEqFlux
+using Flux, FluxDiffEq
 reduction(sol) = sol[1,:]
 diffeq_rd(p,reduction,prob,Tsit5(),saveat=0.1)
 ```
@@ -275,7 +298,7 @@ solves the Lotka-Volterra problem `prob` with our chosen ODE solver arguments,
 and then gets a vector out via `reduction` which is the time series of `x(t)`.
 
 The nice thing about `diffeq_rd` is that it takes care of the type handling
-necessary to make it compatible with the neural network framework (here Flux). To show this,
+necessary to make it compatible with the neural network framework (here Flux.jl). To show this,
 let's define a neural network with the function as our single layer, and then a loss
 function that is the squared distance of the output values from `1`. In Flux, this looks like:
 
@@ -310,7 +333,7 @@ Flux.train!(loss_rd, params, data, opt, cb = cb)
 
 The result of this is the animation shown at the top.
 
-Flux finds the parameters of the neural network (`p`) which minimize
+Flux.jl finds the parameters of the neural network (`p`) which minimize
 the cost function, i.e. it trains the neural network: it just so happens that
 the forward pass of the neural network includes solving an ODE.
 Since our cost function put a penalty whenever the number of
@@ -318,7 +341,7 @@ rabbits was far from 1, our neural network found parameters where our population
 of rabbits and wolves are both constant 1.
 
 Now that we have solving ODEs as just a layer, we can add it anywhere. For example,
-the multilayer perceptron is written in Flux as
+the multilayer perceptron is written in Flux.jl as
 
 ```julia
 m = Chain(
@@ -353,26 +376,27 @@ m = Chain(
   Dense(288, 10), softmax) |> gpu
 ```
 
-As long as you can write down the forward pass, we can take any parameterised,
-differentiable program and optimise it. The world is your oyster.
+The world is your oyster.
 
 ## Why is a full ODE solver suite necessary for doing this well?
 
-Where we have combined an existing solver suite and deep learning library, the
-excellent [torchdiffeq](https://github.com/rtqichen/torchdiffeq) project takes
-an alternative approach, instead implementing solver methods directly in
-PyTorch, including an adaptive Runge Kutta 4-5 (`dopri5`) and an
-Adams-Bashforth-Moulton method (`adams`). However, while their approach is very
-effective for certain kinds of models, not having access to a full solver suite
-is limiting.
+With Julia we have solved the technical issue of embedding native
+differential equation solvers inside of the neural network framework. Details
+of how it was done will follow, but we have to address the necessary question
+first: why was this important to do at all? Or another way to put it is, why
+not just re-implement a few simple ODE solvers into a neural network framework
+and be done with it?
 
-Consider the following example, the [ROBER
-ODE](https://www.radford.edu/~thompson/vodef90web/problems/demosnodislin/Single/DemoRobertson/demorobertson.pdf).
-The most well-tested (and optimized) implementation of an
-Adams-Bashforth-Moulton method is the [CVODE integrator in the C++ package
-SUNDIALS](https://computation.llnl.gov/projects/sundials) (a derivative of the
-classic LSODE). Let's use DifferentialEquations.jl to call CVODE with its Adams
-method and have it solve the ODE for us:
+This is the strategy which was taken with
+[torchdiffeq](https://github.com/rtqichen/torchdiffeq) which implements an adaptive
+Runge Kutta 4-5 (`dopri5`) and an Adams-Bashforth-Moulton method (`adams`)
+The issue is that these two methods are not sufficient for all, or even most, ODEs. A classic example of this is the
+[ROBER ODE](https://www.radford.edu/~thompson/vodef90web/problems/demosnodislin/Single/DemoRobertson/demorobertson.pdf).
+The most well-tested (and optimized) implementation of an Adams-Bashforth-Moulton
+method is the [CVODE integrator in the C++ package SUNDIALS](https://computation.llnl.gov/projects/sundials)
+(a derivative of the classic LSODE).
+Let's use DifferentialEquations.jl to call CVODE with its Adams method and have
+it solve the ODE for us:
 
 ```julia
 rober = @ode_def Rober begin
@@ -384,16 +408,36 @@ prob = ODEProblem(rober,[1.0;0.0;0.0],(0.0,1e11),(0.04,3e7,1e4))
 solve(prob,CVODE_Adams())
 ```
 
-(For those familiar with solving ODEs in MATLAB, this is similar to `ode113`)
+(For those familiar with solving ODEs in  MATLAB, this is similar to `ode113`)
 
-Both this and the `dopri` method from [Ernst Hairer's Fortran
-Suite](https://www.unige.ch/~hairer/software.html) stall and fail to solve the
-equation. This happens because the ODE is
-[stiff](https://en.wikipedia.org/wiki/Stiff_equation), and thus methods with
-"smaller stability regions" will not be able to solve it appropriately (for more
-details, I suggest reading Hairer's Solving Ordinary Differential Equations II).
-On the other hand `KenCarp4()` to this problem, the equation is solved in a
-blink of an eye:
+If you try the `dopri` method from
+[Ernst Hairer's Fortran Suite](https://www.unige.ch/~hairer/software.html), you
+get the same issue:
+
+```julia
+using ODEInterfaceDiffEq
+solve(prob,dopri5())
+```
+
+(note: this is like MATLAB's `ode45`)
+
+Both of these stall and fail to solve the equation. It's not an implementation
+issue either. Our native Julia versions of these methods (`DP5()` and `VCABM`)
+also exhibit the same behavior. It's actually a well-known property of these
+methods on this kind of ODE.
+This ODE is known to be a [stiff ODE](https://en.wikipedia.org/wiki/Stiff_equation),
+and thus methods with "smaller stability regions" will not be able to solve it
+appropriately (for more details, I suggest reading Hairer's Solving Ordinary
+Differential Equations II).
+
+I chose these two methods because these are the ones implemented in torchdiffeq
+and mentioned in the neural ordinary differential equations paper. They are not
+bad methods by any means, they are classics because of how efficient they are
+on a good number of problems. But it's clear from this example that they are
+not applicable to all ODEs. Instead, you need the flexibility to choose other
+solvers which have different properties such as increased stiffness. For example,
+when applying `KenCarp4()` to this problem, the equation is solved in a blink of
+an eye:
 
 ```julia
 sol = solve(prob,KenCarp4())
@@ -403,23 +447,25 @@ plot(sol,xscale=:log10,tspan=(0.1,1e11))
 
 ![ROBER Plot](https://user-images.githubusercontent.com/1814174/51388944-eb18e680-1af8-11e9-874f-09478759596e.png)
 
-This is just one example of subtlety in integration: Stabilizing explicit
-methods via PI-adaptive controllers, step prediction in implicit solvers, etc.
-are all intricate details that take a lot of time and testing to become
-efficient and robust. Different problems require different methods: [Symplectic
-integrators](http://docs.juliadiffeq.org/latest/solvers/dynamical_solve.html#Symplectic-Integrators-1)
-are required to [adequately handle physical many problems without
-drift](https://scicomp.stackexchange.com/questions/29149/what-does-symplectic-mean-in-reference-to-numerical-integrators-and-does-scip/29154#29154),
-and tools like [IMEX
-integrators](http://docs.juliadiffeq.org/latest/solvers/split_ode_solve.html#Implicit-Explicit-(IMEX)-ODE-1)
-are required to handle ODEs which [come from partial differential
-equations](https://www.youtube.com/watch?v=okGybBmihOE). Building a
-production-quality solver is thus an enormous undertaking and relatively few
-exist.
-
-Rather than building an ML-specific solver suite in parallel to one suitable for
-scientific computing, in Julia they are one and the same, meaning you can take
-advantage of all of these methods today.
+Stabilizing explicit methods via PI-adaptive controllers, step prediction in
+implicit solvers, etc. are all intricate details that take a lot of time and
+testing in order to build methods which are efficient for the more difficult
+equations. This is why very few production-quality solvers have ever been
+created. For example, R's deSolve and Python's SciPy simply wrap the older
+Fortran methods, as detailed
+[in the ODE solver summary post](http://www.stochasticlifestyle.com/comparison-differential-equation-solver-suites-matlab-r-julia-python-c-fortran/))
+Not only that, but different problems require different methods.
+[Symplectic integrators](http://docs.juliadiffeq.org/latest/solvers/dynamical_solve.html#Symplectic-Integrators-1)
+are required to [adequately handle physical many problems without drift](https://scicomp.stackexchange.com/questions/29149/what-does-symplectic-mean-in-reference-to-numerical-integrators-and-does-scip/29154#29154),
+and tools like [IMEX integrators](http://docs.juliadiffeq.org/latest/solvers/split_ode_solve.html#Implicit-Explicit-(IMEX)-ODE-1)
+are required to handle ODEs which [come from partial differential equations](https://www.youtube.com/watch?v=okGybBmihOE).
+Thus building and maintaining a full set of optimized differential equation
+solvers is itself an entire research programme. Given the intricacy of such
+a topic, reimplementing the solvers with all of the details into every
+machine learning framework is an infeasible task. By embedding the work of our
+DifferentialEquations.jl project into Flux.jl, we can continue to develop the
+differential equation solvers for traditional scientific computing
+and the users of Flux.jl will benefit from all of our work.
 
 ## What kinds of differential equations are there?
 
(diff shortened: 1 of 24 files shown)


ChrisRackauckas:
Adding in the media next.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,632 @@
+---
+layout: post
+title:  FluxDiffEq.jl: A Julia Library for Neural Differential Equations
+author: Chris Rackauckas, Mike Innes, Yingbo Ma, Lyndon White, Vaibhav Dixit
+---
+
+The [Neural Ordinary Differential Equations](https://arxiv.org/abs/1806.07366)
+paper has been a hot topic ever since it has made a splash as Best Paper of
```
oxinabox:
```suggestion
paper has been a hot topic even before it made a splash as Best Paper of
```

MikeInnes:
Maybe it's better to just write out the loop rather than using `train!`. Being able to use that is kind of neat but probably less clear overall.

ChrisRackauckas:
Some media:

lv_sol

data_points

rober_plot

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,632 @@
...
+differential equations from a data science perspective and show how to mix these
+tools with neural nets.
+
+The advantages of the Julia DifferentialEquations.jl library for numerically solving
+differential equations have been
+[discussed in detail in other posts](http://www.stochasticlifestyle.com/comparison-differential-equation-solver-suites-matlab-r-julia-python-c-fortran/).
+Recently, these native Julia differential equation solvers have successfully been embedded
+into the Flux.jl deep learning package, to allow the use of a full suite of
+highly tested and optimized DiffEq methods within neural networks. Using the new package
+DiffEqFlux.jl, we will show the reader how to easily add differential equation
```
oxinabox:
```suggestion
[DiffEqFlux.jl](https://github.com/JuliaDiffEq/DiffEqFlux.jl/), we will show the reader how to easily add differential equation
```

Lets link it at least the first time

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,632 @@
...
+floodgates opened causing a merge of differential equations merging with machine
+learning, the purpose of this blog post is into introduce the reader to
+differential equations from a data science perspective and show how to mix these
+tools with neural nets.
+
+The advantages of the Julia DifferentialEquations.jl library for numerically solving
+differential equations have been
+[discussed in detail in other posts](http://www.stochasticlifestyle.com/comparison-differential-equation-solver-suites-matlab-r-julia-python-c-fortran/).
+Recently, these native Julia differential equation solvers have successfully been embedded
+into the Flux.jl deep learning package, to allow the use of a full suite of
```
oxinabox:
```suggestion
into the [Flux.jl](https://github.com/FluxML/Flux.jl/) deep learning package, to allow the use of a full suite of
```

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,632 @@
...
+The [Neural Ordinary Differential Equations](https://arxiv.org/abs/1806.07366)
+paper has been a hot topic even before it made a splash as Best Paper of
+NeuralIPS 2018. By mixing ordinary differential equations and neural networks
+they were able to improve training speeds and accuracy over ResNet. Now with the
+floodgates opened causing a merge of differential equations merging with machine
+learning, the purpose of this blog post is into introduce the reader to
+differential equations from a data science perspective and show how to mix these
+tools with neural nets.
+
+The advantages of the Julia DifferentialEquations.jl library for numerically solving
```
oxinabox:
```suggestion
The advantages of the Julia [DifferentialEquations.jl](https://github.com/JuliaDiffEq/DifferentialEquations.jl) library for numerically solving
```

ChrisRackauckas:
Needs https://github.com/JuliaDiff/ForwardDiff.jl/pull/378 and a ForwardDiff tag @jrevels

KristofferC:
That PR still fails test.

ChrisRackauckas:
https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/93 breaks the layer because of the sizing issue. Temporary fix:

```julia
function delay_lotka_volterra(du,u,h,p,t)
  x, y = u
  α, β, δ, γ = p
  du[1] = dx = (α - β*y)*h(p,t-0.1)[1]
  du[2] = dy = (δ*x - γ)*y
end
h(p,t) = ones(eltype(p),2)
prob = DDEProblem(delay_lotka_volterra,[1.0,1.0],h,(0.0,10.0),constant_lags=[0.1])
p = param([2.2, 1.0, 2.0, 0.4])
params = Flux.Params([p])
reduction2(sol) = sol[1,1:101]
function predict_fd_dde()
  diffeq_fd(p,reduction2,101,prob,MethodOfSteps(Tsit5()),saveat=0.1)
end
loss_fd_dde() = sum(abs2,x-1 for x in predict_fd_dde())
loss_fd_dde()
```

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,643 @@
...
+title:  FluxDiffEq.jl: A Julia Library for Neural Differential Equations
+author: Chris Rackauckas, Mike Innes, Yingbo Ma, Lyndon White, Vaibhav Dixit
+---
+
+The [Neural Ordinary Differential Equations](https://arxiv.org/abs/1806.07366)
+paper has been a hot topic even before it made a splash as Best Paper of
+NeuralIPS 2018. By mixing ordinary differential equations and neural networks
+they were able to improve training speeds and accuracy over ResNet. Now with the
+floodgates opened causing a merge of differential equations merging with machine
+learning, the purpose of this blog post is into introduce the reader to
```
jrevels: > a merge of differential equations merging

Should merge the "merge" to reduce the "merge"s 😛

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,634 @@
...
+It is worth noting that FluxDiffEq.jl uses is only around ~50 lines of code to pull this all off.
+
+## Implementing the Neural ODE layer in Julia
+
+Let's go all the way back for a second and now implement the neural ODE layer
+in Julia. Remember that this is simply an ODE where the derivative function
+is defined by a neural network itself. To do this, let's first define the
+neural net for the derivative. We can use a multilayer ...
+
+## The core technical challenge: backpropogation through differential equation solvers
```
GunnarFarneback: propogation -> propagation
Twice more below.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,634 @@
...
+```
+
+(here we once again use the CVODE C++ solvers from SUNDIALS since they are a close
+match to the SciPy integrators used in the neural ODE paper)
+
+This inaccuracy is the reason why the method from the neural ODE paper is not
+implemented in software suites, but it once again highlights a detail. Not
+all ODEs will have a large error due to this issue. And on ODEs where it's not
+a problem, this will be the most efficient way to do adjoint sensitivity
+analysis. Thus once again we arive at the conclusion that one method is not
```
GunnarFarneback: arrive

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,650 @@
...
+will show you how to easily, efficiently, and
+robustly add differential equation solvers to
+neural networks in Julia.
+
+![Flux ODE Training Animation](https://user-images.githubusercontent.com/1814174/51399500-1f4dd080-1b14-11e9-8c9d-144f93b6eac2.gif)
+
+The [Neural Ordinary Differential Equations](https://arxiv.org/abs/1806.07366)
+paper has been a hot topic even before it made a splash as Best Paper of
+NeuralIPS 2018. By mixing ordinary differential equations and neural networks
+they were able to improve training speeds and accuracy over ResNet. Now with the
```
jessebett: We don't make any claims of improved training speeds or accuracy over a comparably sized ResNet. The win is fewer parameters and O(1) memory to achieve similar performance as ResNet:
image

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,650 @@
...
+# Display the ODE with the initial parameter values.
+cb()
+
+Flux.train!(loss_rd, params, data, opt, cb = cb)
+```
+
+The result of this is the animation shown at the top.
+
+Flux.jl finds the parameters of the neural network (`p`) which minimize
+the cost function, i.e. it trains the neural network: it just so happens that here training the neural network happens to include solving an ODE.
```
jessebett: "training the neural network happens to include solving an ODE" kinda implies that it's the optimizer that requires solving the ODE, not the objective.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,650 @@
...
+[discussed in detail in other posts](http://www.stochasticlifestyle.com/comparison-differential-equation-solver-suites-matlab-r-julia-python-c-fortran/).
+Recently, these native Julia differential equation solvers have successfully been embedded
+into the [Flux.jl](https://github.com/FluxML/Flux.jl/) deep learning package, to allow the use of a full suite of
+highly tested and optimized DiffEq methods within neural networks. Using the new package
+[DiffEqFlux.jl](https://github.com/JuliaDiffEq/DiffEqFlux.jl/), we will show the reader how to easily add differential equation
+layers to neural networks using a range of differential equations models, including stiff ordinary
+differential equations, stochastic differential equations, delay differential
+equations, and hybrid (discontinuous) differential equations.
+
+This is the first user-friendly toolbox to combine state-of-the-art differential
```
jessebett: Can you qualify this statement with torchdiffeq? Possibly expand 'user-friendly' to distinguish benefits of DiffEq.jl.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,650 @@
...
+In this form, everything about the problem is determined by the parameter vector (`p`, referred to
+as `θ` in associated literature). The utility of this will be seen later.
+
+## Let's Put an ODE Into a Neural Net Framework!
+
+To understand embedding an ODE into a neural network, let's recall what a layer
+of a neural network actually is. A layer is a transformation 𝐑ⁿ→𝐑ᵐ which
+takes in a vector of size `n` and spits out a new vector of size `m`. That's it!
+So a differential equation layer is simply a layer that takes in a parameter
+vector `p` (of size `n`) and spits out a new vector (of size `m`). You can do this
```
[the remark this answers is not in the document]
ChrisRackauckas: Well `p` can be the input to `u0` and final timespan, that is mentioned later down. So NeuralODE is only using the `p` here for its relation to `u0`, but in a sense it's an expanded form. It does interact oddly with the adjoint method though. @MikeInnes would you recommend having a separate input for these? I believe Flux will differentiate w.r.t. multiple different inputs? The best would be for it to be done via keyword arguments, i.e. `u0 = ..., p = ...`. Then NeuralODE would only use `u0` and other applications would use `p` or both.

blog/_posts/2019-01-18-fluxdiffeq.md
ChrisRackauckas: Maybe there's a better way of putting this together. I mean for it to kind of go with the previous line, essentially saying this is the first toolbox which combines neural networks with well-optimized and battle tested DE solvers for stiff and non-stiff ODEs, DAEs, DDEs, SDEs, etc. torchdiffeq only does the two basic explicit RK methods (and then a few low order fixed time step methods), so I would think the difference is clear. 

The libraries that would need to be qualified against more IMO are FATODE and Sundials which are also production-quality ODE solvers with adjoint sensitivity analysis implemented, but I think in their case it's clear that they aren't hooked up with neural nets (though they could be).

blog/_posts/2019-01-18-fluxdiffeq.md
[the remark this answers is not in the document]
jessebett: Sure but even if `p` is `u0` and final timespan it's weird to talk about the dimensionality of the *layer* being that of `p`, imo. Like in this example you wouldn't care about the dimensionality of the timespan.

Is there, in general, some other part of `p` that is changing through time besides `u`? Because when we talk about the layer dimensionality in any other layer we usually mean the dim of the input state, so `u0` and the dim of the output state, `u(t_end)`.  

I know you understand what I'm talking about, just being really explicit here. Is it worth it, in this blog, to be explicit about `p` dimensionality. Or is that a more niche interest? Could we keep it consistent with NeuralODEs in this blog for comparability if that's a reasonable simplifying assumption for an introductory blog?

ChrisRackauckas:
Here's a Gist with the code. It may make sense to link this in the post somewhere.

https://gist.github.com/ChrisRackauckas/89bb4181bb1c82266daeeec68c7f8d0e

blog/_posts/2019-01-18-fluxdiffeq.md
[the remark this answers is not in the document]
ChrisRackauckas: Is this just a naming thing? What about just calling it `x` instead and noting that `x` is essentially `x = [u0,p,tspan]`, would that make it more clear? Dealing with multiple inputs is probably a better way to handle it if Flux can do that.

>I know you understand what I'm talking about, just being really explicit here. Is it worth it, in this blog, to be explicit about p dimensionality. Or is that a more niche interest?

No, thinking of the solution of an ODE as a function of `p` is definitely not of niche interest. It's the basis of parameter estimation which is by far the biggest use of optimization on ODE solutions. Actually, NeuralODE is very niche in comparison to parameter estimation (which is where the adjoint methods originate from). So for the purposes of integrating data science and ML with differential equations, having the diffeq as both a function of `u0` and `p` are vital to the standard use cases.

blog/_posts/2019-01-18-fluxdiffeq.md
ChrisRackauckas: See if the changed introduction satisfies you. It more qualifies the statement by listing some of the standard features of modern diffeq solver libraries, like distributed parallelism and event handling.

blog/_posts/2019-01-18-fluxdiffeq.md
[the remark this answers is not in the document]
MikeInnes: The confusion here is because Jesse is thinking of optimisation parameters, and Chris is thinking of parameters of the ODE. Probably in all of the neural ODEs work they were the same, but in e.g. the lotka-volterra case they are not; the NN actually generates a vector `p` that gets fed into `f(p)` (which happens to solve an ODE, but isn't a neural ODE in the sense of the paper). In that sense `p` is data in the network rather than an optimisation parameter.

I'm not sure how best to resolve this; one option is to clearly distinguish upfront between the two (and it might be best to describe optimisation parameters as "weights"). I think it's worth spending some time to get this right though, or it'll confuse a lot of people.

ViralBShah:
I think we should put up a repo with all the code rather than a gist.

cormullion:
A few little typos and formatting glitches for you:

line 17-18 "the purpose of this blog post is ~~in~~to introduce the reader" 
line 64 "_a priori_"
line 75 "sigmoid activation functions then the model " ?
line 225 "at the ~~the~~"
line 332 "it just so happens that here that the " - too many 'that's
line 405 "famil _i_ ar"
line 475 "same interface as ODEs, it ~~they~~ can "

blog/_posts/2019-01-18-fluxdiffeq.md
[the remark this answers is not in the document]
jekbradbury: Definitely good to be explicit upfront about the difference; it confused me too.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,657 @@
...
+analysis. Thus once again we arrive at the conclusion that one method is not
+enough.
+
+In DifferentialEquations.jl we are investigating many different methods for
+computing the derivatives of differential equations with respect to parameters.
+We have a [recent preprint](https://arxiv.org/abs/1812.01892) detailing
+some of these results. One of the things we have found is that direct use of
+automatic differentiation can be one of the most efficient and flexible methods.
+Julia's ForwardDiff.jl, Flux.jl, and ReverseDiff.jl can directly be applied to
+perform automatic differentiation on the native Julia differential equation
```
jekbradbury: It's maybe worth mentioning explicitly that this is something only Julia can do well because AD packages in C/C++/Fortran aren't suitable/available/etc. while reimplementing an ODE solver in terms of PyTorch or TensorFlow primitives would be much harder and end up much slower?

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,657 @@
...
+solvers themselves, and this can increase performance while giving new features.
+Our findings show that forward-mode automatic differentiation is fastest when
+there are less than 100 parameters in the differential equations, and that for
+large enough number of parameters adjoint sensitivity analysis is the most efficient. Even
```
jekbradbury:
```suggestion
a large enough number of parameters adjoint sensitivity analysis is the most efficient. Even
```

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,657 @@
...
+Looking familiar now? This is the basis of a residual network (ResNet) which is  type of RNN,
+and is the foundation of many of the best AI systems known to date.
+The genius of the neural ordinary differential equation paper was to notice this and say,
+let's just model the differential equation directly and then solve it using a much better numerical ODE solver methods.
+The creation of numerical ODE solvers is a science that goes
+all the way back to the first computers, and can adaptively choose step sizes $\Delta x$
+and use high order approximations to reduce the number of steps required to get
+sufficiently small error. By swapping out the Euler discretization with the LSODE
+numerical ODE solver, their results were good enough to make everyone think twice
+about reflexively using an RNN.
```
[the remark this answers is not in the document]
ChrisRackauckas: Well, saying "make everyone think twice" I think isn't going to far. It's saying that people should start at least considering it. I think it's at least entered the thought cycle.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,657 @@
...
+## What is the Neural Ordinary Differential Equation (ODE)?
+
+The neural ordinary differential equation is one of many ways to put these two
+subjects together. The simplest way of explaining it is that, instead of
+learning the nonlinear transformation directly, we wish to learn the structures
+of the nonlinear transformation. Thus instead of doing $y=ML(x)$, we put the
+machine learning model on the derivative, $y'(x) = ML(x)$, and now solve the ODE.
+Why would you ever do this? Well, one motivation is that, if you were to do this
+and then solve the ODE using the simplest and most error prone method, the
+Euler method, what you get is equivalent to a [residual neural network](https://arxiv.org/abs/1512.03385).
```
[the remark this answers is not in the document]
ChrisRackauckas: Maybe it's because ODEs are my bread and butter, but I assumed it's pretty straightforward that if you see it as a method for solving some underlying ODE, you just write out that ODE and solve it another way. The "underlying ODE" part is the convergence of the Euler method as dt->0, which if others think should be mentioned in more detail please submit some text.

blog/_posts/2019-01-18-fluxdiffeq.md
ChrisRackauckas: Added a paragraph which makes this explicit.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,672 @@
...
+to define a nonlinear transformation. The first is to explicitly say what it is.
+For example, the sigmoid function is $\sigma(x)=\frac{e^x}{e^x + 1}$. This only works
+if you know the exact functional form that relates the input to the output.
+However, in many cases, such exact relations are not known *a priori*.
+So how do you do nonlinear modeling if you don't know the nonlinearity?
+
+There are two separate ways to then handle this problem. One method is machine
+learning. Usually this is presented as you have some $x$ and you want to predict
+$y$. This generation of a prediction from $x$ is the machine learning model, call it
+$ML$, where you put $x$ into the program and it spits out values $y$. But if you think
```
kmsquire: Suggested wording:
```
There are two common ways to handle this problem. One method is to use machine
learning. In a typical machine learning problem, you are given some input $x$ and 
you want to predict an output $y$. This generation of a prediction $y$ from $x$
is a machine learning model (let's call it $ML$).  During training, we attempt to 
adjust the parameters of $ML$ so that it generates accurate predictions.  We
can then use $ML$ for inference (i.e., produce $y$s for novel inputs $x$).
```

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,672 @@
...
+about it differently, this is actually just a nonlinear transformation $y=ML(x)$.
+The reason why the model $ML$ is interesting is because its form is basic but adapts to the
```
kmsquire: `The reason why` => `The reason that`

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,672 @@
...
+data itself. For example, a simple neural network (in design matrix form) with
+sigmoid activation functions is simply matrix multiplications followed
+by sigmoids, meaning $ML(x)=\sigma(W_{3}\cdot\sigma(W_{2}\cdot\sigma(W_{1}\cdot x)))$ is a three-layer deep
```
kmsquire: `by sigmoids, meaning` => `by application of a sigmoid function.  Specifically, `

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,672 @@
...
+[jump diffusion equations from financial models](http://docs.juliadiffeq.org/latest/tutorials/jump_diffusion.html),
+and the solvers map right over to the Flux.jl neural network frame work through FluxDiffEq.jl
+It is worth noting that FluxDiffEq.jl uses only around ~50 lines of code to pull this all off.
+
+## Implementing the Neural ODE layer in Julia
+
+Let's go all the way back for a second and now implement the neural ODE layer
+in Julia. Remember that this is simply an ODE where the derivative function
+is defined by a neural network itself. To do this, let's first define the
+neural net for the derivative. We can use a multilayer ...
```
kmsquire: Is this is meant to be filled in with more details?

ChrisRackauckas: yes. That's tomorrow :)

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -64,15 +64,18 @@ if you know the exact functional form that relates the input to the output.
 However, in many cases, such exact relations are not known *a priori*.
 So how do you do nonlinear modeling if you don't know the nonlinearity?
 
-There are two separate ways to then handle this problem. One method is machine
-learning. Usually this is presented as you have some $x$ and you want to predict
-$y$. This generation of a prediction from $x$ is the machine learning model, call it
-$ML$, where you put $x$ into the program and it spits out values $y$. But if you think
+There are two common ways to handle this problem. One method is to use machine
```
oxinabox: You still need to tell me both of them immediately.
Then you can go into details abotu each.

blog/_posts/2019-01-18-fluxdiffeq.md
ChrisRackauckas: well, today is tomorrow.

ChrisRackauckas:
The final section on neural ODEs is now complete. The last thing to add is an animation for training the neural ODE which will come very soon. Please get your comments and suggestions in. I plan on releasing this tomorrow, or at the latest Wednesday.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,823 @@
...
+There are multiple ways to do this. The most common is known as (adjoint) sensitivity
+analysis. Sensitivity analysis defines a new ODE whose solution gives the
+gradients to the cost function w.r.t. the parameters, and solves this secondary
+ODE. This is the method discussed in the neural ordinary differential equations
+paper, but actually dates back much further, and popular ODE solver frameworks
+like [FATODE](http://people.cs.vt.edu/~asandu/Software/FATODE/index.html),
+[CASADI](https://web.casadi.org/), and
+[CVODES](https://computation.llnl.gov/projects/sundials/cvodes)
+have been available with this adjoint method for a long time (CVODES came out
+in 2005!). [DifferentialEquations.jl has sensitivity analysis implemented too](http://docs.juliadiffeq.org/latest/analysis/sensitivity.html)
```
Vaibhavdixit02:
```suggestion
in 2005!). [DifferentialEquations.jl has sensitivity analysis implemented too](http://docs.juliadiffeq.org/latest/analysis/sensitivity.html)
```
```suggestion
in 2005!) and lastly [DifferentialEquations.jl provides extensive sensitivity analysis implementation](http://docs.juliadiffeq.org/latest/analysis/sensitivity.html)
```

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,822 @@
...
+There are two common ways to handle this problem. One method is to use machine
+learning. In a typical machine learning problem, you are given some input $x$ and
+you want to predict an output $y$. This generation of a prediction $y$ from $x$
+is a machine learning model (let's call it $ML$).  During training, we attempt to
+adjust the parameters of $ML$ so that it generates accurate predictions.  We
+can then use $ML$ for inference (i.e., produce $y$s for novel inputs $x$).
+But if you think
+about it differently, this is actually just a nonlinear transformation $y=ML(x)$.
+The reason $ML$ is interesting is because its form is basic but adapts to the
+data itself. For example, a simple neural network (in design matrix form) with
```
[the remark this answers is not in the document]
oxinabox: Design matrix is classical term in statistical regression though.
This is deeplearning folk being badies.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,822 @@
...
+impossible for libraries which utilize precompiled shared libraries that call
+C/C++/Fortran packages (SciPy, deSolve), and so this is a unique feature of the
+Julia differential equation suite. While entire ODE solver libraries could in
+theory be recoded into PyTorch or TensorFlow, doing so for a fully featured
+suite would be a difficult process without much promise for performance. For
+example, it is well-known that production-quality ODE solvers in
+C++/Fortran/Julia make heavy use of mutation in first class nonlinear functions
+to stave off the heavy memory pressure of allocating small arrays in a loop,
+but this kind of workflow is
+[mentioned in the documentation as not well-optimized in PyTorch](https://pytorch.org/docs/master/notes/autograd.html#in-place-operations-with-autograd). Other concerns of this fashion
```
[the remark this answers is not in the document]
MikeInnes: Yeah, I think you can make a much stronger case here by talking about different limitations – e.g. the fact that DE solvers have such a wide range of performance regimes and Python/TF do very badly on scalars.

blog/_posts/2019-01-18-fluxdiffeq.md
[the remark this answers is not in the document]
MikeInnes: We should just remove mentions of RNNs unless they are related to the continuous-time recurrence mentioned in the paper. A resnet is not actually a type of RNN anyway, and there's nothing here to show that people should be (or are) throwing out their LSTMs.

blog/_posts/2019-01-18-fluxdiffeq.md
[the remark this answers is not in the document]
MikeInnes: It's not a question of implementation, but why we should see a model like ResNet as being like solving an ODE in the first place. Most people aren't going to come in with that perspective, or have seen that paper, so it's worth pointing out that we can see ResNet as approximating something "infinitely deep". The idea that deeper models are more powerful gives an intuition for why this gives good results.

MikeInnes:
Before we publish let's make sure we run through all the code examples in a repl. Some of the examples are out of date right now (I'm fixing what I can).

blog/_posts/2019-01-18-fluxdiffeq.md
ChrisRackauckas: an extensive

blog/_posts/2019-01-18-fluxdiffeq.md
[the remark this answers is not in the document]
oxinabox: I almost sat down and did the math to show for a suitably general definition of a ResNet, every RNN can be cast as a ResNet.

But @jekbradbury  is 100% right NLP people have not heard of ResNet.

ChrisRackauckas:
"which is type of RNN" -> "which is a type of RNN", saw in an email and just making note

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,822 @@
...
+you want to predict an output $y$. This generation of a prediction $y$ from $x$
+is a machine learning model (let's call it $ML$).  During training, we attempt to
+adjust the parameters of $ML$ so that it generates accurate predictions.  We
+can then use $ML$ for inference (i.e., produce $y$s for novel inputs $x$).
+But if you think
+about it differently, this is actually just a nonlinear transformation $y=ML(x)$.
+The reason $ML$ is interesting is because its form is basic but adapts to the
+data itself. For example, a simple neural network (in design matrix form) with
+sigmoid activation functions is simply matrix multiplications followed
+by application of sigmoid functions. Specifically, meaning $ML(x)=\sigma(W_{3}\cdot\sigma(W_{2}\cdot\sigma(W_{1}\cdot x)))$ is a three-layer deep
```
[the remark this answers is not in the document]
ChrisRackauckas: yes

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,822 @@
...
+is defined by a neural network itself. To do this, let's first define the
+neural net for the derivative. In Flux, we can define a multilayer perceptron
+with 1 hidden layer and a `tanh` activation function like:
+
+```julia
+dudt = Chain(Dense(2,50,tanh),Dense(50,2))
+```
+
+To define a `neural_ode` layer, we then just need to give it a timespan and use
+the `neural_ode` helper function:
```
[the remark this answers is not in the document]
ChrisRackauckas: yes, whatever working makes sense to you here. I usually just use "helper function" as "package function you could've easily written yourself in a few lines"

blog/_posts/2019-01-18-fluxdiffeq.md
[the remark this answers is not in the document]
ChrisRackauckas: Mike, you take this then. You know the ML audience's missing knowledge of this domain better.

blog/_posts/2019-01-18-fluxdiffeq.md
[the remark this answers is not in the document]
jekbradbury: Yeah, but remember that people coming from PyTorch/TF/NumPy assume that if a function is provided in a package it’s because it would be too difficult to implement correctly themselves.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,823 @@
...
+to stave off the heavy memory pressure of allocating small arrays in a loop,
+but this kind of workflow is
+[mentioned in the documentation as not well-optimized in PyTorch](https://pytorch.org/docs/master/notes/autograd.html#in-place-operations-with-autograd). Other concerns of this fashion
+make Julia an exceptional choice for this application.
+
+Altogether, being able to switch between different gradient methods without changing
+the rest of your code is crucial for having a scalable, optimized, and
+maintainable framework for integrating differential equations and neural networks.
+And this is precisely what FluxDiffEq.jl gives the user direct access to. There
+are three functions:
```
kmsquire: `There are three functions:` => `There are three functions with a similar API:`

And remove `With a similar API.` below.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,823 @@
...
+of the nonlinear transformation. Thus instead of doing $y=ML(x)$, we put the
+machine learning model on the derivative, $y'(x) = ML(x)$, and now solve the ODE.
+Why would you ever do this? Well, one motivation is that, if you were to do this
+and then solve the ODE using the simplest and most error prone method, the
+Euler method, what you get is equivalent to a [residual neural network](https://arxiv.org/abs/1512.03385).
+The way the Euler method works is based on the fact that $y'(x) = \frac{dy}{dx}$, thus
+
+$$\Delta y = (y_\text{next} - y_\text{prev}) = \Delta x\cdot ML(x) \text{ which implies that } y_{i+1} = y_{i} + \Delta x\cdot ML(x_{i}).$$
+
+Looking familiar now? This is the basis of a residual network (ResNet) which is a type of recurrent neural network (RNN),
```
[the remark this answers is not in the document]
oxinabox: TO be fair I was working on off some a definition someone else showed me, and they may havebeen wrong or I may have misunderstood.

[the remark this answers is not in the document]
GunnarFarneback: As far as I know a ResNet is strictly feedforward and claiming that it is in some way recurrent would need a convincing motivation. As written it just risks the credibility of the post.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,823 @@
...
+x->neural_ode(dudt,x,tspan,Tsit5(),saveat=0.1)
+```
+
+As a side note, to run this on the GPU, it is sufficient to make the initial
+condition and neural network be on the GPU. This will cause the entire ODE
+solver's internal operations to take place on the GPU without extra data
+transfers in the integration scheme. This looks like:
+
+```julia
+x->neural_ode(gpu(dudt),gpu(x),tspan,BS3(),saveat=0.1)
```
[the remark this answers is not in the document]
ChrisRackauckas: Tsit5 doesn't work on the GPU because no one has fixed v1.0 broadcast regressions yet... :( it literally just needs broadcast added to it but we can't because it would slow it down.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,823 @@
...
+computing the derivatives of differential equations with respect to parameters.
+We have a [recent preprint](https://arxiv.org/abs/1812.01892) detailing
+some of these results. One of the things we have found is that direct use of
+automatic differentiation can be one of the most efficient and flexible methods.
+Julia's ForwardDiff.jl, Flux.jl, and ReverseDiff.jl can directly be applied to
+perform automatic differentiation on the native Julia differential equation
+solvers themselves, and this can increase performance while giving new features.
+Our findings show that forward-mode automatic differentiation is fastest when
+there are less than 100 parameters in the differential equations, and that for
+a large enough number of parameters adjoint sensitivity analysis is the most efficient. Even
```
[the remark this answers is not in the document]
ChrisRackauckas: I put 100 in the sentence before.

blog/_posts/2019-01-18-fluxdiffeq.md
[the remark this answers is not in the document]
jekbradbury: The paper mentions RNNs but it takes a fairly "DL theory" perspective on what that means. I agree that we should leave it out and focus on the ResNet comparison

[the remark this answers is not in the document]
GunnarFarneback: If I remember correctly it argued that differential equations could model both ResNet and RNN but not that those two would be otherwise related. I also think focusing on ResNet is a good plan.

ChrisRackauckas:
neural_ode_predict

neural_ode_ts_fit

blog/_posts/2019-01-18-fluxdiffeq.md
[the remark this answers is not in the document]
ChrisRackauckas: `# Currently Tsit5 is purposely not GPU compatible due to Julia v1.0 bug` is that good wording?

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,742 @@
...
+# Display the ODE with the initial parameter values.
+cb()
+
+Flux.train!(loss_n_ode, ps, data, opt, cb = cb)
+```
+
+![Neural ODE Train](https://user-images.githubusercontent.com/1814174/51585825-dc3f8480-1ea8-11e9-8498-18cf55fba3e6.gif)
+
+Notice that what's not being learned is the solution to the ODE.
+Instead, what's learned is the tiny ODE system from which the ODE
```
oxinabox:
```suggestion
Instead, what we are learning is the tiny ODE system from which the ODE
```

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,742 @@
...
+solution is generated. I.e., the neural network inside the neural_ode
+layer has learned this function:
```
oxinabox:
```suggestion
layer has learns this function:
```

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -10,12 +10,13 @@ robustly add differential equation solvers to neural networks in Julia.
 ![Flux ODE Training Animation](https://user-images.githubusercontent.com/1814174/51399500-1f4dd080-1b14-11e9-8c9d-144f93b6eac2.gif)
 
 The [Neural Ordinary Differential Equations](https://arxiv.org/abs/1806.07366)
-paper has been a hot topic even before it made a splash as Best Paper of NeurIPS
-2018. Among other demonstrations, combining these two disparate fields enabled
-the authors to reduce the number of parameters and reduce memory costs in
-comparison to ResNet. But this is only the beginning: neural networks and
-differential equations were born to be together. This blog post will explain
-why, and start to give a sense of what's possible with state-of-the-art tools.
+paper, by Jesse Bettencourt and collaborators, has been a hot topic even before it
```
oxinabox: No, we can't do this.
It is against academic convention to refer to the paper as _by 3rd author and collaborators_ . 

The paper is listed as by
Ricky T. Q. Chen, Yulia Rubanova, Jesse Bettencourt, David Duvenaud

we thus need to say either

by _Ricky T. Q. Chen and collaborators_

or 

_by Ricky T. Q. Chen, Yulia Rubanova, Jesse Bettencourt and David Duvenaud_

MikeInnes: This is not actually a paper, so it's less important for us to follow that kind of convention unless it's likely to cause frustration for other reasons. I'll try and reword it though.

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -284,7 +272,7 @@ p = param([2.2, 1.0, 2.0, 0.4]) # Initial Parameter Vector
 params = Flux.Params([p])
 
 function predict_rd() # Our 1-layer neural network
-  diffeq_rd(p,reduction,prob,Tsit5(),saveat=0.1)
+  diffeq_rd(p,prob,Tsit5(),saveat=0.1)
```
ChrisRackauckas: this stuff still needs to do the reduction afterwards or it won't return an array and I'm not sure the code would work.

MikeInnes: So `diffeq_rd(...)[1,:]`?

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,742 @@
...
+As a side note, to run this on the GPU, it is sufficient to make the initial
+condition and neural network be on the GPU. This will cause the entire ODE
+solver's internal operations to take place on the GPU without extra data
+transfers in the integration scheme. This looks like[^gpu]:
+
+```julia
+x->neural_ode(gpu(dudt),gpu(x),tspan,BS3(),saveat=0.1)
+```
+
+[^gpu]: We have changed the solver from Tsit5 to BS3, as Tsit5 is currently not GPU compatible.
```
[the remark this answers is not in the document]
MikeInnes: That's why I said "currently". Maybe we should add the link over [currently not GPU compatible] so it's more heavily implied.

blog/_posts/2019-01-18-fluxdiffeq.md
MikeInnes: I'm not totally clear if the slicing does something important here, but conversion to a vector isn't inherently necessary for the AD to work here, so it might still be possible to remove it in some cases.

blog/_posts/2019-01-18-fluxdiffeq.md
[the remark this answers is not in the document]
ChrisRackauckas: This is a delicate subject because then it sounds like it's hard for us to make this GPU compatible. It isn't, we already did it, and we had to revert the PR since v1.0 broadcasting is 3x slower (https://github.com/JuliaLang/julia/issues/28126) and users immediately reported the regression. So we are purposefully choosing to not support GPUs on this algorithm right now because of that, BUT we could add it any time in less than 20 minutes since the code with broadcast is already there and commented out.

How do we politely say this in a way that conveys "okay yes, they essentially have GPU compatibility on everything".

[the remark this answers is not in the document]
MikeInnes: ``
```suggestion
[^gpu]: We have changed the solver from Tsit5 to BS3, as Tsit5 is [currently not GPU compatible]((https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/106)).
```

blog/_posts/2019-01-18-fluxdiffeq.md
```diff
@@ -0,0 +1,741 @@
...
+their complementary ways of describing a nonlinear world. In the Julia ecosystem
+we have merged the differential equation and deep learning packages in such a
+way that new developments in the two domains can directly be used together.
+We are only beginning to understand the possibilities that have opened up with
+this software. We hope that future blog posts will detail some of the cool
+applications which mix the two disciplines, such as embedding our coming
+pharmacometric simulation engine [PuMaS.jl](https://doi.org/10.1007/s10928-018-9606-9)
+into the deep learning framework. With access to the full range of solvers for ODEs,
+SDEs, DAEs, DDEs, PDEs, discrete stochastic equations, and more, we are
+interested to see what kinds of next generation neural networks you build with Julia.
```
oxinabox:
```suggestion
interested to see what kinds of next generation neural networks you will build with Julia.
```

ChrisRackauckas:
Following @ViralBShah 's suggestion, we put the runnable scripts in the model-zoo repo and will link to them after each animation to make it easier to follow along with the post. @MikeInnes will get that merged in a sec, add the links here, and then we are ready to release. So let's say 10:30-11? Any last comments need to get in by then, or at least highlight a previous discussion.

blog/_posts/2019-01-18-fluxdiffeq.md
ChrisRackauckas: It's actually going from values in different vectors in a vector of vectors to a single contiguous vector for the time series of 1 variable. So it's not your average slice. It's not important for AD, but if someone puts the layer function in say the `Chain` code and then assumes linear algebra is going to work (putting a Dense right after), they will get an error.

fredrikekre:
The title seems to be missing from the rendered version.

ChrisRackauckas:
It displays the info table at the top differently than the others:

https://github.com/JuliaLang/www.julialang.org/blob/master/blog/_posts/2019-01-18-fluxdiffeq.md

https://github.com/JuliaLang/www.julialang.org/blob/master/blog/_posts/2018-12-03-ml-language-compiler.md

Anyone know why?

MikeInnes:
I just fixed it, hopefully it'll show up soon.

jekbradbury:
The LaTeX isn’t showing up for me in mobile Safari
73df2982-1e94-41b4-8bac-0bc32362b86c
I think you need to include MathJax at the top of the post, like [this](url
https://github.com/JuliaLang/www.julialang.org/blob/master/blog/_posts/2017-09-06-gsoc-derivative_operators.md)?

Vaibhavdixit02:
The LateX doesn't work in desktop Safari as well

ChrisRackauckas:
https://github.com/JuliaLang/www.julialang.org/commit/49ba1b24889bdcebf71e418e1a0cdb3d085b1a56 should fix it when the post updates

ChrisRackauckas:
we should be good. Still taking any hotfixes as they come, but it's good to post now I think

fredrikekre:
I think 49ba1b24889bdcebf71e418e1a0cdb3d085b1a56 broke this:
screenshot from 2019-01-23 17-22-33

AzamatB:
In some formulas derivative sign is not typeset correctly. For example:
screen shot 2019-01-25 at 11 55 07 am

oxinabox:
Probably need to switch `'` for `^\prime`

PR #275, opened 2019-01-18
Jesse Edits for FluxDiffEq Post

jessebett:


jessebett:
"...Euler method, what you get is equivalent to a recurrent neural network" is incorrect as there aren't per timestep inputs/outputs like an RNN. Resnet is the correct analogy.

PR #283, opened 2019-01-22
Fix buttons linking to newest julia version.

englhardt:
The buttons on the index homepage still show `v1.0` instead of `v1.1`.


englhardt:
PR #281 partly addresses the same issue but does not fix the footer.

PR #314, opened 2019-02-21
Add magma binding project to GSoC

Roger-luo:
As we discussed in Beijing. Hope it's not too late.

cc: @ViralBShah


ViralBShah:
No it's not

PR #321, opened 2019-03-02
Add primitive type tutorial blog

yuehhua:
With English and Traditional Chinese version.


ViralBShah:
We need some reviews.

yuehhua:
Sure.

blog/_posts/2019-03-02-custom-primitiv-type.md
```diff
@@ -0,0 +1,383 @@
+---
+layout: post
+title: Declare primitive type in Julia
+author: 鄭景文（Peter Cheng）, 杜岳華（Yueh-Hua Tu）
+---
+
+Julia provide some APIs for you to declare your own primitive types in Julia. However, when you open [the official docs about primitive type](https://docs.julialang.org/en/v1/manual/types/#Primitive-Types-1), you still don't know how to make your new primitive types work in your code. Therefore, here is a simple tutorial that tell you how to do so.
```
matbesancon:
```julia
Julia provides
```

No need to repeat "in Julia" at the end :)

blog/_posts/2019-03-02-custom-primitiv-type.md
```diff
@@ -0,0 +1,383 @@
...
+
+
+## Introduction
+
+If you want define a primitve type, first, you have to declare it with the name and the number of bits it take with the `primitive type` keyword. 
+
+    primitive type «name» «bits» end
+    primitive type «name» <: «supertype» «bits» end
+
+Currently, Julia only support sizes that are multiples of 8 bits. So like the `Bool` only take 1 bit, you still need to declare it as 8 bits. For example:
```
matbesancon: `Julia only supports`
So even though `Bool` only takes 1 bit, you need to declare it as 8 bits

StefanKarpinski:
In the URL: `primitiv` => `primitive`.

chengchingwen:
There is an English version included. We wrote in both Chinese and English

blog/_posts/2019-03-02-custom-primitiv-type.md
```diff
@@ -0,0 +1,383 @@
...
+
+```julia
+    primitive type Bool <: Integer 8 end
+```
+
+Up to here, we have almost everything from the docs. Next, we are going to show how 
+to use our own primitive type with a simple example.
```
matbesancon:
```suggestion
to define and use our own primitive type with a simple example.
```

blog/_posts/2019-03-02-custom-primitiv-type.md
```diff
@@ -0,0 +1,383 @@
...
+
+## The problem
+
+After declaring our new primitive type, that's say we do the following:
+
+```julia
+    primitive type A 72 end
+```
+
+But soon you will find that you don't know how to create a instance of `A` because Julia don't have a default constructor for a new primitive type. And if you think further, you will notice that there seems to be no methods for you to manipulate those instances. You can't even do bitwise addition becuase `+` donsn't work on our new primitive type.
```
matbesancon:
```suggestion
But soon you will find that you don't know how to create a instance of `A` because Julia does not have a default constructor for primitive types. And if you think further, you will notice that there seems to be no methods for you to manipulate those instances. You can't even do bitwise addition becuase `+` does not work on our new primitive type.
```

blog/_posts/2019-03-02-custom-primitiv-type.md
```diff
@@ -0,0 +1,383 @@
...
+
+Then what can we do? Well, Julia do have functions for you to do those stuff, but most of those function are functions that generated by Julia's C++ source code or from the LLVM. So, before we started, we will take a first look at those functions lying under `Core.Intrinsics`.
```
matbesancon:
```suggestion
Then what can we do? Well, Julia has functions for you to do those stuff, but most of those function are generated by Julia's C++ source code or from LLVM. So, before we started, we will take a first look at those functions lying under `Core.Intrinsics`.
```

blog/_posts/2019-03-02-custom-primitiv-type.md
```diff
@@ -0,0 +1,383 @@
...
+
+a1 + a2 # add the value of a1 and a2
+```
+
+
+## Q&A
+
+Q. What's the benefit of using a custom type? Why not just define a wrapper type of type alias of a built-in primitive type?
+A. 
+First, if you want to dispatch on the type, than type alias might not work since it is actually the same type as the origin.
```
matbesancon:
```suggestion
First, if you want to dispatch on the type, then a type alias will not work since it is actually the same type as the aliased one. 
```

blog/_posts/2019-03-02-custom-primitiv-type.md
```diff
@@ -0,0 +1,383 @@
...
+
+Second, a wrapper type will cause some performance drop owing to the unwrapping process and the compiler might not be able to optimise the code. compare the result of `@code_llvm a1 + a2` on the above example and a wrapper type.
+
+Last but not the least, You can define a custom bitsize for your primitive type and allow LLVM to do the magic.
+
+Q. Where can I find more information?
+A. you can take a look at the [LLVM reference](https://llvm.org/docs/LangRef.html) for the LLVM Intrinsics and the [Julia c++ source code for Intrinsics](https://github.com/JuliaLang/julia/blob/80516ca20297a67b996caa08c38786332379b6a5/src/intrinsics.cpp#L853-L1241) and [header file](https://github.com/JuliaLang/julia/blob/v1.1.0/src/intrinsics.h) for function arguments, more detail at the [codegen](https://github.com/JuliaLang/julia/blob/v1.1.0/src/codegen.cpp) and [ccall](https://github.com/JuliaLang/julia/blob/v1.1.0/src/ccall.cpp), and Julia code example in [Julia's test](https://github.com/JuliaLang/julia/blob/v1.1.0/test/intrinsics.jl) and the [defination of built-in primitive type](https://github.com/JuliaLang/julia/blob/v1.1.0/base/boot.jl).
```
matbesancon:
```suggestion
A. you can take a look at the [LLVM reference](https://llvm.org/docs/LangRef.html) for the LLVM Intrinsics and the [Julia C++ source code for Intrinsics](https://github.com/JuliaLang/julia/blob/80516ca20297a67b996caa08c38786332379b6a5/src/intrinsics.cpp#L853-L1241) and [header file](https://github.com/JuliaLang/julia/blob/v1.1.0/src/intrinsics.h) for function arguments, more detail at the [codegen](https://github.com/JuliaLang/julia/blob/v1.1.0/src/codegen.cpp) and [ccall](https://github.com/JuliaLang/julia/blob/v1.1.0/src/ccall.cpp), and Julia code example in [Julia's test](https://github.com/JuliaLang/julia/blob/v1.1.0/test/intrinsics.jl) and the [defination of built-in primitive type](https://github.com/JuliaLang/julia/blob/v1.1.0/base/boot.jl).
```

blog/_posts/2019-03-02-custom-primitiv-type.md
```diff
@@ -0,0 +1,383 @@
...
+
+
+## Bitboard example
+
+[bitboard](https://en.wikipedia.org/wiki/Bitboard) is a way of store and compute board game state in binary expression, for exampe we want to store a 4x4 board for a [2048 game](https://en.wikipedia.org/wiki/2048_(video_game)) we can do like this:
```
matbesancon:
```suggestion
A [bitboard](https://en.wikipedia.org/wiki/Bitboard) is a way of store and compute board game state in binary expression, for exampe we want to store a 4x4 board for a [2048 game](https://en.wikipedia.org/wiki/2048_(video_game)) we can do like this:
```

blog/_posts/2019-03-02-custom-primitiv-type.md
```diff
@@ -0,0 +1,383 @@
...
+<a id="reference"></a>
+
+## Core.Intrinsics Reference
+
+reference for some basic function under the `Core.Intrinsics`.
+1.  Type construction
+    
+    1.  `Core.Intrinsics.bitcase(type, value)`
+    
+        `bitcase` told the LLVM to make a new same sized value of a primitive type instance `value`         and mark it as type `type`. The type of `value` must have the same bitsize as `type`.
```
matbesancon: why these spaces between `value` and "and". Same applies to next lines

chengchingwen: Oh, I guess it's due to some problems in my editor, let me remove it.

blog/_posts/2019-03-02-custom-primitive-type.md
```diff
@@ -0,0 +1,382 @@
...
+    primitive type Bool <: Integer 8 end
+```
+
+Up to here, we have almost everything from the docs. Next, we are going to show how 
+to define and use our own primitive type with a simple example.
+
+
+## The problem
+
+After declaring our new primitive type, that's say we do the following:
```
matbesancon:
```suggestion
After declaring our new primitive type, that is to say we do the following:
```

chethega:
Do we really want to document/blog about this as long as
```
julia> primitive type B 24 end
julia> B(i)=Core.Intrinsics.trunc_int(B,Int128(i));

julia> [B(i) for i=1:100]
signal (11): Segmentation fault
```
aka https://github.com/JuliaLang/julia/issues/29193 https://github.com/JuliaLang/julia/issues/29053 https://github.com/JuliaLang/julia/issues/26026 ?

I'd feel much better if we deferred documentation / public blogging / advertisement of primitive types until they actually work reliably.

chethega:
FWIW, the same issue appears with your bitsize, just harder to spot. Example:
```
julia> primitive type A 72 end
julia> A(i)=Core.Intrinsics.zext_int(A,i);
julia> arr=fill(A(1), 10^7);
ERROR: ReadOnlyMemoryError()
julia> arr=[A(i) for i=1:30]
30-element Array{A,1}:
 A(0x000000000000000001)
 A(0x000000000000000002)
 A(0x000000000000000003)
 A(0x000000000000000004)
 A(0x000000000000000005)
 A(0x000000000000000006)
 A(0x000000000000000007)
 A(0x000000000000000008)
 A(0x000000000000000009)
 A(0x00000000000000000a)
 ⋮                      
 A(0x600000000000000023)
 A(0xc000007fa6e4335630)
 A(0x020000000000000001)
 A(0x030000000000000023)
 A(0x050000000000000004)
 A(0x070000000000000006)
 A(0x090000000000000008)
 A(0x0b000000000000000a)
 A(0x0d000000000000000c)
```
The issue is that everybody has different opinions on whether the size of A is 9 byte or 16 byte (8 byte + 1 byte + 7 bytes padding for alignment). Hence the allocated buffer is sometimes too small and we corrupt memory. Also, `arrayref`, `unsafe_load(pointer(arr, i))`, `unsafe_load(pointer(arr), i))` are inconsistent in computing offsets, and I think arrayref is inconsistent between interpretation (REPL) and compiled code.

yuehhua:
@chethega Thank you for telling us these issues.
Maybe we could hold on the PR until the issues are solved.
@chengchingwen How do you think?

chengchingwen:
Ah, so that's the reason why there is no documentation yet. @yuehhua I'm fine with it, but after that we might need to change the content a little bit.

ViralBShah:
Bump. Is this happening?

chengchingwen:
@ViralBShah As @chethega mentioned, there are still some bugs of primitive type. Should we just ignore those bugs, or should we close this PR for now?

PR #323, opened 2019-03-05
removed University of Sydney course

xiaodaigh:
Dr Olver is no longer at the University of Sydney. See http://wwwf.imperial.ac.uk/~solver/


ViralBShah:
This is a course that was taught in Julia that year and it forwards to his new website. I think this is fine as is.

Cc'ing @dlfivefifty

PR #326, opened 2019-03-07
Separate graphs in the GSoC part

matbesancon:


jpfairbanks:
To follow up on your slack question, I think that having Graphs as a top level category is good. JuliaGraphs is as broad a scope as "Images" or "DifferentialEquations"

matbesancon:
That's also what made me do it this way

PR #335, opened 2019-03-21
try to show all authors on the debugger blogpost

KristofferC:


timholy:
Seems fine to me, I don't know what I'm doing. Is there some place one can see the build process?

KristofferC:
Webstuff is typically debugged in production I think. Lookgs good now.

ViralBShah:
I can never figure out where the github pages build logs are, and for the life of me, I can't install the jekyll environment for our website on my computer. So, I end up doing website stuff in production.

I believe it is much easier to set up the environment on linux than on a mac.

PR #336, opened 2019-03-21
fix list syntax in debuggers blog

fredrikekre:
Starting a line with a number and a dot is interpreted as a numberd list


fredrikekre:
list

Issue #69, opened 2018-08-03
Tiny Julia logo resizing issue


ViralBShah:
https://twitter.com/hrdwrknvrstps/status/1025442923180175360

ViralBShah:
@essenciary Would be good to fix this one. Happens for me too.

essenciary:
Pushed fix

Keno:
<img width="564" alt="screen shot 2019-01-04 at 6 07 07 pm" src="https://user-images.githubusercontent.com/1291671/50859405-44e31900-1361-11e9-8c66-77500cbcba88.png">

I still see this

Keno:
![screen shot 2019-01-08 at 4 20 01 pm](https://user-images.githubusercontent.com/1291671/50859421-54626200-1361-11e9-9a53-4eb02123fe69.png)

I still see this

essenciary:
Hmmm, odd, it was properly fixed for sure. OK, lmk.

ViralBShah:
Actually, I decided against changing the menu. Is it possible to fix this?

essenciary:
@ViralBShah ok, I figured out what it was, cause I _knew_ I fixed it. It's because of the GSoC extra menu item. Basically now the menu doesn't fit on smaller screens. So yeah, there's no room for everything. If I keep the size of the logo, it looks like this: 

<img width="1042" alt="image" src="https://user-images.githubusercontent.com/11292378/52068439-f5b59000-257c-11e9-9499-f8550caa2713.png">

ViralBShah:
It just occurred to me yesterday that it was most likely due to the GSoC. Would reducing the logo size help?

essenciary:
Indeed, it was the GSoC item. I commented in the other issue (one video). I tweaked it a bit so for smaller resolutions, both the spacing between menu items and the logo become smaller. Looks good now.

PR #341, opened 2019-04-14
Change to JSoC

ChrisRackauckas:
As the Julia Language is growing, it's moving away from just having Google Summer of Code to now participating in multiple programs (NumFOCUS Small Development Grants, and Google Season of Docs). This is a WIP take on what this change would look like. The GSoD section was added on the top of the now renamed Programs webpage since it will be the next program for which applications are due (and we can keep cycling this). I added a DifferentialEquations project example for season of docs (these projects would likely move to their own page). I added a stub for NumFOCUS Small Development Grants, change the name of the "GSoC administrators" to the "Julia Language Programs Committee" and suggested a new email.

The issue though is that, "programs" has a bad double meaning here... but it is the word that NumFOCUS uses.


simonschoelly:
The NumFOCUS page has the advantage that "programs" is a mouse over button, which shows the subtopics. This kind of helps with discoverability.

Some other ideas for names, although I don't think any of them is much better:
* participate
* volunteering
* opportunities

ViralBShah:
These changes are all generally good. The one thing to check is that whatever the new menu item text is and resize the browser ensuring the julia logo doesn't become tiny. Basically watch out for #69

ViralBShah:
The name Programs could be confusing.  Maybe just refer to it as JSoC? Julia Season of Contributions? But it has the same acronym as Summer of Code.

matbesancon:
> The name Programs could be confusing. Maybe just refer to it as JSoC? Julia Season of Contributions? But it has the same acronym as Summer of Code.

"Contribution projects"?

soc/ideas-page.md
```diff
@@ -1,22 +1,56 @@
...
+The Julia organization is a participant in Google Summer of Code. The Julia organization generally fields a large number of students (>20) mentored by open source developers and university professors at leading institutions from all around the world. Students who are accepted into this prestigious program are paid a summer salary to contribute to the open source tooling around Julia and its package ecosystem.
 
 This page lists a bunch of project ideas, meant to serve as starting points as you develop your own plan. You're free to modify them to your tastes, or propose something totally different in an area you're interested in.
 
 Check out the [application guidelines](guidelines/) once you have a project in mind.
 
 **Table of Contents**
 
-We have our project ideas organised roughly into the skillsets required:
+We have our project ideas organized roughly into the skillsets required:
```
MikeInnes: boo

MikeInnes:
"Programs" is fine in context but confusing as a stand-alone menu item. I think "Funded Projects", "Summer Projects" or "Student Projects" gets the gist across better. "JSoC" would also be fine as it's recognisable for people looking for GSoC.

ViralBShah:
I think either JSoC or Projects.

ChrisRackauckas:
Alright, given the feedback, I changed the tab to JSoC for Julia Seasons of Contributions. This will be a running title for all of the contributions programs, including GSoC, GSoD, and various development grants handled through the organization. The email for the committee is jsoc@julialang.org (Viral, can we get this made?). Then we keep the same #jsoc channel in the Slack as well.

I also sectioned out the GSoD projects, and after this PR we should get Flux and others to submit some project ideas so we can get our GSoD submission in. Any other things to discuss?

PR #352, opened 2019-04-24
citation paragraph

### LINKED ISSUE #304: bibtex for citations
It would be handy to be able to give bibtex output for the citations. One option is to have a [bibtex] link for each citation which would expand to give the full bibtex format. This could either be autogenerated via the citation template (https://github.com/JuliaLang/www.julialang.org/blob/master/_includes/citation.html), or add a bibtex field to the YAML to allow it to be overridden (e.g. to deal with non-ASCII characters, etc.).

matbesancon:
Fixes #304

### DIFF
--- a/research/index.html
+++ b/research/index.html
@@ -21,6 +21,10 @@ redirect_from:
 <div>{% include citation.html pub=beks17 %}</div>
 </p>
 
+<p>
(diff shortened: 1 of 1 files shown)


matbesancon:
ok maybe not fixing #304 stricto sensu since we're not adding the citation bib directly but linking to it on the main repo

ViralBShah:
Added a couple of lines for packages to also adopt this: d77f3583e39e6e61840a082351c196fc8085740a

ViralBShah:
Probably should go into documentation for package developers and discourse.

PR #356, opened 2019-05-05
Add ARM binaries to downloads page

ararslan:
Fixes https://github.com/JuliaLang/julia/issues/30870.

### DIFF
--- a/downloads/index.html
+++ b/downloads/index.html
@@ -48,6 +48,112 @@ title:  Julia Downloads
   <p>
     Please see <a href="platform.html">platform specific instructions</a> for further installation instructions and if you have trouble installing Julia.
   </p>
+
+  <div class="row">
+    <div class="col-12">
+      <h2>Currently supported platforms</h2>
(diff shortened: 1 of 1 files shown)


ViralBShah:
Looks great. I think eventually we want to maintain the tiers only on one place, but let's merge away for now.

jekbradbury:
There's a decent amount of scrolling needed to find the actual links now. I wonder if you'll see a (tiny) drop in the download stats...

ararslan:
I would be surprised if anyone would give up on downloading altogether due to scrolling, but you're right, this introduces a nontrivial amount of non-links to a page people would go to for links. Perhaps we should look at how other projects structure their downloads page while making levels of support clear.

staticfloat:
Thanks Alex!

ViralBShah:
I moved the supported platforms to the bottom. I think it works well there. What we need is a link on the top (next to platform-specific instructions to the tiers table at the bottom).

PR #366, opened 2019-05-09
Update index.html

musm:
Can we also remove the extensions?


ViralBShah:
This is good. But let's keep the extensions.

PR #367, opened 2019-05-09
DiffEqFlux: `BS3` -> `Tsit5` and remove footnote

YingboMa:
Now, Tsit5 can run on a GPU, too.

CC: @ChrisRackauckas


ChrisRackauckas:
Should the Chinese translation get updated as well? Or will that happen automatically? I'm not sure how that was done.

PR #399, opened 2019-06-25
update VSCode Extension URL

pfitzseb:


ViralBShah:
The documentation links to a docs page that has no content.

PR #411, opened 2019-07-23
Add downloads for 1.3.0-alpha

ararslan:
No macOS yet ;-;


ararslan:
I think we should wait to merge until we have macOS binaries available given that macOS users constitute a significant portion of the user base.

PR #416, opened 2019-07-29
community standards: clarify that it applies to verbal communication too

StefanKarpinski:
This was written long pre-JuliaCon when written communication was all we ever did. These days it's entirely plausible that someone might be speaking, rather than writing, in a public forum which is part of the Julia community.


StefanKarpinski:
I will let this sit for 24 hours in case there are any objections (hard to imagine) and then merge since this seems to have sufficient community support.

StefanKarpinski:
I think it's safe to merge this now.

PR #417, opened 2019-08-01
Artifacts blog post

staticfloat:
This is a first draft of an Artifacts blog post that I wrote up, since there are many, many moving parts.  There's still more to write here, and I think it might help to have some diagrams as well, so don't expect to merge this anytime soon.


blog/_posts/2019-08-01-artifacts.md
```diff
@@ -0,0 +1,179 @@
...
+```julia
+dependencies = [
+    "https://github.com/JuliaBinaryWrappers/c_simple_jll.jl/releases/download/c_simple%2Bv1.2.3%2B7/build_c_simple.v1.2.3.jl",
+]
+```
+
+You now simply write:
+```julia
+dependencies = [
+    "c_simple_jll",
```
KristofferC: Is there a need for a UUID here?

blog/_posts/2019-08-01-artifacts.md
```diff
@@ -0,0 +1,179 @@
...
+
+The `processed_output` artifact contains no `download` stanza, and so cannot be installed.  An artifact such as this would be the result of code that was previously run, generating a new artifact and binding the resultant hash to a name within this project.
+
+## Working with artifacts
+
+Artifacts can be manipulated using convenient APIs exposed from the `Pkg.Artifacts` namespace.  As a motivating example, let us imagine that we are writing a package that needs to load the [Iris machine learning dataset](https://archive.ics.uci.edu/ml/datasets/iris).  While we could download the dataset to a file within the package directory, this is undesirable because it (a) modifies state within a package directory that may be read-only and (b) is not shared across multiple versions of our package.  Instead, we will first check to see if our `iris` artifact exists on-disk and if it does not, we will create it by downloading the appropriate file, then bind the result into our `Artifact.toml` file:
```
StefanKarpinski:
```suggestion
Artifacts can be manipulated using convenient APIs exposed from the `Pkg.Artifacts` namespace.  As a motivating example, let us imagine that we are writing a package that needs to load the [Iris machine learning dataset](https://archive.ics.uci.edu/ml/datasets/iris). While we could just download the dataset during a build step into the package directory, and many packages currently do precisely this, that has some significant drawbacks. First, it modifies the package directory, making package installation stateful, which we'd like to avoid. In the future, we would like to reach the point where packages can be installed completely read-only, instead of being able to modify themselves after installation. Second, the downloaded data is not shared across different versions of our package. If we have three different versions of the package installed for use by various projects, then we need three different copies of the data, even if it hasn't changed between those versions. Moreover, each time we upgrade or downgrade the package, unless we do something clever (and probably brittle), we will have to download the data again.  With artifacts, we will instead check to see if our `iris` artifact already exists on-disk and only if it doesn't will we download and install it, after which we can bind the result into our `Artifact.toml` file:
```

blog/_posts/2019-08-01-artifacts.md
```diff
@@ -0,0 +1,179 @@
...
+
+# BinaryBuilder.jl
+
+You may be able to guess that `BinaryBuilder.jl` knows how to generate `Artifact.toml` files ([example](https://github.com/JuliaBinaryWrappers/c_simple_jll.jl/blob/master/Artifact.toml)), but that's not the only change.  We were tired of the current best practice of needing to manually express the graph of all dependencies within your package's `deps/build.jl` file.  To date, there has not been a simple, concise method of recursively installing your binary dependencies, users have been forced to resort to strategies such as embedding the `build.jl` files of all binary dependencies within their own packages.  This works, but it's way clunkier than we'd like.  Luckily, we already have a package manager that knows how to deal with recursive dependencies, and so a simple solution was devised; as a part of the output of a `BinaryBuilder.jl` run, we will generate a wrapper julia package that will simultaneously allow us to express the DAG of binary dependencies, as well as provide boilerplate julia wrapper code that will make dealing with libraries and executables much simpler.  We refer to these autogenerated packages as "jll" packages.
+
+## jll packages
```
StefanKarpinski:
```suggestion
## Julia library (jll) packages
```

blog/_posts/2019-08-01-artifacts.md
```diff
@@ -0,0 +1,179 @@
...
+
+The autogenerated packages are true Julia packages, with the notable inclusion of `Artifact.toml` files that download the appropriate version of whatever binary artifacts were built by `BinaryBuilder` and uploaded to the repository's GitHub release.  BinaryBuilder attempts to upload all packages to `JuliaBinaryWrappers/$(package_name)_jll.jl`, however this is, of course, configurable.  An example package [is given here](https://github.com/JuliaBinaryWrappers/c_simple_jll.jl), the most interesting part being the new API that is exposed by these autogenerated packages.
```
StefanKarpinski:
```suggestion
The autogenerated packages are normal Julia packages, with the notable inclusion of `Artifact.toml` files that download the appropriate version of whatever binary artifacts were built by `BinaryBuilder` and uploaded to the repository's GitHub release.  BinaryBuilder attempts to upload all packages to `JuliaBinaryWrappers/$(package_name)_jll.jl`, however this is, of course, configurable.  An example package [is given here](https://github.com/JuliaBinaryWrappers/c_simple_jll.jl), the most interesting part being the new API that is exposed by these autogenerated packages.
```

blog/_posts/2019-08-01-artifacts.md
```diff
@@ -0,0 +1,179 @@
...
+git-tree-sha1 = "1c223e66f1a8e0fae1f9fcb9d3f2e3ce48a82200"
+```
+
+This `Artifact.toml` binds three artifacts; one named `socrates`, one named `c_simple` and one named `processed_output`.  The single required piece of information for an artifact is its `git-tree-sha1`.  Because artifacts are addressed only by their content hash, the purpose of an `Artifact.toml` file is to provide metadata about these artifacts, such as binding a human-readable name to an appropriate content hash, providing information about where an artifact may be downloaded from, or even binding a single name to multiple hashes, keyed by platform-specific constraints such as operating system or libgfortran version.
+
+## Artifact types and properties
+
+In the above example, the `socrates` artifact showcases a non-platform-specific artifact with multiple download locations.  When downloading and installing the `socrates` artifact, URLs will be attempted in-order until one succeeds.  The `socrates` artifact is marked as `lazy`, which means that it will not be automatically downloaded when the containing package is installed, but rather will be downloaded on-demand when the package first attempts to use it.
```
StefanKarpinski:
```suggestion
In the above example, the `socrates` artifact showcases a platform-independent artifact with multiple download locations.  When downloading and installing the `socrates` artifact, URLs will be attempted in-order until one succeeds.  The `socrates` artifact is marked as `lazy`, which means that it will not be automatically downloaded when the containing package is installed, but rather will be downloaded on-demand when the package first attempts to use it.
```

blog/_posts/2019-08-01-artifacts.md
```diff
@@ -0,0 +1,179 @@
...
+
+The `c_simple` artifact showcases a platform-specific artifact, where each entry in the `c_simple` array contains keys that help the calling package choose the appropriate download based on the particulars of the host machine.  Note that each artifact contains both a `git-tree-sha1` and a `sha256` for each download entry.  This is to ensure that the downloaded tarball is secure before attempting to unpack it, as well as enforcing that all tarballs must expand to the same overall tree hash.
```
StefanKarpinski:
```suggestion
The `c_simple` artifact showcases a platform-dependent artifact, where each entry in the `c_simple` array contains keys that help the calling package choose the appropriate download based on the particulars of the host machine.  Note that each artifact contains both a `git-tree-sha1` and a `sha256` for each download entry.  This is to ensure that the downloaded tarball is secure before attempting to unpack it, as well as enforcing that all tarballs must expand to the same overall tree hash.
```

blog/_posts/2019-08-01-artifacts.md
staticfloat: As I was transitioning from the dream state to consciousness this morning, I realized that users will want the ability to put anything they could put into `pkg> add` here.  E.g. they'll want to be able to say `c_simple_jll@v1.2.3`, and also `https://privategitlabinstance/my/private/project_jll.git@v1.2.3`.  So perhaps the only foolproof way to deal with this is to have `dependencies` support not only strings (that must be parsed) but also `PackageSpecs`.

blog/_posts/2019-08-01-artifacts.md
```diff
@@ -0,0 +1,179 @@
...
+# Pkg + BinaryBuilder -- The Next Generation
+
+Over the past few months, we have been iterating on and refining a design for `Pkg` to natively be able to reason about binary objects that are not Julia packages.  While the motivating application for this work has been improving the installation experience for binaries built with [`BinaryBuilder.jl`](https://github.com/JuliaPackaging/BinaryBuilder.jl), the artifacts subsystem is much more general.
+
+# Pkg Artifacts
+
+Pkg artifacts, as outlined in [`Pkg.jl#1234`](https://github.com/JuliaLang/Pkg.jl/issues/1234), provides a convenient way to associate containers of data with Julia projects and packages.  Artifacts are referred to by content-hash, or optionally by a name that is bound to a hash through an `Artifacts.toml` file.  An example `Artifacts.toml` file is shown here as an example:
+
+```TOML
+[socrates]
```
NHDaly: Is there a reason this one is in single-brackets? Or just a typo?

StefanKarpinski: No typo, that's how TOML syntax works: a title in single brackets introduces a single-table entry, a title in double brackets appends a table to an array of tables. There is only one `socrates` artifact stanza which applies on all platforms, there are multiple `c_simple` artifact stanzas which apply on different platforms.

blog/_posts/2019-11-14-artifacts.md
```diff
@@ -0,0 +1,232 @@
...
+        force=true
+    )
+end
+
+# Point gdk to our cached loaders
+ENV["GDK_PIXBUF_MODULE_FILE"] = joinpath(artifact_path(loaders_cache_hash), "loaders.cache")
+ENV["GDK_PIXBUF_MODULEDIR"] = gdk_pixbuf_loaders_dir
+```
+
+Within the `__init__()` method of `Gtk`, this is looking to see if a local cache of `gdk-pixbuf` loaders has been generated.  This cache is machine-specific, and must be generated upon first run of the module.  Doing so requires invoking a binary from the `gdk_pixbuf_jll` package, which is done using that JLL package's `gdk_pixbuf_query_loaders` export.  By writing the result out to a new artifact and binding that artifact to a `MutableArtifacts.toml` file (an arbitrarily-named `.gitignore`'d file), we are able to dynamically cache binary objects that are kept separately from the rest of our package data.  Gtk is thereafter informed of the cache location through environment variables.
```
KristofferC: Storing configuration next to the source file should be considered bad practice imo.

staticfloat: I totally agree with you, but since we don't have `Pkg.cache()` yet (https://github.com/JuliaLang/Pkg.jl/issues/796#issuecomment-523154714) this is the only location that a package can stash files without potentially conflicting with other versions of itself at the moment.

StefanKarpinski: Maybe mention that there is a plan to improve this in the future?

staticfloat:
Wow, my blind estimate of how much buffer time would be needed was exactly right.  This is ready to merge, from my end.

fredrikekre:
fix the date?

PR #419, opened 2019-08-05
Julia User and Developer Survey results blog post

abhijithch:
A blog post on the Julia User and Developer Survey results.


ViralBShah:
Thanks @abhijithch. Let's also upload the pdf file that was presented. Let's list all the three people from the slides as authors of this post as well.

PR #422, opened 2019-08-09
Add OpenGraph metadata

### LINKED ISSUE #421: Add og:image tags
It was pointed out in slack that julialang.org lacks og:image tags, so sites like facebook will pick one of the screenshots at random. We should add one, perhaps the same as we use for the JuliaLang repo:

https://developers.facebook.com/tools/debug/sharing/?q=julialang.org

asinghvi17:
Is this the right place?  Not sure...
cc @Keno

Fixes #421


karajan9:
That image link in this case really looks "shady af" :stuck_out_tongue:

asinghvi17:
Thanks for pointing that out :D

index.html
```diff
@@ -4,6 +4,10 @@
 julia_version: v1.1
 ---
 
+<meta property="og:title" content="The Julia Language"/>
+<meta property="og:image" content="http://www.julialang.org/images/logo.png"/>
```
Keno: Should we use the image that we're using for the repo (the top image from https://discourse.julialang.org/t/open-graph-image-for-the-julia-repository/23384/11? Would have to be added to the images/ folder in this repo). Also this should probably go into https://github.com/JuliaLang/www.julialang.org/blob/aef767c6c5070f8acea7e44aadc20013752dfe57/_includes/head.html.

asinghvi17: Agreed, would look a lot better that way - thanks for the suggestion!  Will make the changes.

asinghvi17:
@Keno done.

asinghvi17:
@ViralBShah http://www.julialang.org/images/julia-open-graph.png seems not to be parseable by Facebook?  https://developers.facebook.com/tools/debug/sharing/?q=julialang.org

ViralBShah:
Should we revert the PR?

asinghvi17:
No, it doesn't do any damage, but it doesn't fulfil its intended purpose either...

ViralBShah:
You may need to wait for 5 mins for the update to propagate to julialang.org.

asinghvi17:
Yep, it's done now!

PR #431, opened 2019-08-20
Set current release links to 1.2

ararslan:


fredrikekre:
Should 1.1 be moved to older releases as unmaintained?

ararslan:
Yeah, good idea

ararslan:
Done

fredrikekre:
Links on the frontpage and in the footer too? :)

index.html
```diff
@@ -1,7 +1,7 @@
 ---
 layout: homepage
 title:  The Julia Language
-julia_version: v1.1
+julia_version: v1.2
```
ararslan: > Links on the frontpage and in the footer too? :)

Those are taken care of by this change

ararslan:
Thanks for the review, @fredrikekre ❤️

ViralBShah:
Would be nice to have variables in github pages for the various release versions, so that we don't have to update them manually.

PR #436, opened 2019-08-25
blog post: Julia's release process

StefanKarpinski:


blog/_posts/2019-08-28-release-process.md
```diff
@@ -0,0 +1,125 @@
...
+
+# Patch releases
+
+- Patch releases increment the last digit of Julia's version number, e.g. going from `1.2.3` to `1.2.4`.
+
+- Patch releases, following [SemVer](https://semver.org/), should only contain bug fixes, low-risk performance improvements, and documentation updates. Of course, what exactly constitutes a bug fix can be more subjective than one might naïvely imagine since people write code that relies on buggy behavior. In general, we try to be very conservative with patch releases and use [PkgEval](https://github.com/JuliaComputing/NewPkgEvaljl) to ensure that there's minimal risk. People should be confident that they can just upgrade to the latest patch release without worrying about it breaking things.
+
+- Patch releases should also avoid changing internals unless it is necessary to fix a bug. Even though changing non-public code is technically fair game in any release, we want to avoid it in the name of minimizing the risk associated with patch upgrades as much as possible.
```
ChrisRackauckas: try to give an imperfect but good working definition of "public code". Documented and exported? Exported? "Well-known"? This is hard but probably necessary to set in stone in some sense.

blog/_posts/2019-08-28-release-process.md
```diff
@@ -0,0 +1,125 @@
...
+
+- Patch releases will be released approximately monthly for the currently active release branchs (more below), unless there are insufficient bug fixes on the release branch to warrant a new release, in which case a month might get skipped.
+
+- About five days before a patch release is supposed to go out, we will run PkgEval on the backports branch; if it looks good, we'll merge it and then freeze the release branch and announce on discourse that the release branch is ready for testing. If everything looks good after five days, the new patch version will be tagged.
```
ChrisRackauckas: Define PkgEval. This is something most readers likely won't know about.

ViralBShah:
Just thinking out aloud, if this should become part of the devdocs too.

StefanKarpinski:
Sure. The style is very blog post but if someone wants to edit it and make it part of the dev docs, that would be great.

blog/_posts/2019-08-28-release-process.md
```diff
@@ -0,0 +1,143 @@
...
+
+# Patch releases
+
+- Patch releases increment the last digit of Julia's version number, e.g. going from `1.2.3` to `1.2.4`.
+
+- Patch releases, following [SemVer](https://semver.org/), should only contain bug fixes, low-risk performance improvements, and documentation updates. Of course, what exactly constitutes a bug fix can be more subjective than one might naïvely imagine since people write code that relies on buggy behavior. In general, we try to be very conservative with patch releases and use [PkgEval](https://github.com/JuliaComputing/NewPkgEvaljl) to ensure that there's minimal risk. People should be confident that they can just upgrade to the latest patch release without worrying about it breaking things.
```
giordano: The address of PkgEval is missing a dot:
```suggestion
- Patch releases, following [SemVer](https://semver.org/), should only contain bug fixes, low-risk performance improvements, and documentation updates. Of course, what exactly constitutes a bug fix can be more subjective than one might naïvely imagine since people write code that relies on buggy behavior. In general, we try to be very conservative with patch releases and use [PkgEval](https://github.com/JuliaComputing/NewPkgEval.jl) to ensure that there's minimal risk. People should be confident that they can just upgrade to the latest patch release without worrying about it breaking things.
```

blog/_posts/2019-08-28-release-process.md
```diff
@@ -0,0 +1,143 @@
...
+
+- What a major release does allow, however, is fixing obvious API design mistakes—the kinds of bad, confusing APIs that everyone will be glad to be rid of. It also allows changing low-level things that will break some libraries, but which need to be broken in order to make really fundamental improvements to the language. Changes to Julia's AST representation and how macros work fall into this latter category.
+
+- My favorite example of a breaking change that I want to make in Julia 2.0 is to turn multithreading on by default instead of requiring the user to opt in. Technically this is breaking: if someone wrote thread-unsafe code in 1.x, assuming single-threaded execution by default, if we suddenly turn threading on in Julia 1.x, their code will stop working, breaking our SemVer compatibility promise for non-breaking minor releases. So we need to wait until 2.0 to flip that switch.
```
KristofferC: I don't understand why we need to wait until 2.0 to enable multithreading by default. The code will only stop working when it is run under in a multithreading context, but on its own, it works perfectly well. And it is up to the person running that code multithreaded to make sure it is threadsafe.

KristofferC: Or wait, did you mean that "enable multithreading by default" means spawning tasks to be non-sticky by default? If that's the case, then yes that would be breaking. I thought you means just running with `nthreads() > 1` by default.

StefanKarpinski: I mean setting `JULIA_NUM_THREADS` equal to the number of cores by default.

StefanKarpinski: I think I explained pretty clearly why it's a breaking change: code that worked, assuming that there will be only one thread will break if there are suddenly multiple threads.

ViralBShah:
Would it be useful to discuss triage calls and process here?

KristofferC:
Yeah, I think that is interesting enough to warrant its own post.

fredrikekre:
Title doesn't seem to render.

PR #447, opened 2019-09-07
blog: quick post about Profiler.jl update work

vtjnash:
This blog post comes with a PR! https://github.com/JuliaLang/julia/pull/33190


blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+-#if defined(_OS_WINDOWS_) || defined(_OS_FREEBSD_)
+ #  define JL_DISABLE_FPO
+-#endif
+ 
+ #include <iostream>
+ #include <sstream>
+```
+
+But otherwise, this seems fairly accurate, but yikes, what an arrow! I needed over 40 frames this way, and that makes this pretty hard to interpret.
+Spiky frame-graphs like this can be quite misleading to look at since the visual size (height) isn't relevant to performance.
```
timholy: I know you know this, but the *width* is the part that's relevant to performance---height is just recursion depth. Once you internalize that, they aren't misleading at all, as they show two "orthogonal" axes.

If there's still an underlying point you're trying to make, it might need rephrasing.

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+    14 int.jl      53 +
+   212 boot.jl    330 eval
+  5717 REPL[2]      1 fib_r
+  6028 REPL[1]      2 fib
+
+julia> count(==(0), Profile.fetch())
+585
+```
+
+So that's pretty bad—we have about 212 samples, and about 2000% of them were in `fib`. Clearly not true. What happened here? We computed that `fib` (and `fib_r`) each appeared an average of 10-20 times in every backtrace, so we over-represent them in our flat-format output. OK, so first thing to do is to fix that: each frame should only be counted once towards each output line.
```
timholy: Might want to explain where the 212 comes from---why is `eval` a reliable indicator of the actual number of samples?

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+  ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎  2   int.jl:53; +
+  ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎  2   int.jl:52; -
+  ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎     1   int.jl:53; +
+  ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎   1   int.jl:52; -
+  ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎ 1   int.jl:53; +
+  ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    1   int.jl:53; +
+  ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎    ╎  2   int.jl:53; +
+```
+
+We also see this has issues with identifying the stack in many cases (almost 40% failed!), similar to `perf`. One way to improve that (~10% failure rate) is to disable the omit-frame-pointer option, like we did before. Currently this requires this source code change:
```
timholy: A bit unclear. Are you saying that if you disable the option, you're left with a 10% failure rate?

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+Spiky frame-graphs like this can be quite misleading to look at since the visual size (height) isn't relevant to performance.
+And it make it really hard to see where the time is being spent in aggregate.
+
+Julia implements a `noisefloor` option as well as `mincount` and `maxdepth` to help with some situations, but they do nothing for this example.
+The `mincount` option (also present in `Instruments.app` for min and max # samples; and in `perf` as a min percent threshold) drops any line that doesn't meet that threshold. The `maxdepth` option (aka `print_limit` in `perf`) cuts off the tree after some recursion limit.
+The `noisefloor` option is similar to `mincount`, but operates as a relative fraction of the parent frame instead of a constant threshold (I'm not aware of a similar option in other tools). This can help hide frames from the tree that contribute little to the immediate performance. This is based on the value engineering principle of focusing on the largest contributors first. For many problems, these can be greatly beneficial.
```
timholy: 2nd mention of `noisefloor`, would be better to consolidate. This explanation also suggests that it's a constant fraction, but when I last worked on this it tested whether the child was more than `noisefloor*sqrt(parent)` samples, as that's the fluctuation expected from sampling noise.

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+
+However, none of these option are helpful for our tree above. We need more tools in our toolbox!
+
+Let's now examine how we can try to consolidate the recursive frames somehow. One option is the flat reports we looked at earlier.
+
+Here's that report with the counts corrected, and some extra information also now compiled for it:
```
timholy: A bit unclear; is this a consequence of your PR? Some readers might look for a different set of options you passed or something.

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+ Count  Overhead File      Line Function
+   228         2 ./boot.jl  330 eval
+     2         2 ./int.jl    52 -
+    25        25 ./int.jl    53 +
+   321       155 ./REPL[2]    1 fib_r
+   514       426 ./REPL[1]    2 fib
+Total snapshots: 610
+```
+
+That's better! But now we've had to give up on the tree view, losing the ability to see the context of each call. Some of those functions are very common, like `+`, and it might not be sufficient to see them in aggregate like this. Can we combine the benefits of both views somehow?
```
timholy: Since we don't know what changes you made to the code, it's unclear why `format=:tree` is now broken. (Hopefully it isn't?!?)

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+
+First, the Instruments.app tool we've looked at above has an option to collapse recursive calls, which will take a callstack with one repeated entry and replace it with one entry that takes the sum of that recursion. This has no effect on our example code, since there's two entries being repeated. While this mode can simplify the stack view and make it easier to explore, it doesn't provide us with additional insight into the code.
+
+There's another visualization-only tool called Speedscope which has [an option to do this](https://github.com/jlfwong/speedscope/pull/68). It imports profile files created by many other tools and languages, so we'll import the earlier `perf` file and see how it does against our lineup:
+
+![Speedscope bad result](/images/blog/2019-09-06-profilers/profile-speedscope.png)
+
+Looks about right! Only one silly nit: it claims the sum of all frames identical to the highlighted frame (just itself) took 367.38 ms (good), but that the highlighted frame alone took 1.05 s (oops, wrong ordering for the summarizing and the counting). But otherwise, this seems to show about the right information we've come to expect.
```
timholy: The part after `oops` is unclear.

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+        33  5.22% 99.37%         33  5.22%                | + ./int.jl:53
+----------------------------------------------------------+-------------
+                                                 1   100% |   - ./int.jl:2 (inline)
+         1  0.16% 99.53%          1  0.16%                | - ./int.jl:52
+----------------------------------------------------------+-------------
+```
+
+Let's combine all those ideas to make a new option for our Julia plain-text tree view: flattened recursion.
+
+  * recur – Controls the recursion handling in `:tree` format. `off` (default) prints the tree as normal. `flat` instead compresses any recursion (by ip), showing the approximate effect of converting any self-recursion into an iterator. `flatc` does the same but also includes collapsing of C frames (may do odd things around `jl_apply`).
```
timholy: Bulleting & quoting seem weird here (it's a 1-item list, unclear what's a keyword argument and what's a value that argument can have).

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+julia> Profile.print(format=:tree, recur=:flat)
+106 ./REPL[1]:1; fib_r(::Int64)
+250 ./REPL[2]:2; fib(::Int64)
+275 ./task.jl:333; (::REPL.var"##26#27"{REPL.REPLBackend})()
+ 275 /Users/jameson/julia/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:118; macro expansion
+  275 /Users/jameson/julia/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:86; eval_user_input(::Any, ::REPL.REPLBackend)
+   275 ./boot.jl:330; eval(::Module, ::Any)
+  ╎ 273 ./REPL[2]:2; fib(::Int64)
+  ╎  309 ./REPL[1]:1; fib_r(::Int64)
+  ╎   469 ./REPL[2]:2; fib(::Int64)
```
timholy: 469 > 309 seems a bit counterintuitive.

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+                                                 1   100% |   - ./int.jl:2 (inline)
+         1  0.16% 99.53%          1  0.16%                | - ./int.jl:52
+----------------------------------------------------------+-------------
+```
+
+Let's combine all those ideas to make a new option for our Julia plain-text tree view: flattened recursion.
+
+  * recur – Controls the recursion handling in `:tree` format. `off` (default) prints the tree as normal. `flat` instead compresses any recursion (by ip), showing the approximate effect of converting any self-recursion into an iterator. `flatc` does the same but also includes collapsing of C frames (may do odd things around `jl_apply`).
+
+This is no longer a literal interpretation of the stack, but shows a particular linearization where each recursive call was re-written as a loop.
```
timholy: There's not quite enough information given to understand the rules here. I think what you mean is that if an IP appears twice in the trace, you truncate the trace at the second one. But why the second one? Why not the first one?

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+  ╎  309 ./REPL[1]:1; fib_r(::Int64)
+  ╎   469 ./REPL[2]:2; fib(::Int64)
+  ╎    36 ./REPL[1]:1; fib_r(::Int64)
+  ╎    2  ./REPL[2]:2; fib(::Int64)
+  ╎    33 ./int.jl:53; +
+  ╎    1  ./int.jl:52; -
+  ╎  3   ./REPL[2]:2; fib(::Int64)
+```
+
+I think this looks cool, and vaguely even shows how I would conceptualize the functions in my head. This is availble right now only on [my jn/profile-recurflat](https://github.com/JuliaLang/julia/tree/jn/profile-recurflat) branch, but should be a PR soon, and show up in the JuliaLang v1.4 release.
```
timholy: You might need to expand upon what's good about this representation: why does `fib` sometimes show up as a child of itself? Walk people through how many times it does the direct call vs the recursive call.

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+
+And finally, we're working on a [memory profiler](https://github.com/JuliaLang/julia/pull/31534) for JuliaLang to let you explore questions of memory consumption, similar to the time-duration-consumed questions explored above. So watch for more updates to come!
+
+
+---
+
+Post-note: There was also an off-by-one lookup error in Profile.jl too that's will soon be resolved. This is almost never an issue, but we can sometimes notice it as bogus information showing some frame that should never have been reached. For example, we can profile a call to the expensive `cpuid` instruction, and see that there's a bogus frame that makes it appear we used some code that happened to be located outside the target function was somehow getting used.
```
timholy: "that's" -> "that"

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
+---
+layout: post
+title: Profiling tool wins and woes.
+author: <a href="http://github.com/vtjnash/">Jameson Nash</a>
+---
+
+Profiling tools are awesome. They let us see what actually is affecting our program performance. Profiling tools also are terrible. They lie to us and give us confusing information. They also have some surprisingly new developments: [brendangregg's oft-cloned flamegraphs tool](http://www.brendangregg.com/flamegraphs.html) was created in 2011! So here I will be investigating some ways to make our profile reports better; and looking at ways in which they commonly break, to raise awareness of those artifacts in the reports.
```
vchuravy: `oft-cloned` often cloned?
```suggestion
Profiling tools are awesome. They let us see what actually is affecting our program performance. Profiling tools also are terrible. They lie to us and give us confusing information. They also have some surprisingly new developments: [brendangregg's often cloned flamegraphs tool](http://www.brendangregg.com/flamegraphs.html) was created in 2011! So here I will be investigating some ways to make our profile reports better; and looking at ways in which they commonly break, to raise awareness of those artifacts in the reports.
```

blog/_posts/2019-09-06-profilers.md
```diff
@@ -0,0 +1,405 @@
...
+}
+int main(void) {
+    return fib(40);
+}
+```
+
+On my macOS machine, this takes on the order of one second, so that's usually a good number of samples.
+
+```
+$ clang -Wall -Werror -pedantic -std=c11 -fno-optimize-sibling-calls -fno-inline -O3 -g fib.c -o fib
```
vchuravy: Reference why `-fno-optimize-sibling-calls` is important

blog/_posts/2019-09-06-profilers.md
vtjnash: yes it does. I'll review my code and see if something isn't being updated correctly. (It wouldn't be the first bug that writing out carefully the inputs and outputs here has caught! I've just already squashed the others into the history, haha.)

blog/_posts/2019-09-06-profilers.md
vtjnash: I actually have no idea why that happens either. It show up in most of the other reports too—this is just the first one where it sticks out like a sore thumb. I'll add a note to call that out.

blog/_posts/2019-09-06-profilers.md
vtjnash: Ah, haha, this is the same mistake I pointed out in Speedscope above (https://github.com/JuliaLang/www.julialang.org/pull/447#discussion_r322006957). I need to rework some things.

blog/_posts/2019-09-06-profilers.md
vtjnash: Neither. It's not truncated, it's combined.

Fun fact: the new `format=:flat` algorithm forms a subset of this new `tree-flat` option, but where all calls (instead of just recursive ones) are reinterpreted as if called from a single interpreter loop. To put that in the context of Julia, if you had a tree of calls that all went through `jl_apply`, the `tree-flatc` and `format=flat` output should start to coincide (plus-or-minus some formatting and indentation).

PR #464, opened 2019-10-06
Update packages.html

logankilpatrick:
Updated figures per the Julia computing numbers here: https://juliacomputing.com

### DIFF
--- a/_includes/packages.html
+++ b/_includes/packages.html
@@ -26,7 +26,7 @@
       </div>
 -->
       <p>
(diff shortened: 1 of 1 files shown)


ViralBShah:
@KristofferC Is the 3000 packages number ok - given that there was a plan to remove a number of packages that were not 1.0 compatible. Can you recommend what the right thing to do is? Report the current number - or wait until the deletion is carried out?

PR #475, opened 2019-10-21
README: remove rvm, install jekyll locally

vtjnash:
I decided I don't like RVM, but I found out we can somewhat improve the `bundle install` step.


ViralBShah:
merge?

vtjnash:
Sure. I just make it a PR so people get a notification.

PR #503, opened 2019-12-11
3000 packages

PallHaraldsson:
From Google headquarters talks: https://youtu.be/QFw5yzZuZQw?t=1137

[Yes, title of that slide says 10 million but newsletter has 11 million, may also be outdated.]

juliacomputing.com still says 2500 registers packages, I guess outdated.

### DIFF
--- a/_includes/packages.html
+++ b/_includes/packages.html
@@ -26,7 +26,7 @@
       </div>
 -->
       <p>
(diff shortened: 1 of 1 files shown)


ViralBShah:
The right number is 2,500.

PallHaraldsson:
Thanks for merging with more correct number. I however see now (closer to 3000...):

"Listing all 2753 registered packages for the Julia programming language.
Made with NewPkgEval.jl, last updated on 2019-12-03. "

Issue #516, opened 2019-12-24
Add installation instructions for linux


abhishalya:
This should add instructions on how to add binaries to the path on linux so Julia can be run from the command line.

PR #518, opened 2019-12-24
Add more instructions for linux installation

abhishalya:
This adds few more instructions for linux installation.
Also it adds a method on how to add Julia's bin folder
to `PATH` variable.

Closes https://github.com/JuliaLang/www.julialang.org/issues/516

### DIFF
--- a/downloads/platform.html
+++ b/downloads/platform.html
@@ -72,27 +72,36 @@ title:  Platform specific instructions
 
       <p>
         The generic Linux and FreeBSD binaries do not require any special installation steps, but you will need to ensure that your system can find
-        the <code>julia</code> executable. First, extract the <code>.tar.gz</code> file downloaded from the <a href="index.html">downloads page</a> to a folder on your computer.
+        the <code>julia</code> executable.
+      </p>
+
+      <p>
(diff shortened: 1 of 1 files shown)


abhishalya:
It would be really nice to have netlify deploy previews here.

abhishalya:
Have a look at https://julialang-preview.netlify.com/downloads/platform.html for the preview.

ViralBShah:
What do we need to do to set up the previews?

For linux, let's remove the solution with the linking and just have the PATH based solution. It should basically be: `PATH=$PATH:<julia-directory>/bin`.

ViralBShah:
Also, let's suggest add the path in `~/.bash_profile` of the equivalent and not in /etc/environments etc. In general we do not require admin/root access for Julia installation and usage. So ideally, all the sudo stuff should be removed.

abhishalya:
>What do we need to do to set up the previews?

We just need to add this repository to netlify.com, it will ask for build command. That's it. It should install all deps automatically.

abhishalya:
@logankilpatrick @ViralBShah Can you have a look at this again? :)

ViralBShah:
I think this is good. I'll tweak it slightly later, but merging for now. Thanks!

PR #521, opened 2019-12-24
Add instructions to Windows Installation

### LINKED ISSUE #514: Add Instructions for Adding Binaries to PATH Windows 10
Windows 10 lacks instructions for installing binaries and adding them to PATH. I am currently working on this.

soumitradev:
Adds instructions for Windows Users (Windows 7 and Above) for adding Julia to PATH.

The preview is live on here: https://julia-preview.netlify.com/downloads/platform.html

Resolves #514

### DIFF
--- a/downloads/platform.html
+++ b/downloads/platform.html
@@ -31,8 +31,8 @@ title:  Platform specific instructions
         <p>
           If on Windows 10,
           <ol>
-            <li>Hit the Windows Key, and type in "env", and select "Edit the system environment variables"
-            <li>Click the "Environment Variables..." button.
+            <li>Hit (Windows Key + R) to open run, and type in <code>sysdm.cpl</code> and hit enter. This will open up the System Properties. Click on the "Advanced" tab.
+            <li>Now, click the "Environment Variables..." button.
             <li>Under the "System Variables" section (the lower half), find the row with "Path" in the first column, and click edit.
             <li>The "Edit environment variable" UI will appear. Here, click "New", and paste in the text you copied into the Notepad.
             <li>Hit OK and dismiss all the dialogs. You can now run Julia from the command line!
--- a/downloads/platform.html
+++ b/downloads/platform.html
@@ -41,7 +41,7 @@ title:  Platform specific instructions
         <p>
           Else, if on Windows 7 or 8,
           <ol>
-            <li>Go to the Windows Desktop and right click on "My Computer" and hit "Properties"
+            <li>Hit (Windows Key + R) to open run, and type in <code>sysdm.cpl</code> and hit enter. This will open up the System Properties.
             <li>In the System Properties window, click the Advanced tab, and then click "Environment Variables".
             <li>In the System Variables window, highlight Path, and click Edit.
             <li>In the Edit System Variables window, move the cursor to the end of the field.
--- a/downloads/platform.html
+++ b/downloads/platform.html
@@ -24,10 +24,33 @@ title:  Platform specific instructions
         Julia is available for Windows 7 and later, both 32 bit and 64 bit.
         <ol>
           <li>Download the Windows julia.exe installer for your platform. 32-bit julia works on both x86 and x86_64. 64-bit julia will only run on 64-bit Windows (x86_64).
-          <li>Run the downloaded program to extract julia
-          <li>Double-click the julia shortcut in the unpacked folder to start julia
+          <li>Run the downloaded program to extract Julia.
+          <li>At the "Choose Installation directory" step, copy the Path in the "Destination Folder" into a Notepad or similar file for future reference, and finish installation.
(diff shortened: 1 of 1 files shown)


ViralBShah:
I don't know enough about windows, but will boot up my VM to try it out.

downloads/platform.html
```diff
@@ -24,11 +24,34 @@ <h2 id='Windows'>Windows</h2>
...
+            <li>Click the "Environment Variables..." button.
+            <li>Under the "System Variables" section (the lower half), find the row with "Path" in the first column, and click edit.
+            <li>The "Edit environment variable" UI will appear. Here, click "New", and paste in the text you copied into the Notepad.
+            <li>Hit OK and dismiss all the dialogs. You can now run Julia from the command line!
+          </ol>
+        </p>
+        <p>
+          Else, if on Windows 7 or 8,
+          <ol>
+            <li>Go to the Windows Desktop and right click on "My Computer" and hit "Properties"
```
abhishalya: `My Computer` icon might not be available for some users since they should be enabled first. So, a more general solution here too would be useful.

downloads/platform.html
```diff
@@ -24,11 +24,34 @@ <h2 id='Windows'>Windows</h2>
...
-          <li>Run the downloaded program to extract julia
-          <li>Double-click the julia shortcut in the unpacked folder to start julia
+          <li>Run the downloaded program to extract Julia.
+          <li>At the "Choose Installation directory" step, copy the Path in the "Destination Folder" into a Notepad or similar file for future reference, and finish installation.
+          <li>At this point, you can decide whether you want to invoke Julia by simply typing <code>julia</code> in the command line, or paste the entire path you saved in the Notepad. To invoke Julia by simply typing <code>julia</code> in the command line, perform the following steps.
         </ol>
+        <p>
+          If on Windows 10,
+          <ol>
+            <li>Hit the Windows Key, and type in "env", and select "Edit the system environment variables"
```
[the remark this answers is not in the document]
soumitradev: True, I’ll change that next morning. Thanks for pointing out! I did not realise that a web result may pop up.

downloads/platform.html
soumitradev: True, but I don’t have a Windows 7 OS around me. Maybe a Windows 7 user can change this? Anyways, this was the only method I found for Windows 7, so I added it.

soumitradev:
@abhishalya Just made suggested changes to Windows 10 Instructions, but I'm not sure about Windows 7 & 8. Maybe someone with Windows 7 or 8 could help?

fluctix:
@soumitradev i can confirm that windows 7 instructions are same as that on windows 10

soumitradev:
@fluctix Thanks, just added the necessary changes. This PR should be good to go now.

ViralBShah:
Thanks all.

downloads/platform.html
```diff
@@ -24,11 +24,34 @@ <h2 id='Windows'>Windows</h2>
...
+          <ol>
+            <li>Hit (Windows Key + R) to open run, and type in <code>sysdm.cpl</code> and hit enter. This will open up the System Properties. Click on the "Advanced" tab.
+            <li>Now, click the "Environment Variables..." button.
+            <li>Under the "System Variables" section (the lower half), find the row with "Path" in the first column, and click edit.
+            <li>The "Edit environment variable" UI will appear. Here, click "New", and paste in the text you copied into the Notepad.
+            <li>Hit OK and dismiss all the dialogs. You can now run Julia from the command line!
+          </ol>
+        </p>
+        <p>
+          Else, if on Windows 7 or 8,
```
KristofferC: Can we remove this? Windows 7 is EOL and no one uses Windows 8.

downloads/platform.html
```diff
@@ -24,11 +24,34 @@ <h2 id='Windows'>Windows</h2>
...
+          <li>At this point, you can decide whether you want to invoke Julia by simply typing <code>julia</code> in the command line, or paste the entire path you saved in the Notepad. To invoke Julia by simply typing <code>julia</code> in the command line, perform the following steps.
         </ol>
+        <p>
+          If on Windows 10,
+          <ol>
+            <li>Hit (Windows Key + R) to open run, and type in <code>sysdm.cpl</code> and hit enter. This will open up the System Properties. Click on the "Advanced" tab.
+            <li>Now, click the "Environment Variables..." button.
+            <li>Under the "System Variables" section (the lower half), find the row with "Path" in the first column, and click edit.
+            <li>The "Edit environment variable" UI will appear. Here, click "New", and paste in the text you copied into the Notepad.
+            <li>Hit OK and dismiss all the dialogs. You can now run Julia from the command line!
```
KristofferC: This doesn't seem to work?

Screenshot from 2020-01-27 12-23-17

KristofferC: Also, shouldn't it be a User environment variable and not a System variable?

ViralBShah:
The last time this came up, it was discussed that since Julia does work on Windows 7, we should support. 

I think the decision to not support Windows 7 needs to be taken across the board.

KristofferC:
Made an issue instead https://github.com/JuliaLang/www.julialang.org/issues/615

PR #528, opened 2019-12-26
Redirecting julialang.org/download to julialang.org/downloads

fungktt:
As seen from the title.

Tested here: https://julialang-test.netlify.com/download/ which redirects to julialang.org/downloads

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,1 @@
+Redirect 302 /download/ https://julialang.org/downloads/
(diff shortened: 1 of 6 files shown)


logankilpatrick:
Note that this does not appear to be working. JuliaLang.org/download still shows an error 404 for me

ViralBShah:
Same here.

ViralBShah:
Maybe Github pages doesn't like this plugin?

ViralBShah:
I hear that we should be moving to Hugo from Jekyll.

ViralBShah:
We probably should revert this.

PR #530, opened 2019-12-27
Fix dead links

abhishalya:
This fixes most of the dead links, if not all.

Most of the links are updated to their current version while few are removed since they were no where to be found. I think I've fixed most of them, do have a look at the [actions log](https://github.com/abhishalya/www.julialang.org/actions) to know more.

Also, I've used [remark-lint-no-dead-urls](https://github.com/davidtheclark/remark-lint-no-dead-urls) because I think it works better than the current one. One of the primary reason to use it was that it ignores the relative-urls. Do have a look, I think it would be much better to get both of the commits merged.

PS. This took a *lot* of time.


abhishalya:
Few more things:

- `localhost` should be ignored, there is an option available with remark but we'll need to add a config file for that, I wasn't sure of this, so I just ignored [this](https://github.com/JuliaLang/www.julialang.org/pull/530/checks?check_run_id=364919845#step:5:5) result.
- It seems to me that app.juliadiffeq.org takes time to load and hence it times out. The url itself should be correct.
- This link -> https://zh.wikipedia.org/zh-hk/兒童黑話 is actually working, seems to me that due to language the tools aren't able to parse the response properly.
- LinkedIn links need users to login, due to which they are also thrown by the tools.
- I'm not sure the use of `target` in `manual/standard-library-reference/index.md` file.

abhishalya:
In total I think this PR fixes about ~160 links.

ViralBShah:
This is really great. Thank you!

PR #535, opened 2019-12-28
Redirecting julialang.org/download to julialang.org/downloads (Improving PR #528)

fungktt:
Improving PR #528

This requires the line `gem 'jekyll-redirect-from'` in the Gemfile so if #528 was reverted, please add this here.

This method works both locally and on netlify.

test: https://julialang-test.netlify.com/downloads/


ViralBShah:
Thanks - this does  work!

PR #536, opened 2019-12-28
Remove old projects from gsoc

haampie:
These descriptions have not been changed for years, so it's better to remove them for now. They should later be replaced with a proper concrete project in numerical linalg


ChrisRackauckas:
I thought this needed to be done, thanks

PR #537, opened 2019-12-28
Updating the outdated README.md

fungktt:
As seen [here](https://help.github.com/en/github/working-with-github-pages/testing-your-github-pages-site-locally-with-jekyll), the command `./bin/jekyll serve` is now `bundle exec jekyll serve` instead.

I found out as `./bin/jekyll serve` didn't work for `ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-darwin19]`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ then
 from the root of this repository to fetch all dependencies and install them locally.
 You only have to do this once.
 
-Now you need to check your ruber version by running `ruby -v` in the terminal. 
+Now you need to check your ruby version by running `ruby -v` in the terminal. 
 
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Let me check my ruby version and we can add this with a  note for the different versions. I’ll update and merge tonight.

fungktt:
@logankilpatrick Please type `ruby -v` in your Terminal. Thanks.

logankilpatrick:
I was on `ruby 2.6.3p62` so I updated it to show the different ways for both to avoid any confusion.

PR #545, opened 2020-01-06
rm old soc project

Roger-luo:
These have been done in G/JSoC 2019 or externally. Sorry to update it so late. I'll add some new ideas later.


logankilpatrick:
Thanks!

PR #547, opened 2020-01-09
Remove myself from GSoC

simonbyrne:
Unfortunately I won't have time to supervise GSoC this year.

I also removed the openlibm project, as we're fazing that out.


logankilpatrick:
Thanks!

PR #563, opened 2020-01-12
Delete plotting.html

### LINKED ISSUE #550: Re-think where plotting.html will live
Plott.html currently sits in the download folder because it's a core aspect of what a Julia user may want to do when playing around with and using Julia shortly after downloading. 

Some thought needs to go into if this is the best approach.

logankilpatrick:
Resolved #550 

I do not think that this page is used anywhere on the website. I just looked and could not find it. We should either add it somewhere or remove this file.


logankilpatrick:
@ViralBShah thoughts?

PR #564, opened 2020-01-12
Update 404.html

logankilpatrick:
Fixed links on 404 page from markdown format that wasn't rendering to HTML link formatting.


logankilpatrick:
Preview uses Netlify so we cant see 404's correctly.

PR #565, opened 2020-01-12
Fix broken URLs for blog articles

benelsen:
Added the old slugs for the blog articles from the the filenames.
This should fix the 404s.


logankilpatrick:
Looks good in the preview.

PR #566, opened 2020-01-12
Update head.html

logankilpatrick:


ViralBShah:
That might be a simple enough fix. Looks like it ought to work.

logankilpatrick:
did not appear to do the job

PR #569, opened 2020-01-12
head.html: Fix absolute css links

abhishalya:


logankilpatrick:
This is tricky. Not sure why it isn't working

abhishalya:
@logankilpatrick Looks good now

Issue #570, opened 2020-01-12
Consider switching links at the top of the website to Absolute URLs


logankilpatrick:
if you go to julialang.org/jsoc/test and then get the 404 error but choose to click any of the buttons, they do not work to re-direct you anymore. They all stop working and just add "/jsoc" or whatever other button you press to the url.

ViralBShah:
Same problem as #555

PR #573, opened 2020-01-12
Fix relative links

abhishalya:
Closes https://github.com/JuliaLang/www.julialang.org/issues/570


abhishalya:
@ViralBShah Can you check, this looks good now

PR #575, opened 2020-01-12
Update head.html

logankilpatrick:


logankilpatrick:
I think this finally fixes the 404 issue.

PR #585, opened 2020-01-15
Adapt fluxdifffeq blog post to Hugo

goggle:
Adapts the blog post https://julialang.org/blog/2019/01/fluxdiffeq/ to the new Hugo syntax.

### DIFF
--- a/content/blog/2019-01-18-fluxdiffeq.md
+++ b/content/blog/2019-01-18-fluxdiffeq.md
@@ -6,7 +6,7 @@ title: DiffEqFlux.jl – A Julia Library for Neural Differential Equations
 slug: fluxdiffeq
 ---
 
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Thanks!

logankilpatrick:
Probably should have merge #582

goggle:
Hmm, I think we need both, don't we?
If I understand this correctly, #582 updates the programming syntax, my PR updates the syntax to render the blog post.

logankilpatrick:
This PR just broke the post so I reverted.

logankilpatrick:
I will try again after I merge Chris's PR.

logankilpatrick:
Okay, can you try again @goggle

ChrisRackauckas:
I updated just the code syntax and not much else

ChrisRackauckas:
I would be very surprised if that was the issue

PR #587, opened 2020-01-15
Adapt fluxdifffeq blog post to Hugo

goggle:
Here is the second try to adapt the syntax for the fluxdiffeq blog post.
It looks fine in my browser when I start the hugo server.

What did not work on the first try? Just merge conflicts with Chris' PR?


goggle:
@ChrisRackauckas If you want, you can skim over https://julialang.org/blog/2019/01/fluxdiffeq/ and see if everything looks like it used to look on the old website.

logankilpatrick:
Great work! Looked good when I skimmed through.

cormullion:
There’s also a Chinese translation of this post, which would probably require the same edits.

PR #591, opened 2020-01-16
Fix ML&PL broken blog articles permalinks, fix inconsistant pi encoding

benelsen:
I found 3 dead permalinks which were caused by `&` in the slugs. Hugo seemingly removes those from the URLs.
Given the time passed since the switch and that it might not be ideal to have `&` in URLs I thought it would be better to just add a redirect for the old permalinks and keep the current URLs.

I also fixed two πs that were encoded as character entity references which rendered inconsistently on the the page of the post and the list of blog posts.


logankilpatrick:
Thanks!

PR #593, opened 2020-01-17
Adapt Chinese translation of fluxdifffeq blog post to Hugo

goggle:
This is a follow-up PR to #587.

Please note that I've changed `\gamma` to `\delta` on line 307, because I think that was a typo.


logankilpatrick:
Thanks!

PR #595, opened 2020-01-17
Re-enable link to ARMv7 binaries.

fredrikekre:
5d35489638d6260b6a3d5a284e36368451811991 seems to have been lost in the transition to Hugo, cc @ararslan


logankilpatrick:
Thanks!

PR #601, opened 2020-01-21
Julia at University of Veracruz

jmejia8:
Taller de Algoritmos Evolutivos y de Inteligencia Colectiva usando Julia.
Workshop on Evolutionary and Swarm Intelligence Algorithms using Julia.


logankilpatrick:
Thanks!

PR #602, opened 2020-01-23
Fix alignment of buttons in Homepage banner

### LINKED ISSUE #592: Buttons on banner do not align nicely
The banner on the homepage currently looks like this:
image

I'm not a design expert, but the alignment of the three buttons looks wrong to me.
I would suggest to remove the github button from the banner (and maybe put it somewhere else on the homepage) and align the two colored buttons in the center below "The Julia Programming Language".

goggle:
This PR intends to fix #592.

Here are some screenshots of the banner after this change:
Desktop:
image
Tablet:
image
Mobile:
image

### DIFF
--- a/layouts/partials/jumbotron.html
+++ b/layouts/partials/jumbotron.html
@@ -16,7 +16,7 @@
 -->
 </br/>
       <div class="row">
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Thanks!

PR #612, opened 2020-01-26
Remove trace estimation project from jsoc ideas

mohamed82008:
This project was nearly completed last year in https://github.com/luca-aki/TraceEstimation.jl so there is not enough work for a new student.


ViralBShah:
Any new projects we can add?

mohamed82008:
Not for linalg for me. I will add some under Turing.

PR #613, opened 2020-01-26
Fix missing entries on /research

benelsen:
@juliohm noticed that some entries were missing from the research page.
For some reason the year in the yaml header needs to be quoted.
This PR fixes those entries for which this wasn't the case and now show up in the list.
I also added a note in the README.md of this requirement (ideally this shouldn't be one imo, but I don't know where it emanates from)


ViralBShah:
@abhishalya Any ideas how to avoid this?

abhishalya:
Yeah, will fix this in the next PR.

Issue #614, opened 2020-01-27
Publication list formatting


thofma:
While looking at the publication list, I noticed that the package names are all lowercase. Also the list of packages always ends with an ", ", which is then followed by ".".

Also the author list of https://github.com/JuliaLang/www.julialang.org/blob/88404b1c83b4a121370b1be30c0747808c6d6710/content/research/BST19.md (the first item on the list) is not formatted properly.

I looked for a quick fix, but I am lacking the necessary Hugo skills.

ViralBShah:
@abhishalya

abhishalya:
Thanks for reporting, will fix this soon.

abhishalya:
I've opened a PR to fix this. It does take care of the `, ` delimit thing. But I see that by default Hugo lowercases the frontmatter variables. There's some discussion [here](https://discourse.gohugo.io/t/would-be-great-if-frontmatter-variables-didnt-have-to-be-all-lowercased/1608/3) on that.

PR #617, opened 2020-02-02
Fix delimit for research packages

abhishalya:
Closes https://github.com/JuliaLang/www.julialang.org/issues/614

### DIFF
--- a/layouts/partials/citation.html
+++ b/layouts/partials/citation.html
@@ -3,11 +3,19 @@
 </head>
 
 <b>{{ .Title }}</b>.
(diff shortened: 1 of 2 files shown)


ViralBShah:
Please merge if this is good to go.

thofma:
Does this also fix the case sensitivity issue? If not, then this should not close #614.

ViralBShah:
Or open a separate issue for case sensitivity? Apparently, this is something about hugo itself.

thofma:
I will open a new issue then.

PR #628, opened 2020-02-08
Add issue button in page footer

logankilpatrick:


logankilpatrick:
This does not really look that good

ViralBShah:
Maybe just add it as text rather than as a button?

PR #657, opened 2020-03-13
Add Google Analytics back

logankilpatrick:


tlienart:
Oops! good catch, looks good to me

PR #674, opened 2020-03-18
Add UCX project

shashi:


ViralBShah:
@keno always told me that the thing to use for this was [LibFabric](https://ofiwg.github.io/libfabric/). Does someone know the difference between UCX and fabric? Are they related in any way?

Keno:
They are basically competing standards with similar goals. I do agree that ucx seems to have gotten a bit more traction recently, but that's probably just because it's what's used on summit. E.g. I don't think ucx works on Cray machines. Anyway, I think working on either is fine. They should be basically interchangable for our purposes.

vchuravy:
I spend some time looking at both and I like UCX better and there is also https://github.com/rapidsai/ucx-py which is supposedly used by Dask.

PR #675, opened 2020-03-21
Update download links to v1.4.0

ararslan:
Also move v1.3.1 to the old releases page and update the homepage download button to say 1.4.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -137,69 +137,6 @@ Checksums for this release are available in both, [MD5](https://julialang-s3.jul
 @@ @@
 
 
-
-## Upcoming release: v1.4.0-rc2 (Feb 25, 2020)
-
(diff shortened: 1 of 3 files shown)


ViralBShah:
@logankilpatrick can we delete all these failing CI things?

logankilpatrick:
It should be all set now

ViralBShah:
@tlienart Is there a way in Franklin to set some values - like lts_release, stable_release, etc. so that we don't have to edit these things manually in each release and they are uniformly updated across the website?

tlienart:
Yes, basically just define variables in the `config.md`, these can then be used in HTML using `{{fill varname}}`

If someone gives me more detailed specs of what/where this is desired, I’m happy to put this in place

ViralBShah:
> Yes, basically just define variables in the `config.md`, these can then be used in HTML using `{{fill varname}}`
> 
> If someone gives me more detailed specs of what/where this is desired, I’m happy to put this in place

@tlienart Filed #677 for further discussions

PR #679, opened 2020-03-22
closes #677: adds templating for downloads

### LINKED ISSUE #677: Templating the downloads page
Here are some thoughts for templating the downloads page with version number variables in Franklin so that it is easier to update it for future releases:

1. `stable_release`: x.y.z format
2. `lts_release`: x.y.z format
3. `upcoming_release`: x.y.z-{alpha, beta, rcn} format

The button on the landing page can then simply use `stable_release`. Any other ideas?

tlienart:
**Note**: this requires Franklin >= 0.6.11. 

In the file `config.md` there is now:

```
@def stable_release = "1.4.0"
@def stable_release_short = "1.4"
@def stable_release_date = "March 21, 2020"
@def lts_release = "1.0.5"
@def lts_release_short = "1.0"
@def lts_release_date = "Sep 9, 2019"
```

On the front page and download page there are `{{stable_release}}` and similar which do the insertion, for instance:

**`index.html`** (front page)
```html
<a class="btn btn-success btn-lg btn-block" href="/downloads/" role="button">Download v{{stable_release_short}}</a>
```

**`downloads/index.md`** (main downloads page)

```
## Current stable release: v{{stable_release}} ({{stable_release_date}})
```

and in the table:

```html
<td colspan="3"> <a href="https://julialang-s3.julialang.org/bin/winnt/x86/{{stable_release_short}}/julia-{{stable_release}}-win32.exe">32-bit</a> </td>
```

I'll test all links first then will merge

cc @ViralBShah

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -21,9 +21,9 @@ Different OSes and architectures have varying [tiers of support](/downloads/#cur
 @@
 
 
-## Current stable release: v1.4.0 (March 21, 2020)
+## Current stable release: v{{stable_release}} ({{stable_release_date}})
 
-Checksums for this release are available in both [MD5](https://julialang-s3.julialang.org/bin/checksums/julia-1.4.0.md5) and [SHA256](https://julialang-s3.julialang.org/bin/checksums/julia-1.4.0.sha256) formats.
+Checksums for this release are available in both [MD5](https://julialang-s3.julialang.org/bin/checksums/{{stable_release}}.md5) and [SHA256](https://julialang-s3.julialang.org/bin/checksums/{{stable_release}}.sha256) formats.
 
 @@row @@col-12
 ~~~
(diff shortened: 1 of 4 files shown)


fredrikekre:
Could also add the "Upcoming release" info.

tlienart:
I did initially but @ararslan removed the corresponding section in https://github.com/JuliaLang/www.julialang.org/pull/675 (?) 

Should I bring it back?

fredrikekre:
I mean that we can add the templating for it, and check if those variables are defined.

tlienart:
Right that makes sense, I'll do that

tlienart:
Ok done and tested with and without, I'll do a last check of the links then will merge

downloads/index.md
```diff
@@ -136,6 +136,70 @@ Checksums for this release are available in both, [MD5](https://julialang-s3.jul
 ~~~
 @@ @@
 
+{{ifdef upcoming_release}}
+
+## Upcoming release: v{{upcoming_release}} ({{upcoming_release_date}})
+
+ We're currently testing release candidates for Julia v1.4.0, an upcoming minor release in the 1.x series of releases. We encourage developers and interested users to try it out and report any issues they encounter. As a prerelease, it should not be considered production-ready; it's intended to give users a chance to try out 1.4 with their code before the full release.
```
fredrikekre:
```suggestion
 We're currently testing release candidates for Julia {{ fill ?? }}, an upcoming minor release in the 1.x series of releases. We encourage developers and interested users to try it out and report any issues they encounter. As a prerelease, it should not be considered production-ready; it's intended to give users a chance to try out {{ fill upcoming_release_short }} with their code before the full release.
```

tlienart: nicely spotted, thanks

tlienart:
Here's the (unchanged) preview: https://julialang.netlify.com/downloads/ blc still happy; @fredrikekre happy for me to merge?

ViralBShah:
Seems like this is a new feature for Franklin - so shouldn't Franklin go to 0.7?

ViralBShah:
Your point about the versioning also makes me wonder if we can set up website repos just like Julia package repos with Project.toml etc., so that it is easier to work with.

tlienart:
It's a very small change & I'm working on one extra thing which I'd like to bring in before releasing 0.7 but yes it's coming!

tlienart:
> Your point about the versioning also makes me wonder if we can set up website repos just like Julia package repos with Project.toml etc., so that it is easier to work with.

If I understand your comment properly, that's already a thing, Franklin accepts a site `Project.toml`; for instance https://github.com/tlienart/franklindocs/blob/master/Project.toml

ViralBShah:
Looking great!

ViralBShah:
> > Your point about the versioning also makes me wonder if we can set up website repos just like Julia package repos with Project.toml etc., so that it is easier to work with.
> 
> If I understand your comment properly, that's already a thing, Franklin accepts a site `Project.toml`; for instance https://github.com/tlienart/franklindocs/blob/master/Project.toml

Ah - great! Would it be easy to set up this repo in the same way?

tlienart:
Yes pretty much, I just didn't see the need yet because the only code that's being evaluated is the Rosetta thing which doesn't require any dependency.

But if in the future some blog posts run some Julia code then yes there will be a need for a Project.toml.

Note that there's a catch, currently it's assumed that the Project.toml is valid for the whole website; but if you have code on multiple blog posts (say), then there may be a point in the future where things clash. That may require improvements on Franklin's side to handle multiple toml files but I haven't really thought seriously about how to do  this best 😄

ViralBShah:
Yeah that's a good point. Maybe it forces people to keep any evaluated code up to date, which may be a feature!

ViralBShah:
I guess supporting multiple Julia versions would require that Julia itself be served through BB, which doesn't exist yet.

PR #686, opened 2020-03-26
Over 3000 packages!

PallHaraldsson:
[skip ci]

### DIFF
--- a/index.html
+++ b/index.html
@@ -404,7 +404,7 @@ end
 
       <br>
 
-      <p>Julia has been downloaded over 13 million times and the Julia community has registered over 2,800 Julia packages for community use.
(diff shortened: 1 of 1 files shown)


PallHaraldsson:
By my county there are now 2993 (plus jll packages), but with 15 new packages in the pipeline waiting to be registered, many from today! So, wait with merging a bit, should it take long?

Congratulation.

I have a feeling 13 millions downloads might also be outdated....

tlienart:
thanks

ViralBShah:
I think you can just count the number of directories in the general registry

ViralBShah:
Let's not update number of downloads. I think over 10 million downloads is what we should use going forward.

PR #697, opened 2020-03-30
Fix placement of 64-bit ARM on Downloads page

AnAverageHuman:


ViralBShah:
Thanks!

ViralBShah:
@tlienart While the deploy stage failed (as you had noted before), would it get deployed when I merge?

tlienart:
yes!

tlienart:
(all people who merge are part of the org and so can trigger a GitHub action that relies upon the DEPLOY_KEY; at least that's my understanding)

PR #708, opened 2020-04-04
Update Research to include 2020 list and new entry

bayesthm:


tlienart:
sorry for the delay and thanks!

PR #710, opened 2020-04-11
Add RSS link in /blog

abhishalya:


abhishalya:
Looks like this:
image

_layout/head.html
```diff
@@ -28,7 +28,11 @@
 
 {{ispage /blog/*}}
 <div class="container blog-title">
-	{{isdef title}}<h1>{{fill title}}</h1>{{end}}
+  {{isdef title}}<h1>{{fill title}}
+    <a type="application/rss+xml" href="https://julialang.org/feed.xml">
+      <i style="color:#e2802f" class="fa fa-rss-square"></i>
+    </a>
+  </h1>{{end}}
```
tlienart, line 35: This will show  on *every* blog post, if that's fine then :+1:

tlienart:
Preview: 
- https://julialang.netlify.com/blog/
- https://julialang.netlify.com/blog/2019/12/yao-differentiable-quantum-programming/

_layout/head.html
abhishalya: I think it should occur wherever possible, I thought this would be better than keeping it in navbar.

PR #713, opened 2020-04-14
Add Calendar to the community section 

logankilpatrick:
Hey @tlienart, is it possible to imbed this iframe into the markdown?

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -3,6 +3,10 @@
 
 Please take a moment to read the [Julia Community Standards](/community/standards/). We expect that your participation in any Julia related forum, online or offline, respects these standards.
 
+## Events
+
+<iframe src="https://calendar.google.com/calendar/embed?src=julialang.org_komauaqet14eog9oiv3p6o7pmg%40group.calendar.google.com&ctz=America%2FNew_York" style="border: 0" width="800" height="600" frameborder="0" scrolling="no"></iframe>
+
 ## Diversity and Inclusion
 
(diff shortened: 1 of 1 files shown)


ViralBShah:
Can I see this calendar on the web itself, or do I need to add it to my google calendar?

logankilpatrick:
Yes, if we embed it, you should be able to see it without signing in or anything, but I need to figure out how to embed it since we are using Markdown and not HTML.

tlienart:
Here: https://julialang.netlify.com/community/

ViralBShah:
@logankilpatrick Shows an empty calendar right now. We should populate the calendar before merging this PR.

abhishalya:
It doesn't scale properly on mobile devices. So maybe we should fix that too.

logankilpatrick:
@abhishalya we need some additional code in JS or something like that since by default iframe HTML tags have a specific length and width in pixels that will not auto change based on the platform.

logankilpatrick:
@ViralBShah what other events should be added to the calendar? I added the AMA for Thursday.

ViralBShah:
Shall we add some ofthe events of the past week? What about all those twitch livestreams?

logankilpatrick:
@ViralBShah I added the virtual meetup form last week. Once this goes live, if the folks who did those streams want to go back in and add past streams they will be able to.

logankilpatrick:
I am ready to merge this if you all are

PR #716, opened 2020-04-15
Create hackathons.md

logankilpatrick:


ViralBShah:
Would it be better as a blog post? My main thought is that transient content otherwise gets stale if made part of the regular website.

logankilpatrick:
Maybe medium/long term these should end up as blog posts or a separate section but I just want a short and easy way for people to access this Info and just going to JuliaLang.org/hackathons seems to be the least friction

ViralBShah:
Once we establish a URL, we sort of have to carry it along as people link to it externally at all. So if the plan is to move it soon, it would be best to avoid it. I am not against the idea - just want to consider this before we merge.

BTW, we have a good tool to draw attention. The altert bar on top of julialang.org. We can make it announce the hackathon. We can also do bit.ly or goo.gl short URLs for easy sharing. 

We should also share the hackathon as a meetup on the big Julia meetup groups.

logankilpatrick:
We would keep the hackathons landing page, the content would just change as things formalize over time and we figure out a better way to list these events.

PR #717, opened 2020-04-15
Better instructions (for Ubuntu at least)

PallHaraldsson:
[skip ci]

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -86,10 +86,10 @@ tar -xvzf julia-x.y.z-linux-x86\_64.tar.gz
 This will extract the files to a folder named `julia-x.y.z`. We would refer this as `<Julia directory>`. To run Julia, you can do any of the following:
 
 *   Create a symbolic link to `julia` inside a folder which is on your system `PATH`
-*   Add Julia's `bin` folder to your system `PATH` environment variable
+*   Add Julia's `bin` folder (with full path) to your system `PATH` environment variable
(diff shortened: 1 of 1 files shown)


PallHaraldsson:
[The failed check must be unrelated?! And why happening since I did "skip ci"?]

A. Strangely with tilde (~) for home older WILL work to invoke julia, but not in all contexts.

FYI: I was trying out from python3:

>>> import TopLoader

and it wouldn't find julia. I guess because:

`$ which julia
`

returned nothing. I'm not sure it's a bug in `which` (I guess it is?!), but it took me a lot of time to debug why, and I think it would be helpful for other users to know the surest way.

I was a little rusty (used a symbolic link before, and decided to try following the instructions to the letter), and .bash_profile (alone) doesn't actually work.

B. Before I got an error (ELF) because julia installed with snap is only 32-bit. Maybe warn against using it in this here (it took a lot of time figuring out why the last time), as it also works only for most situations.

C. [Unrelated] I actually get a different error with TopLoader now but I'm familiar with that one (it's unrelated to the PATH problem):

> >>> import TopLoader
> TopLoader V 0.0.1
> Happy Loading!
> -----------------
> Traceback (most recent call last):
> [..]
> julia.core.UnsupportedPythonError: It seems your Julia and PyJulia setup are not supported.
> [..]
>

tlienart:
> [The failed check must be unrelated?! And why happening since I did "skip ci"?]

yeah you can ignore that

PR #724, opened 2020-04-16
Bump stable release to 1.4.1

ararslan:
There are no ARMv7 binaries for 1.4.1.

### DIFF
--- a/config.md
+++ b/config.md
@@ -11,9 +11,9 @@
 @def author = ""
 
 <!-- Templating of the Downloads -->
-@def stable_release = "1.4.0"
+@def stable_release = "1.4.1"
 @def stable_release_short = "1.4"
-@def stable_release_date = "March 21, 2020"
+@def stable_release_date = "April 14, 2020"
 @def lts_release = "1.0.5"
 @def lts_release_short = "1.0"
 @def lts_release_date = "Sep 9, 2019"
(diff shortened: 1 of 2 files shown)


ViralBShah:
@staticfloat When might we get the armv7 binaries? If they are being built, it may be ok to hold to have everything.

ararslan:
I attempted to build them but wasn't able to get a buildbot build through. It seems no recent ARMv7 binaries have made it through at all. I thought the point of bumping the tier back to 3 for ARMv7 was that we wouldn't hold releases based on the availability of binaries, since we typically don't expect them to exist.

ViralBShah:
Ok. But how did we get the 1.4 binaries?

ararslan:
Elliot magic

staticfloat:
We bumped some deps between 1.4.0 and 1.4.1 that don't have quite the right names for 1.4.  I've duplicated some names, so now we've got a build going through here: https://build.julialang.org/#/builders/5/builds/10723

ararslan:
Looks like the build failed. :/

staticfloat:
I fixed it: https://s3.amazonaws.com/julialangnightlies/pretesting/linux/armv7l/1.4/julia-381693d3df-linuxarmv7l.tar.gz

ararslan:
Everything is now in place and the checksums have been updated.

ViralBShah:
Love the PR! So much better than it used to be! Thanks @tlienart

PR #725, opened 2020-04-20
Document limitation for pre-Bulldozer AMD CPUs

PallHaraldsson:
[skip ci]

https://github.com/JuliaLang/julia/issues/35215#issuecomment-616523407

In short, I would additionally suggest 1.3 as LTS, unless 1.4 or newer can be compiled to work (and published here) on same CPUs. See more in the comment, on two possible LTS versions.

[I'm closing since no interest.]

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -86,6 +86,7 @@ Checksums for this release are available in both [MD5](https://julialang-s3.juli
 ~~~
 
 Checksums for this release are available in both, [MD5](https://julialang-s3.julialang.org/bin/checksums/julia-{{lts_release}}.md5) and [SHA256](https://julialang-s3.julialang.org/bin/checksums/julia-{{lts_release}}.sha256) formats.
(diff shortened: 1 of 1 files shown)


ViralBShah:
We don't want this in a downloads page and need to do this for all architectures. The message needs to be a bit more specific too.

PallHaraldsson:
What do you mean "need to do this for all architectures"? AMD (older than 2011) are the earliest CPUs I know of failing on 1.4. Older Intel should work, but I'm not sure older still Intel works on any version. I did not look into Via, or POWER etc. just thought this might be helpful here..., doesn't rule out documenting also elsewhere.

Note however LTS suggestion. Then message could be simplified to "some older computers still cork on Julia LTS".

ViralBShah:
Please stop mentioning the 1.3 LTS. We will not be doing this.

PallHaraldsson:
I intentionally had "very old" not specific, as if I mention a year, it's about when the CPU arch is made, and I'm not sure for how long sold (after discontinued). People may not know or care their CPU is Bulldozer-based or older.

PR #726, opened 2020-04-20
Added GSoD to Jsoc section

logankilpatrick:


jsoc/index.md
```diff
@@ -26,6 +26,13 @@ Those participating in the program can check out [the getting started page](/lea
 Are you interested in Julia but not super familiar with the language?
 Head over to [JuliaAcademy](https://juliaacademy.com) to take some free online Julia classes.
 
+## Google Season of Docs (GSoD)
+
+The Julia organization is applying to participant in [Google's Season of Docs](https://developers.google.com/season-of-docs).
```
tlienart, line 31: applying to participate ?

jsoc/index.md
```diff
@@ -26,6 +26,13 @@ Those participating in the program can check out [the getting started page](/lea
...
+The Julia organization generally fields a large number of students mentored by open source developers, university professors at leading institutions from all around the world, and technical writers in the Julia Community. 
+Students who are accepted into this prestigious program are paid a stipend to contribute open source documentation around  around Julia and its package ecosystem.
```
tlienart, line 33: around  2x, contribute _to_ ? (maybe not)

jsoc/index.md
logankilpatrick: Yeah, we are not in yet, but we need to have out projects list publicly available

jsoc/index.md
logankilpatrick: Fixed this, thanks for pointing it out

jsoc/index.md
tlienart: oh sorry I meant `participant <> participate`

abhishalya:
It would be really nice to have Julia participating in GSoD. Although the number of people being selected is usually low (like max 2 per org as mentioned in their FAQs).

logankilpatrick:
Who knows what it will be like in the future though, we only have 1 session worth of data at this point. Perhaps the program will grow :)

PR #728, opened 2020-04-22
Updated and re-scope GSoD Projects

logankilpatrick:


logankilpatrick:
This is a WIP, but unless there is something major that needs to be fixed, we can merge. I'm planning to source the rest of the work to the project's mentors.

tlienart:
There are some mild conflicts with the project.md page due to #747 which should probably take precedence. But otherwise this should probably be merged?

PR #729, opened 2020-04-22
Add Why Julia page

logankilpatrick:
Will eventually be linked on the home page with the code example(s).

### DIFF
--- a/whyjulia.md
+++ b/whyjulia.md
@@ -1,0 +1,7 @@
+# Why Julia?
+
+"Why Julia?" is a series of community made videos where people talk about the reason(s) that they use the Julia Programming Language.
+
+~~~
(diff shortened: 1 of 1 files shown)


ViralBShah:
This video is more of a how rather than a why. I would prefer to have good content ready before we merge this PR. Can we make it a "draft PR" in the meanwhile?

logankilpatrick:
This individual chose to not just tell but __show__ why Julia is powerful in the case of Package Dev. Converting to draft PR until we get more videos.

ViralBShah:
Ah yes - I see how that fits. Perhaps we just need the right text captions to set the context. I do like that it is a 4 minute video. Sorry - my intention was not to sound harsh about the contribution (revisiting my comment).

PR #730, opened 2020-04-22
Add Why Julia page

logankilpatrick:
Will eventually be linked on the home page with the code example(s).

Waiting on more videos in the playlist.

### DIFF
--- a/whyjulia.md
+++ b/whyjulia.md
@@ -1,0 +1,7 @@
+# Why Julia?
+
+"Why Julia?" is a series of community made videos where people talk about the reason(s) that they use the Julia Programming Language.
+
+~~~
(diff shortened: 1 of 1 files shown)


ViralBShah:
Do we still want to leave this open?

logankilpatrick:
We can merge or close it, I still think the idea is valid and I'll continue to follow up on it. 

Getting it merged may inspire others to do a video

ViralBShah:
I don't think we should merge with incomplete content. We need a few high quality videos from diverse participants. Some from users, some from developers or well known people and such would greatly help.

We can leave it open.

logankilpatrick:
Can always re-visit if the initiative takes off. Perhaps something to ask of folks for JuliaCon 2021

PR #738, opened 2020-04-26
bringing automath/code back (patched Franklin)

tlienart:
this was in before, then removed bc there  was an issue noticed by Chris, now this is fixed so bringing it back.

It allows people who would write blog posts to forget  adding `hascode=true` or `hasmath=true` and let Franklin figure it out.

https://julialang.netlify.com


abhishalya:
Wow, this would be very nice!

PR #740, opened 2020-04-28
add a /packages page

StefanKarpinski:


ararslan:
Might be nice to add a sentence or two on how the provided options here differ, since it's not immediately obvious.

StefanKarpinski:
Since we're redirecting from pkg.julialang.org to a 404 I wanted to get something up ASAP.

StefanKarpinski:
Anyone have any idea how to deploy the site these days? Is merging the PR enough?

logankilpatrick:
Yes

StefanKarpinski:
Ok, well, that's horrifically broken, but I have no idea how to preview locally or see the build preview, so I guess at least there's something when you go to https://pkg.julialang.org now.

StefanKarpinski:
Nope, that's broken too because it redirects to https://julialang.org/packages without the trailing slash which Franklin doesn't consider the same as with the trailing slash, so we're really in an unfortunate state here. I guess anyone going to pkg.julialang.org will just think it's down. Not much I can do about it at the moment.

logankilpatrick:
I got it

logankilpatrick:
#741

logankilpatrick:
https://julialang.netlify.app is where you can see the build previews

fredrikekre:
> Nope, that's broken too because it redirects to https://julialang.org/packages without the trailing slash which Franklin doesn't consider the same as with the trailing slash

I think that is up to the webserver, not the site generator. https://julialang.org/packages redirects to https://julialang.org/packages/ for me at least.

PR #741, opened 2020-04-28
Update and rename index.html to index.md

logankilpatrick:
Fixes the packages page


StefanKarpinski:
I don't really get why this was broken, I copied the `learning` directory and then edited...

StefanKarpinski:
But thanks for fixing it for me :)

logankilpatrick:
Must have been an old learning directory as the new one uses markdown

StefanKarpinski:
Seems I was on a branch and did a pull to update which updated nothing, then rebased on top of master, which lead to this broken situation. Thanks for rescuing me.

logankilpatrick:
Also, if you want the url julialang.org/packages rather than julialang.org/packages/ we need to take the file out of the packages folder and make a packages.md file in the root of the repo.

StefanKarpinski:
Our other top-level pages have the trailing slash, so for consistency I think we should do that.

fredrikekre:
> Also, if you want the url julialang.org/packages rather than julialang.org/packages/ we need to take the file out of the packages folder and make a packages.md file in the root of the repo.

What? Both those options should be equivalent (`packages/index.html`).

logankilpatrick:
@fredrikekre I don't make the rules I just play by them :) But actually, all the pages within folder require an extra "/" in order to access them, this seems standard/normal given they are in a subdir

fredrikekre:
> @fredrikekre I don't make the rules I just play by them :)

I don't understand wat you mean. both `/packages.md` and `packages/index.md` should result in the exact same thing, that is `packages/index.html`. The webserver/browser redirects `packages` to `packages/`. If that didn't happen for you then it is probably a caching problem.

PR #754, opened 2020-05-01
Add Robust Footer 

logankilpatrick:
Making a robust footer for the site

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -39,10 +39,6 @@
   display: block;
 }
 
-.footer-heading{
-  color:#4D64AE;
-}
-
 #menu ul{
     width: 100%;
     background: #cacaca;
@@ -482,6 +478,11 @@ h4 {
 a {
 	color: #005bc0;
 }
+
+.footer-heading{
+  color:#4D64AE;
+}
+
 
 table th {
   background-color: #eee;
--- a/_css/app.css
+++ b/_css/app.css
@@ -46,6 +46,23 @@
     padding: 0;
 }
 
+.footer-copy {
+    padding-top: 2em;
+    margin-left: auto;
+    margin-right: auto;
+}
+.footer-copy ul > a {
+    font-size: 95%;
+}
+.footer-copy ul > a:first-of-type {
+    font-size: 115%;
+    font-weight: bold;
+}
+#footer-bottom {
+    font-size: 90%;
+    padding-bottom: 2em;
+}
+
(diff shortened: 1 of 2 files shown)


logankilpatrick:
Leaving this as a real PR rather than a draft so I can view deploy previews, though I should actually probably do this locally so I will convert to a draft.

ViralBShah:
Let's have it deploy to a preview.

logankilpatrick:
I intend to use https://www.python.org 's footer for reference.

logankilpatrick:
https://julialang.netlify.app

ViralBShah:
Hmm, needs some work.

logankilpatrick:
Indeed, much more is needed. I have been trying to distract myself from all the other things I need to do this weekend so I will build this out to be extensive/polished sometime this weekend or during next week.

tlienart:
I think it's a good start! some concrete points:

- font size is too small (use a separate div for the list so that you have a fine-grain control over it)
- the margin-bottom should be bigger
- probably use two columns with max 4 elements per column
- I imagine you want the points to be links

logankilpatrick:
It's not perfect but looks good as a first pass

tlienart:
Looks much better!

A few minor comments:

* Make list items a bit smaller than the blue titles
* Use a lighter shade of blue so that the constrast is better with the background, possibly use bold as well, 
* Add padding at the top

```css
.footer-copy {
    padding-top: 2em;
    margin-left: auto;
    margin-right: auto;
}
.footer-copy ul > a {
    font-size: 95%;
}
.footer-copy ul > a:first-of-type {
    font-size: 115%;
    font-weight: bold;
}
#footer-bottom {
    font-size: 90%;
    padding-bottom: 2em;
}
```

logankilpatrick:
Where did you add that CSS in? At the bottom of the file or in a particular location? I was running into some issues with some CSS getting overwritten based on where I put it in the file.

tlienart:
At the bottom of `franklin.css` but you could put it at the bottom of `app.css`

fredrikekre:
Did the CSS not get included? Now the footer just looks like an unreadable wordcloud.

tlienart:
I'll open a PR

tlienart:
ok this is quite annoying, there's some external resource, font or something that doesn't  get loaded properly and somehow craps on the CSS... I don't have the time to look into this right now but will open an issue

PR #759, opened 2020-05-01
Don't fail the build on pull requests made from forks

DilumAluthge:


logankilpatrick:
Thanks! Will wait and see if it does not error and then merge.

DilumAluthge:
@tlienart Might be useful to add this to the default template

Issue #762, opened 2020-05-02
Blog style


Keno:
I'm not sure I like the way the blog looks on the new website. For reference, here's how it used to look, so that's what I'm using as a reference:

https://web.archive.org/web/20160322095210/http://julialang.org/blog/2016/03/parallelaccelerator

I think the new blog should have bigger margins, particularly on large monitors. I have a large external monitor and having the text stretch from all the way across makes it bit hard to read, because it requires a lot of head movement. Usually blogging systems restrict the width of the text to 700px or so (e.g. medium.com). The old font also seems slightly better to me, but that may just be an artifact of not liking the width of the text. There are other differences, like code blocks used to be indented, rather than out-dented as they are now (we do have syntax highlighting now though, which is of course better). Also, the size of the embedded images on the blog is a bit jarring. That is probably another artifact of the width, since they'll scale to fill the width, which makes them larger than the author probably intended.

tlienart:
Sounds good, btw I ported everything from existing style so I plead non guilty.

So actionable items would be

* change font for Georgia or similar serif font
* limit the max content width to 700px 
* make code blocks narrower than the body

all these are pretty simple changes, there's one question as to whether this should only be applied on blog pages, all pages, or all pages minus front page?

### Part of the way

Here's one way to do all of it apart from the media queries (width) which are a bit more finicky.

**config.md**

```
@def div_content = "container main"
```

**_layout/head.html**

```
{{ispage /blog/*}}
  {isnotpage /blog/index}}
    <style>
	  .main {
		  font-family: Georgia;
	  }
	  .main pre {
		  margin-left: auto;
		  margin-right: auto;
		  width: 90%;
	  }
    </style>
  {{end}}
{{end}}
```

PR #765, opened 2020-05-03
suggested blog layout in  light of #762

tlienart:
Affects (only) blog pages

* Serif font (Georgia)
* Slightly bigger font
* Slightly bigger inline code
* Narrower page

cc @Keno  ( #762 )

I'll also mention this on Slack#website so that people can comment

PS: I've not tested this to death so feedback and help welcome.

### DIFF
--- a/_layout/head.html
+++ b/_layout/head.html
@@ -18,6 +18,26 @@
   </style>
   {{end}}
 
+  <!-- Specific style for blog pages (except the /blob/index) -->
+  {{ispage /blog/*}}{{isnotpage /blog/index}}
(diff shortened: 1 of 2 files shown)


Keno:
Yes, I think this looks much better. That said, I don't think the font size increase is really necessary - it feels a bit big. If anything I think it could be a ticks smaller.

tlienart:
done 

Screen Shot 2020-05-04 at 09 40 59

Keno:
I like it. Let's do it. We can always tweak after.

PR #770, opened 2020-05-04
CSS for footer

tlienart:
Not sure what happened in #754 , the CSS was there but somehow didn't appear on the site. This just re-triggers it.


tlienart:
oh ok I know, this is actually subtle, the franklin.css file does not get loaded because there's something above it that fails, I'll remove it and put the  CSS elsewhere.

PR #773, opened 2020-05-04
Add preview video for rr post

Keno:
@tlienart how do I add extra `meta` tags on a per-page basis? This would have to go in as an `og:video` tag on the rr post page.


Keno:
cc @ViralBShah who'd requested a video preview for social media

tlienart:
Just modify `_layout/head.html` and put the meta stuff in a `{{ispage /blog/2020/...}} ... {{end}}`

Filling `/...` with the rest of the path to your blog post

Keno:
I think it would be better to allow that to be set on each blog page, since Ideally we'd want to set relevant meta tags on a per-post basis. Editing the head HTML for each post seems like a suboptimal solution.

ViralBShah:
Right setting it per blog page would be ideal, if possible.

tlienart:
Ok that’s also doable, can you just tell me the precise HTML you would like to appear? We can just have a local page var on blog posts which, if defined, gets displayed in the head. Looks like

`{{ifdef meta}} ... HTML ... {{end}}`

And the HTML can be shaped by the meta var.

Keno:
Allowing arbitrary meta tags seems most general. There's a bunch of interesting ones not just video. See https://ogp.me/

tlienart:
Please see #776

PR #776, opened 2020-05-05
[draft] adding functionality for meta information for blog pages

tlienart:
Blog pages which require specific additional information should now just add a `meta` page variable like so:

```
@def meta = ("property" => "og:video",
             "content" => "http://example.com/bond/trailer.swf")
```

where the first element of the pairs are valid meta attributes (property, content, name, ...). 
The above will lead to the inclusion  of

```html
<meta property="og:video" content="http://example.com/bond/trailer.swf" >
```

@Keno this is good to go apart from the `content` part in your blog post, if you just tell me what URL you want in there (or message), I'll fix it and merge.

### DIFF
--- a/_layout/head.html
+++ b/_layout/head.html
@@ -38,6 +38,9 @@
   </style>
   {{end}}{{end}}
 
+  <!-- OGP Metadata -->
+  {{ifdef meta}}{{meta}}{{end}}
+
 </head>
 
 <body>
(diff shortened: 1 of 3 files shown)


Keno:
Ideally, we'd have support for multiple meta tags.

Keno:
> if you just tell me what URL you want in there (or message), I'll fix it and merge.

The URL should point to the video in my other PR

tlienart:
Both done you can now specify the specs as suggested above or pass a list of such specs.

Keno:
LGTM, let's merge the other PR first, so the video is actually there.

Issue #771, opened 2020-05-04
Footer CSS not loaded properly + issue in mobile mode


tlienart:
The CSS of the  footer is not loaded properly. If the site is built locally, it will look fine but online it looks like this:

<img width="675" alt="Screen Shot 2020-05-04 at 10 51 09" src="https://user-images.githubusercontent.com/10897531/80949911-3857bf00-8df5-11ea-9f17-33f93e9335a8.png">

when looking at the console, it looks like some resource doesn't get loaded properly which might be a cause (not sure), I tried shifting the order of stylesheets etc around but wasn't successful.

Someone with more experience with  web dev might be able to see what's going on here.

**Edit**: there's also an issue on mobile mode where the footer doesn't span the whole width and doesn't look good. That should probably be another issue.

DilumAluthge:
Closed by #784 ?

tlienart:
yes thanks!

PR #784, opened 2020-05-07
automatic bloglist and fixes of pifont

tlienart:
* the list of blog on the blog main page is now generated automatically via `{{blogposts}}`, the corresponding function is defined in `utils.jl` if people want to adapt it.
* @cormullion 's pifont blog post had a few issues, mainly it still used indented code blocks which now must be explicitly opted in (and mess things up otherwise) that's now fixed.
* the issue with the footer CSS #771 is fixed, the CSS paths were not specific enough and ended up clashing

Preview:
* https://julialang.netlify.app/blog/
* https://julialang.netlify.app/blog/2020/05/rr/
* https://julialang.netlify.app/blog/2018/03/pifonts/

Side note: @keno the metadata on your blog post was improperly specified (use of `=` instead of `=>` and some properties were not in `"..."`). This caused improper build of the site. Now it's invisible because @DilumAluthge made it such that CI doesn't show failure if the site fails to build but that also means we have to make sure green CI actually means the site is built properly. One way to check this is to check the netlify preview & make sure your changes were added properly.


DilumAluthge:
I think that the only allowed failure is if the pushing of the preview fails.

That is, if the build passes but pushing the preview fails, then CI should be green..

But if the build fails, CI should be red.

@tlienart Do you have an example build where the site failed to build but CI was green?

DilumAluthge:
We only set `continue-on-error: true` for the `Deploy (preview)` step.

If the `Build site` step fails, CI should be red.

https://github.com/JuliaLang/www.julialang.org/blob/2c1c3bcefb483e91fcb029d4a47ceaebab19b3e9/.github/workflows/deploy.yml#L39-L59

DilumAluthge:
> Side note: @Keno the metadata on your blog post was improperly specified (use of `=` instead of `=>` and some properties were not in `"..."`). This caused improper build of the site.

So I just checked the CI of #764. The build itself passed without any errors, i.e. Franklin didn't seem to detect any problems with the build.

tlienart:
Thanks for checking Dilum, I think that's fine, maybe the key message  here should be to verify the preview to make sure it looks as it should before merging into master, if possible.

DilumAluthge:
> Thanks for checking Dilum, I think that's fine, maybe the key message here should be to verify the preview to make sure it looks as it should before merging into master, if possible.

Yeah, the bottom line is always look at the preview before you merge.

I just wanted to clarify that e.g. the problem with Keno's blog post PR (#764) doesn't have anything to do with the changes we made in #759. That is, even if we hadn't merged #759, the CI would still have been green on #764, because Franklin successfully built #764 without any errors.

Again, like you said, the core takeaway is that even if Franklin successfully builds your PR without any errors, and CI is green, you should still look at the preview before merging.

tlienart:
right, now the footer is also done, closes #771

Issue #787, opened 2020-05-08
Link to source in the footer


fredrikekre:
I think it would be nice to link the the source repo in the footer. It makes it easier to find it if you spot a typo or w/e.

I propose adding this in the footer instead of an "Edit on GitHub" button at the top, which is also quite common. I think a note in the footer would fit better with the rest of the layout.

tlienart:
Should the link point to the exact page or generally to the repo? In franklindocs for instance: https://franklinjl.org/syntax/divs-commands/ the link at the top points to the specific file.

tlienart:
#787 is now in with a floating edit button

PR #790, opened 2020-05-08
Add a floating link to edit the source of pages

tlienart:
Suggested by @fredrikekre in #787 

~~**Now**~~ (see screenshot further below, floating link)

The first link (on GitHub) links to the source of the page for instance:

* front page => https://github.com/JuliaLang/www.julialang.org/blob/master/index.html
* learning page => https://github.com/JuliaLang/www.julialang.org/blob/master/learning/index.md


Keno:
I'm wondering if we could make it more prominent, without being obstrusive. Maybe a small button the one if the sides or something. I'm sure there's prior art somewhere.

tlienart:
Yeah fair enough it's not visible enough. 

I'm not too sure where to place said button though. It could be on the same line as the title, a bit like where the RSS button used to be but right-aligned? Or should it be kept in the footer somewhere and just made more visible there?

Keno:
I was imagining something like a floating action button ala https://materializecss.com/floating-action-button.html

tlienart:
Ok here it is. The button appears on pages that  aren't  the front page or the blog front page as I thought that would be a distraction. Otherwise it appears on all pages provided the width of the page is big enough. Styling etc could be adapted.

Preview should be at https://julialang.netlify.app screenshot below: (bottom right button)

Screen Shot 2020-05-09 at 11 38 54

Keno:
I think the button is a little large. Maybe reduce to 75% of the current size?

tlienart:
sounds good

PR #794, opened 2020-05-10
WIP: blog post on invalidations

timholy:
Several of us have been poking at invalidations lately. I've finally gotten some decent tooling developed and my head wrapped around sources. Here's a summary of my findings.


blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,518 @@
...
+```julia
+julia> using SIMD
+
+julia> @time display(plot(rand(5)))
+  7.238336 seconds (26.50 M allocations: 1.338 GiB, 7.88% gc time)
+```
+
+Because so much got invalidated by loading SIMD, Julia had to recompile many methods before it could once again produce a plot, so that in terms of time it was almost as expensive as the first usage.
+The size of the effect varies substantially depending on what task you are trying to achieve and what packages you load to do the invalidation.
+(If you do this with StaticArrays rather than SIMD there's no increase in latency,
```
ChrisRackauckas, line 273: Maybe demonstrate this first. I feel like the reader might be getting here like "oh no, Julia is doomed!" so it might be good to start by saying "loading some packages doesn't effect ... . Here StaticArrays ... however notice SIMD ..."

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,518 @@
...
+| SIMD | 2799 |
+| StaticArrays | 2852 |
+| Optim | 3171 |
+| Images | 3638 |
+| Flux | 3697 |
+| Plots | 4002 |
+| DataFrames | 4048 |
+| JuMP | 4666 |
+| Makie | 6118 |
+| DifferentialEquations | 6777 |
```
ChrisRackauckas, line 236: oh wow. Can I see this list? I'm a bad bad man.

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,518 @@
...
+`reduce_empty(::typeof(Base.mul_prod), ::Type{F}) where F<:FixedPoint` is strictly more specific than `reduce_empty(::Function, ::Type{T} where T)`.
+It's curious that this is listed as a backedge of `reduce_empty(::Base.BottomRF{typeof(max)}, ::Type{VersionNumber})`, since the latter is a concrete signature on its own.
+This case too seems like something that should be fixed in Julia's invalidation logic.
+
+Moving backward, we get to `sizeof(::Type{X}) where X<:FixedPoint`.
+Simply put, this looks like a method that we don't need; perhaps it dates from some confusion, or an era where perhaps it was necessary.
+So we've discovered the first place where a developer could do something to productively decrease the number of invalidations, in this case by just deleting the method.
+
+You'll also notice one example where the new method is *less specific*.
+It is not clear why such methods should be invalidating, and this may be a Julia bug.
```
ChrisRackauckas, line 424: This whole section was great.

blog/2020/05/invalidations.md
ianshmean: It's not future-proof given the reporting will likely change, but you can just throw this action into a PR for the report https://github.com/ianshmean/Plots.jl/pull/1/files

timholy: You're not bad, it's mostly a consequence of having lots of dependencies that extend low-level functionality. (Packages that are essentially totally isolated do not cause invalidations, it's when you start extending functionality that this becomes an issue.) It's not very hard to generate this yourself if you can build Julia, but to save you the trouble I've posted it [here](https://gist.github.com/timholy/88fe5376f0d1460b9b9d9bd253f51b99). About 50% of the nominal invalidations come from just 7 methods (out of 90 total that cause invalidations). And the packages for these 7 are:
```
 StaticArrays
 StaticArrays
 Intervals
 Unitful
 ProgressLogging
 SymbolicUtils
 Intervals
```

I should probably develop printing options that, e.g., allow you to cull the ones with few children, but for now this is the state-of-the-art.

Once Julia stops considering ambiguous type-intersections as invalidating, your total will go down a lot (estimated to fix about two-thirds of "your" total count).

ChrisRackauckas: StaticArrays seems to be showing up in a lot of these. If that's in the sysimage, I wonder how many downstream packages get a faster compile time (+ the ambiguous fix).

KristofferC: > If that's in the sysimage

You can just put it in your sysimage and try (PackageCompiler.jl)

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,518 @@
...
+if isa(x, Bool)
+    value = 2
+elseif isa(x, Int)
+    value = 1
+else
+    value = f(x)::Int
+end
+```
+
+This is [union-splitting], a key optimization for performance in the face of uncertain types.
```
KristofferC, line 129: While union splitting is indeed important I think this specific optimization (union splitting based on the number of methods on `f`) is not really a key optimization. I think most people would expect (and be fine) with a dynamic call here. After all, you're are just one based method overload away from getting `Any` anyway, so relying on this optimization feels pretty brittle.

The important case is imho when you infer that the input type is a union and then split on that.

timholy: I rephrased it a bit, see if this is better.

timholy:
I greatly expanded the information about how to fix problems and added an overall summary.

Having chosen a demo problem to investigate, it seemed reasonable to also submit the fix, which is https://github.com/JuliaLang/julia/pull/35839 (it will also require some changes to Pkg).

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,762 @@
+@def authors = "Tim Holy"
+@def published = "11 May 2020"
+@def title = "Analyzing sources of compiler latency in Julia: method invalidations"
+@def rss_pubdate = Date(2020, 5, 11)
+@def rss = """Julia is fast, but compiling Julia code takes time. This post analyzes why it's sometimes necessary to repeat that work, and what might be done to fix it."""
+
+
```
tlienart, line 7: You could (but don't have to) add a `\toc` here to insert a table of content. Use `@def mintoclevel = 2` and `@def maxtoclevel = 3` to control depth.

Btw all your section headers are level 1 (`#`) which will look a bit funny because the title of the blog post is level 1, you might want to increase the level of everything by one `#`

timholy:
Thanks @tlienart! Any tips on controlling text color? It's not necessary, but the diagnostics like `more specific` print in cyan, and it might be fun to replicate that.

tlienart:
No unfortunately that would require having a specific highlighter for code output, doable but not done :)

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,784 @@
...
+However, one more interesting case we've not seen before is
+
+```julia-repl
+julia> tree = trees[end-7]
+insert unsafe_convert(::Type{Ptr{T}}, m::Base.RefValue{FA}) where {N, T, D, FA<:FieldArray{N,T,D}} in StaticArrays at /home/tim/.julia/packages/StaticArrays/mlIi1/src/FieldArray.jl:124 invalidated:
+   mt_backedges: signature Tuple{typeof(Base.unsafe_convert),Type{Ptr{_A}} where _A,Base.RefValue{_A} where _A} triggered MethodInstance for unsafe_convert(::Type{Ptr{Nothing}}, ::Base.RefValue{_A} where _A) (159 children) more specific
+```
+
+In this case, the signature that triggered the invalidation, `Base.unsafe_convert(::Type{Ptr{_A}} where _A, ::Base.RefValue{_A} where _A)`,
+has been only partially specified: it depends on a type parameter `_A`.
```
mbauman, line 520: Isn't this two separate `_A` type parameters?  Or is it actually a confusing printing of the diagonal dispatch `Base.unsafe_convert(::Type{Ptr{_A}}, ::Base.RefValue{_A}) where _A`?

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,784 @@
...
+5-element Array{SnoopCompile.InstanceTree,1}:
+ MethodInstance for unsafe_convert(::Type{Ptr{Nothing}}, ::Base.RefValue{_A} where _A) at depth 1 with 0 children
+ MethodInstance for unsafe_convert(::Type{Ptr{T}}, ::Base.RefValue{Tuple{Vararg{T,N}}}) where {N, T} at depth 1 with 2 children
+ MethodInstance for _show_default(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Any) at depth 1 with 113 children
+ MethodInstance for _show_default(::IOContext{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Any) at depth 1 with 37 children
+ MethodInstance for _show_default(::IOContext{REPL.Terminals.TTYTerminal}, ::Any) at depth 1 with 2 children
+```
+
+and see all the `MethodInstance`s that called this one.
+You'll notice three `_show_default` `MethodInstance`s here;
```
mbauman, line 538:
```suggestion
You'll notice three `_show_default` `MethodInstance`s with the bulk of the children here;
```

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,784 @@
...
+a little digging reveals that this is defined as
+
+```
+function _show_default(io::IO, @nospecialize(x))
+    t = typeof(x)
+    ...
+end
+```
+
+So the `@nospecialize` annotation, designed to reduce the number of cases when `_show_default` needs to be recompiled, causes the methods *it* uses to become triggers for invalidation.
```
mbauman, line 548: Perhaps I'm dense but it wasn't immediately obvious to me why this was the case.  Speaking out loud, is this because within this method body, `x::Any`... and thus any time you `f(..., x, ...)` for `f`s with small method tables it'll inline and/or union-split?

How's this related to ambiguous cases?

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,784 @@
...
+Replacing `keys(dct)` with `Base.KeySet(dct)` (which is the default return value of `keys`) eliminated a very consequential invalidation, one that triggered seconds-long latencies in the next `Pkg` command after loading Revise.
+The benefits of this change in Pkg's code went far beyond helping Revise; any package depending on the OrderedCollections package (which is a dependency of Revise and what actually triggered the invalidation) got the same benefit.
+With these and a few other relatively simple changes, loading Revise no longer forces Julia to recompile much of Pkg's code the next time you try to update packages.
+
+### Redirecting call chains
+
+Let's return to our FixedPointNumbers `reduce_empty` example above.
+A little prodding as done above reveals that this corresponds to the definition
+
+```julia-repl
```
mbauman, line 629: A refresher of the inserted method is helpful here (I had to scroll back up to grok this section)

```suggestion
```julia-repl
julia> tree = trees[end-1]
 insert reduce_empty(::typeof(Base.mul_prod), ::Type{F}) where F<:FixedPoint in FixedPointNumbers at /home/tim/.julia/packages/FixedPointNumbers/w2pxG/src/FixedPointNumbers.jl:225 invalidated:
    backedges: MethodInstance for reduce_empty(::Function, ::Type{T} where T) triggered MethodInstance for reduce_empty(::Base.BottomRF{typeof(max)}, ::Type{VersionNumber}) (136 children) more specific

```

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,784 @@
...
+   backedges: MethodInstance for reduce_empty(::Function, ::Type{T} where T) triggered MethodInstance for reduce_empty(::Base.BottomRF{typeof(max)}, ::Type{VersionNumber}) (136 children) more specific
+```
+
+`reduce_empty(::typeof(Base.mul_prod), ::Type{F}) where F<:FixedPoint` is strictly more specific than `reduce_empty(::Function, ::Type{T} where T)`.
+This might look like one of those "necessary" invalidations.
+However, even though it's marked "more specific," the new method `reduce_empty(::typeof(Base.add_sum), ::Type{F}) where F<:FixedPoint` can't be reached from a call `reduce_empty(::Base.BottomRF{typeof(max)}, ::Type{VersionNumber})`:
+
+```julia-repl
+julia> mi, node = tree[:backedges, 1]
+MethodInstance for reduce_empty(::Function, ::Type{T} where T) => MethodInstance for reduce_empty(::Base.BottomRF{typeof(max)}, ::Type{VersionNumber}) at depth 0 with 143 children
```
mbauman, line 472: I'm curious — why does this print this as a `MethodInstance for reduce_empty(::Base.BottomRF{typeof(max)}, ::Type{VersionNumber})` if Julia only specialized for `::Type{T} where T`?  I thought the `MethodInstance`s reflected the specializations.

tkf:
Just FYI, I proposed to add `reduce_empty(op, Union{})` as it actually fixes a bug https://github.com/JuliaLang/julia/pull/35843.  I'm just mentioning it here since it may interact with the "correctness" of the blog post for Julia 1.6 (e.g., the result of the introspection can be different after merging it?).

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,784 @@
...
+) => Int64
+```
+
+This may seem a lot more complicated, but the take-home message is actually
+quite simple. First, look at those `arrayref` statements: they are annotated
+`Any`, meaning that Julia can't predict in advance what they will return.
+Immediately after each reference, notice that there are two `isa` statements
+followed by a `ϕ`; all of this is essentially equivalent to
+
+```
```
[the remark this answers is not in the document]
mbauman: Oh yes, yes it is. https://github.com/JuliaLang/www.julialang.org/pull/794#discussion_r423015634

blog/2020/05/invalidations.md
timholy: Yeah, this may be a problematic example. I'll try to find something better.

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,784 @@
...
+
+you'll see a much longer output.
+A large number of invalidations derive from the fact that StaticArrays defines a method for `!=`, which invalidates the fallback definition
+
+```
+!=(x, y) = !(x == y)
+```
+
+Since such definitions account for hundreds of nominal invalidations, it would be well worth considering whether it is possible to delete the custom `!=` methods.
```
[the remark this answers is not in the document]
mbauman: Fascinating!  And the reason that `==(::Any, ::Any)` doesn't have any dependencies is surely because its method table is so huge that doesn't do any call-site optimizations, yeah?

My initial thoughts were that defining `==` would just cascade its invalidations through `!=`... but that's not true because there's not a backedge for `(::Any, ::Any)` there!  And that's because the call-site optimization happened on `!=(::Any,::Any)` and the `==` is safely generically called within that method.

blog/2020/05/invalidations.md
mbauman: Interesting, I didn't know we sometimes specialized on `::Type{T}` without the explicit type-var.  That makes this make more sense to me now.

mbauman:
FWIW, I found it helpful for my own understanding to also look at the problem from "the other side" — that is, just looking at what backedges exist and watching how they're updated/changed.  Once I got this down, the invalidations became more obvious to me.  The key being that the backedges from _abstract_ specializations/entries are the ones that are going to be prone to invalidation by method addition:

<details>
```julia
julia> function backedges(f)
           mm = methods(f)
           mt = mm.mt
           println("### Method table backedges ###")
           mtbe = isdefined(mt, :backedges) ? mt.backedges : []
           for (tt, mi) in Iterators.partition(mtbe, 2)
               println(tt => mi)
           end
           println("### Method instance backedges ###")
           for m in mm.ms
               m.specializations === nothing && continue
               Base.visit(m.specializations) do mi
                   !isdefined(mi, :backedges) && return
                   for edge in mi.backedges
                       println(mi.specTypes => edge)
                   end
               end
           end
       end
backedges (generic function with 1 method)

julia> f(x::Int) = 1
f (generic function with 1 method)

julia> f(x::Bool) = 2
f (generic function with 2 methods)

julia> function applyf(container)
            x1 = f(container[1])
            x2 = f(container[2])
            return x1 + x2
        end
applyf (generic function with 1 method)

julia> backedges(f)
### Method table backedges ###
### Method instance backedges ###

julia> applyf([1,2])
2

julia> backedges(f)
### Method table backedges ###
### Method instance backedges ###
Tuple{typeof(f),Int64} => MethodInstance for applyf(::Array{Int64,1})

julia> applyf([false, true])
4

julia> backedges(f)
### Method table backedges ###
### Method instance backedges ###
Tuple{typeof(f),Bool} => MethodInstance for applyf(::Array{Bool,1})
Tuple{typeof(f),Int64} => MethodInstance for applyf(::Array{Int64,1})

julia> applyf(Any[1, false])
3

julia> backedges(f)
### Method table backedges ###
Tuple{typeof(f),Any} => MethodInstance for applyf(::Array{Any,1})
### Method instance backedges ###
Tuple{typeof(f),Bool} => MethodInstance for applyf(::Array{Bool,1})
Tuple{typeof(f),Bool} => MethodInstance for applyf(::Array{Any,1})
Tuple{typeof(f),Int64} => MethodInstance for applyf(::Array{Int64,1})
Tuple{typeof(f),Int64} => MethodInstance for applyf(::Array{Any,1})

julia> f(x) = 3
f (generic function with 3 methods)

julia> backedges(f)
### Method table backedges ###
### Method instance backedges ###
Tuple{typeof(f),Bool} => MethodInstance for applyf(::Array{Bool,1})
Tuple{typeof(f),Bool} => MethodInstance for applyf(::Array{Any,1})
Tuple{typeof(f),Int64} => MethodInstance for applyf(::Array{Int64,1})
Tuple{typeof(f),Int64} => MethodInstance for applyf(::Array{Any,1})

julia> applyf(Any[1, false])
3

julia> backedges(f)
### Method table backedges ###
### Method instance backedges ###
Tuple{typeof(f),Bool} => MethodInstance for applyf(::Array{Bool,1})
Tuple{typeof(f),Bool} => MethodInstance for applyf(::Array{Any,1})
Tuple{typeof(f),Int64} => MethodInstance for applyf(::Array{Int64,1})
Tuple{typeof(f),Int64} => MethodInstance for applyf(::Array{Any,1})
Tuple{typeof(f),Any} => MethodInstance for applyf(::Array{Any,1})
```
</details>

It's rather interesting how type-instabilities in core Julia code have this non-obvious secondary "cost."

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,733 @@
...
+
+function applyf(container)
+    x1 = f(container[1])
+    x2 = f(container[2])
+    return x1 + x2
+end
+```
+
+Here I've defined two functions; `f` is a function with two very simple methods,
+and `arrayf` is a function with just one method that supports `Any` argument at all.
```
ianfiske, line 43: Minor typo?

```suggestion
and `applyf` is a function with just one method that supports `Any` argument at all.
```

timholy:
@mbauman, I wondered about showing that directly, but it was already a long blog post and I didn't really feel like turning it into a tutorial on all of Julia's internal structures (since those are liable to change anyway). But perhaps it would make sense to advertise this via MethodAnalysis, since a level of indirection provides mechanisms for support as internals change.

> It's rather interesting how type-instabilities in core Julia code have this non-obvious secondary "cost."

I agree that type-instability is, in practice, a major contributor, but really this can happen any time a method has a non-concrete signature. Demo:

```julia
julia> using SnoopCompile

julia> f(::Integer) = 1
f (generic function with 1 method)

julia> function applyf(container)
           x1 = f(container[1])
           x2 = f(container[2])
           return x1 + x2
       end
applyf (generic function with 1 method)

julia> c = Any[1, true]
2-element Array{Any,1}:
    1
 true

julia> applyf(c)
2

julia> trees = invalidation_trees(@snoopr f(::Bool) = 2)
1-element Array{SnoopCompile.MethodInvalidations,1}:
 insert f(::Bool) in Main at REPL[6]:1 invalidated:
   mt_backedges: signature Tuple{typeof(f),Any} triggered MethodInstance for applyf(::Array{Any,1}) (0 children) more specific
   backedges: superseding f(::Integer) in Main at REPL[2]:1 with MethodInstance for f(::Integer) (1 children) more specific
   2 mt_cache

julia> applyf(c)
3
```

Obviously invalidation was necessary here. It's possible any time you call a method with a more specific set of types than its signature and then later provide a new method that gets called for those types.

But the general ethic against type-piracy means that in practice a lot of stuff won't invalidate. Hence type-instability is a major back door through which invalidation becomes more of a risk.

Interesting observation: above I got both an `mt_backedges` and a `backedges` invalidation. If instead I had made the new method `f(::Float64) = 2`, I only get the `mt_backedges` one. Makes total sense, but perhaps it's a clue to pay attention to when trying to understand partial-specialization invalidations.

mbauman:
> I agree that type-instability is, in practice, a major contributor, but really this can happen any time a method has a non-concrete signature.

I'd say that calling `applyf(Any[1, false])` is a type instability resulting from its abstractly-typed container argument.  Had you not _called_ `applyf` with a `Vector{Any}` we wouldn't have an invalidation!  Now, in normal code it's just fine — and necessary — for dynamic behaviors, but these problems are all stemming from core infrastructural code that got compiled into the sysimg.

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,733 @@
...
+```
+    M = M::Module
+```
+
+immediately after the `for` statement to fix the problem.
+Not only does this fix the invalidation, but it lets the compiler generate better code.
+
+The other case was a call from `Pkg` of `keys` on an AbstractDict of unknown type
+(due to partial specialization).
+Replacing `keys(dct)` with `Base.KeySet(dct)` (which is the default return value of `keys`) eliminated a very consequential invalidation, one that triggered seconds-long latencies in the next `Pkg` command after loading Revise.
```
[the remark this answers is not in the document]
timholy: We do get zero-cost abstraction when inference works. However, this seems to be a case where it doesn't. As of today I have better tools, so maybe I'll actually revert that change and fix it "properly." But hold on to your hat, you'll see this takes a lot more sophistication than the cheap version.

For reference, we're investigating https://github.com/JuliaLang/Pkg.jl/blob/fff698d3d7d90a55431ad2f4fa0af9719d899954/ext/TOML/src/print.jl#L53

```julia
julia> using MethodAnalysis

julia> using Pkg

julia> f = Pkg.TOML._print
_print (generic function with 2 methods)

julia> direct_backedges(f)
2-element Array{Any,1}:
                    MethodInstance for _print(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Dict{String,Array{Dict{String,Dates.DateTime},1}}) => MethodInstance for #print#10(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(Pkg.TOML.print), ::Base.GenericIOBuffer{Array{UInt8,1}}, ::Dict{String,Array{Dict{String,Dates.DateTime},1}})
 MethodInstance for _print(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Dict{String,Array{Dict{String,Dates.DateTime},1}}, ::Array{String,1}) => MethodInstance for _print(::Base.GenericIOBuffer{Array{UInt8,1}}, ::Dict{String,Array{Dict{String,Dates.DateTime},1}})
```

Hmm, that's curious, all the calls are concretely typed for the dict, which makes it a mystery why the `keys` call wasn't inferred. Since this is a keyword method, maybe we should check the kw function. Since it turns out we get a hit, let's be selective and filter for them:

```julia
julia> fkw = Core.kwfunc(f)
_print##kw (generic function with 2 methods)

julia> bes = filter(pr->!isconcretetype(pr.first.specTypes), direct_backedges(fkw))
2-element Array{Any,1}:
                    MethodInstance for (::Pkg.TOML.var"#_print##kw")(::NamedTuple{(:sorted,),Tuple{Bool}}, ::typeof(Pkg.TOML._print), ::IOStream, ::AbstractDict) => MethodInstance for #print#10(::Base.Iterators.Pairs{Symbol,Bool,Tuple{Symbol},NamedTuple{(:sorted,),Tuple{Bool}}}, ::typeof(Pkg.TOML.print), ::IOStream, ::AbstractDict)
 MethodInstance for (::Pkg.TOML.var"#_print##kw")(::NamedTuple{(:sorted,),Tuple{Bool}}, ::typeof(Pkg.TOML._print), ::IOStream, ::AbstractDict, ::Array{String,1}) => MethodInstance for (::Pkg.TOML.var"#_print##kw")(::NamedTuple{(:sorted,),Tuple{Bool}}, ::typeof(Pkg.TOML._print), ::IOStream, ::AbstractDict)

julia> using AbstractTrees

julia> print_tree(bes[1].first, 10)
MethodInstance for (::var"#_print##kw")(::NamedTuple{(:sorted,),Tuple{Bool}}, ::typeof(_print), ::IOStream, ::AbstractDict)
└─ MethodInstance for #print#10(::Pairs{Symbol,Bool,Tuple{Symbol},NamedTuple{(:sorted,),Tuple{Bool}}}, ::typeof(print), ::IOStream, ::AbstractDict)
   └─ MethodInstance for (::var"#print##kw")(::NamedTuple{(:sorted,),Tuple{Bool}}, ::typeof(print), ::IOStream, ::AbstractDict)
      ├─ MethodInstance for (::var"#28#32")(::IOStream)
      │  ├─ MethodInstance for #open#284(::Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(open), ::var"#28#32", ::String, ::Vararg{Any,N} where N)
      │  ├─ MethodInstance for #open#284(::Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(open), ::var"#28#32", ::String, ::Vararg{String,N} where N)
      │  │  └─ MethodInstance for open(::var"#28#32", ::String, ::String)
      │  │     ├─ MethodInstance for #bind_artifact!#24(::Linux, ::Nothing, ::Bool, ::Bool, ::typeof(bind_artifact!), ::String, ::String, ::SHA1)
 ⋮
```
So we've identified `bind_artifact!` as one likely culprit (because the dict appears after that).

Using Cthuhlu one can discover that the fallback definition https://github.com/JuliaLang/Pkg.jl/blob/fff698d3d7d90a55431ad2f4fa0af9719d899954/src/Artifacts.jl#L632 is one culprit, it should be `Dict{String,Any}()`.

However, that's not the only culprit: the other is https://github.com/JuliaLang/julia/issues/15276.

And, it involves not just `bind_artifact!` but also `load_telemetry_file` and `get_auth_header`. 

Especially since it involves 15276, I'm just not sure this is blog worthy. Thoughts?

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,733 @@
...
+
+```
+reduce_empty(::Base.BottomRF{typeof(max)}, ::Type{VersionNumber}) = VersionNumber(0)
+```
+
+In that case we could delete the `length` check and rely on this to produce the desired outcome.
+One should take this approach only if one can be convinced that version 0 can act as a universal "neutral element" for reductions involving `max` over `VersionNumber`s.
+Adding this to `Pkg` would be type-piracy, since `max`, `BottomRF`, and `VersionNumber` are all defined in `Base`, so fixing it this way would best be done in Base.
+
+But there are other possible fixes.
```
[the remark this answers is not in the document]
timholy: I don't think that's a bug. We have lots of methods that result in an ambiguity error if you use `Union{}` as a type. Try `zero`, `one`, etc.

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,733 @@
...
+A little higher up the tree we see a call to `mapreduce`, and this presents another opportunity because `mapreduce` allows you to supply an `init` value:
+
+```julia-repl
+julia> mapreduce(identity, max, Set(VersionNumber[]); init=VersionNumber(0))
```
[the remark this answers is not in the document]
timholy: Yes, I think it's best to discover from the tree.

blog/2020/05/invalidations.md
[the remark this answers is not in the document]
timholy: https://github.com/JuliaLang/Pkg.jl/pull/1824

blog/2020/05/invalidations.md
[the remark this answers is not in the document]
tkf: I respond it to https://github.com/JuliaLang/julia/pull/35843#issuecomment-628221479 as I think discussing it here would be distracting.

blog/2020/05/invalidations.md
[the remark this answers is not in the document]
tkf: Thanks for the extra in-depth explanation!

Would using `TOML.print(io, convert(Dict{String,Any}, dict), sorted=true)` everywhere in Pkg work too?  Since Pkg knows the concrete type with which it is working, I think it makes sense to type-stabilize at the API boundary.  (This way, `keys` in TOML is still generic.)

By the way, my point on "zero-cost abstraction" is not only about `KeySet` (but `KeySet` is a very good example). It's more about balancing abstraction and load-time optimization in general. Are we going to have to restrict abstraction for reducing load-time? Or would it turn out to be not much of a concern and even help us find/implement a better abstraction? `reduce_empty` is a bit hopeful example in this direction.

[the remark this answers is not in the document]
timholy: > Would using TOML.print(io, convert(Dict{String,Any}, dict), sorted=true) everywhere in Pkg work too? 

Yes, though that wasn't clear to me at the outset. But the PR I submitted is a better solution since it fixes the root cause.

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,734 @@
...
+end
+```
+
+This is [union-splitting], a performance optimization for cases where an object might be of one of several types.
+
+### Triggering method invalidation
+
+One point in particular is worth noting: what's up with that final `f(x)::Int` call?
+Didn't it already handle all the possibilities?
+Well, all the possibilities *so far*.
```
c42f, line 145: There's something odd about this: the world age in which `applyf` was compiled didn't have any other possibilities for `f(x)`. So isn't hitting this fallback always going to result in a world age error?

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,734 @@
...
+
+```julia-repl
+julia> mapreduce(identity, max, Set(VersionNumber[]); init=VersionNumber(0))
+v"0.0.0"
+```
+
+Perhaps we could just call this instead of `maximum`.
+However, it's a bit uglier than the original;
+perhaps a nicer approach would be to allow one to supply `init` as a keyword argument to `maximum` itself.
+While this is not supported on Julia versions up through 1.5, it's a feature that seems to make sense, and this analysis suggests that it might also allow developers to make code more robust against certain kinds of invalidation.
```
c42f, line 714: I admit to reading this (and various of the other suggestions for potentially removing invalidations) with a certain amount of dismay!

For a post on a general blog like the julialang blog, I feel like it sends an odd message to readers: "Be careful of abstractions you guys! Actually you might need to know about the implementation detail and call into that on occasion!"

Seasoned Julia developers will know that there's a maintenance and clarity cost and have an idea of when to make the tradeoff. But I fear that won't be true of everyone reading this post.

blog/2020/05/invalidations.md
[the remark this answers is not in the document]
tkf: I feel what is better is not entirely objective here (e.g., depending less on compiler implementation details may be preferable in some situations).

Anyway, as I said, `KeySet` is just one concrete example that load-time optimization can potentially be in conflict with the abstraction of the interface.  I don't think addressing one example solves the root problem.  Putting this as an open hard problem in the blog post seems to be a good balance to me. (Though I'm aware maybe I have an extreme point of view for this. Maybe the post as-is is fine.)

blog/2020/05/invalidations.md
tkf: Yeah, I had a similar comment https://github.com/JuliaLang/www.julialang.org/pull/794#discussion_r424454133

(Having said that, I think this particular part about `init` makes totally sense. `reduce` is not well-defined on possibly-empty collection if `op` is not a monoid; i.e., it needs `init` in general. So the invalidation hunt teaches us it's good to pay attention to algebra :smile:.)

timholy:
> I also feel rather uncomfortable that this might encourage people to remove abstractions

The right way to treat this is as a running diary of what choices we'd have to make to fix invalidations. Writing them down helps me think it through and encourages other people to do the same, and when we don't like what we see that means we need to come up with a better approach. This morning's `@invoke` was designed to alleviate the pain point you expressed above (I felt the same thing and am glad to see others do too), and Jeff even thinks we can do most of it automatically, so sit tight and wait for the unfolding saga...

I'll make sure I continue to try to document the painful parts of this just to provoke myself and others to complain. But unless it drags on (and it's not, it's going really fast right now), I'll wait to actually merge this until the most important fixes to Julia itself have been made and we can brag about how much better 1.6 will be.

odow:
I'm looking into this for JuMP, https://github.com/jump-dev/JuMP.jl/pull/2278, and I'd appreciate some advice.

I've got the number of invalidations down from 6630 to 1896, but it didn't have any noticeable impact on compile time. I'm also not really sure whether we should be fixing these at the JuMP level, or in the dependent packages with type instability: https://github.com/jump-dev/JuMP.jl/pull/2278#issuecomment-657628033.

This is part of an effort to speed up JuMP's tests, where compile time is 1000 times larger than runtime https://github.com/jump-dev/JuMP.jl/pull/2277.

cc @mlubin

timholy:
I'm finding that on `master` together with the master branch of Pkg, the number of invalidations triggered by JuMP is down to 2626. It looks to me like most of the remaining ones will be fixed by https://github.com/JuliaLang/julia/pull/35877. Before changing JuMP, I might suggest waiting until that merges and then checking again?

FYI, the impact of fixing invalidations is sometimes less direct than what you're measuring. Eliminating an invalidation eliminates the cost of *recompiling* the invalidated method instances. So, if your package invalidates `foo(::String, ::Int)` but your package doesn't use it, you'll see no direct benefit for yourself. What you're fixing is a problem for *other* users of `foo(::String, ::Int)` who would otherwise notice a mysterious slowdown of their user experience every time they load JuMP.

This effect cascades throughout the ecosystem: you load lots of lower-level packages, and if they are good citizens then *you* will receive benefit. So generally you're fixing things for the packages that come after you.

You can see this pretty easily in retrospect. On Julia 1.4, certain Pkg operations are slowed by more than 2s just by loading Revise. Loading Plots, plotting something, and then loading any of a host of other packages means that your next plot will be almost as slow as the first one ("time to second plot"). On Julia 1.6 most of that has been fixed, at least for a lot of the stuff in Base, Pkg, JuliaInterpreter, and Revise. Package *combinations* are probably the next frontier. But I expect https://github.com/JuliaLang/julia/pull/35877 to be very impactful, so I'd really recommend that anyone who isn't brave enough to cherry-pick that PR onto current master just avoiding doing much about this until that merges.

odow:
Thanks, this is very helpful.

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,734 @@
...
+julia> using SnoopCompile
+
+julia> invalidation_trees(@snoopr f(x::String) = 3)
+1-element Array{SnoopCompile.MethodInvalidations,1}:
+ insert f(x::String) in Main at REPL[7]:1 invalidated:
+   mt_backedges: signature Tuple{typeof(f),Any} triggered MethodInstance for applyf(::Array{Any,1}) (0 children) more specific
+```
+
+Let's walk through this output a bit.
+`@snoopr` turns on some debugging code inside Julia, and then executes the supplied statment;
```
mattBrzezinski, line 353:
```suggestion
`@snoopr` turns on some debugging code inside Julia, and then executes the supplied statement;
```

blog/2020/05/invalidations.md
```diff
@@ -0,0 +1,734 @@
...
+  7.238336 seconds (26.50 M allocations: 1.338 GiB, 7.88% gc time)
+```
+
+Because so much got invalidated by loading SIMD, Julia had to recompile many methods before it could once again produce a plot, so that in terms of time it was almost as expensive as the first usage.
+The size of the effect varies substantially depending on what task you are trying to achieve and what packages you load to do the invalidation.
+
+It's worth noting that you can escape much of this cost by loading all
+packages at the outset.
+Loading time is somewhat increased by invalidation, but the cost of
+first-time usage is typically larger. When possible, it's best to get
```
kdheepak, line 295: > Loading time is somewhat increased by invalidation, but the cost of
> first-time usage is typically larger.

Can you clarify this line? Do you mean "the cost of first-time usage _after invalidation_"?

timholy:
OK, after a long hiatus I came back and polished this up. I'm still waiting for a youtube video to convert (it's been stuck on "99% processed" for hours, not sure what's up with that...), so inserting the video link will have to wait until that's finished.

Can someone post a link to the preview, which I understand is generated for all PRs? I am not successfully figuring out where to find it.

@JeffBezanson and @vtjnash, I took the liberty of adding you as authors, since you've done a lot to reduce invalidations. Let me know if that's OK, and feel free to request changes in the content!

blog/2020/08/invalidations.md
```diff
@@ -0,0 +1,367 @@
...
+
+This prints out a list of method definitions (shown in light purple) that triggered invalidations (shown in yellow).
+In very rough terms, you can tell how "important" these are by the number of children;
+the ones with more children are listed last.
+You can gain more insight into exactly what happened, and what the consequences were,
+with `ascend`:
+
+![snoopr_simd_ascend](/assets/blog/2020-invalidations/SIMD_ascend.png)
+
+You're referred elsewhere for information about how to navigate this interactive menu,
```
odow, line 340: "referred elsewhere" isn't very helpful. How about

`The [SnoopCompile documentation](https://timholy.github.io/SnoopCompile.jl/stable/snoopr/) has information about how to...`

blog/2020/05/invalidations.md
timholy: Larger than loading. Invalidation can affect both, but changes to the first-time usage may have more effect.

blog/2020/05/invalidations.md
timholy: It does give you a good MethodError if you pass in a type that it doesn't work with.

timholy:
Barring concerns, I'll merge this Monday.

timholy:
I fixed the date and added a link to the video: https://www.youtube.com/watch?v=7VbXbI6OmYo. The main part is a tutorial in how to use SnoopCompile/Cthulhu to fix inference-related invalidations.

ianshmean:
Great vid! 

For the preview: https://julialang.netlify.app/previews/pr794/blog/2020/08/invalidations/

The trick is the `pr$(pr_num)` part

blog/2020/08/invalidations.md
```diff
@@ -0,0 +1,367 @@
...
+Recently, the [SnoopCompile] package gained the ability to analyze invalidations and help developers fix them.
+Because these tools will likely change over time, this blog post will only touch the surface; people who want to help fix invalidations are encouraged to read [SnoopCompile's documentation](https://timholy.github.io/SnoopCompile.jl/stable/snoopr/) for further detail.
+There is also a [video](https://www.youtube.com/watch?v=7VbXbI6OmYo) available with a live-session fixing a real-world invalidation, which might serve as a useful example.
+
+But to give you a taste of what this looks like, here are a couple of screenshots.
+These were taken in Julia's REPL, but you can also use these tools in [vscode] or other environments.
+
+First, let's look at a simple way (one that is not always precisely accurate, see SnoopCompile's documentation) to collect data on a package (in this case, loading the SIMD package):
+
+![snoopr_simd](/assets/blog/2020-invalidations/SIMD_invalidations.png)
```
johnnychen94, line 330: Not sure if this is my case only. The words in this screenshot are of smaller fonts than others.

image

timholy: Yeah, the output is fairly wide for the first one and it squashes to the width of the page. We can either choose to truncate and keep the fontsize similar or see the full display. Preference?

blog/2020/08/invalidations.md
```diff
@@ -0,0 +1,367 @@
...
+
+Unfortunately, method invalidation is (or rather, used to be) pretty common.
+First, let's get some baseline statistics.
+Using the [MethodAnalysis] package, you can find out that a fresh Julia session has almost 50,000 `MethodInstance`s tucked away in its cache.
+These are mostly for `Base` and the standard libraries.
+(There are some additional `MethodInstance`s that get created to load the MethodAnalysis package and do this analysis, but these are surely a very small fraction of the total.)
+
+Using [SnoopCompile], we can count the number of invalidations triggered by loading various packages into a fresh Julia session:
+
+| Package | Version | # invalidations (Julia 1.5) | # invalidations (Julia 1.6) |
```
ianshmean, line 215: Suggestion.. Perhaps it's better to reference this as `master (githash, 1.6-pre)`, given that 1.6 is fast evolving? I'm sure this post will be well-visited beyond the finalization of 1.6, so just saying 1.6 might confuse things a little?

blog/2020/08/invalidations.md
johnnychen94: I prefer to keep it as it is if reproducing a similar screenshot is non-trivial since you probably already fixed that :P

blog/2020/08/invalidations.md
timholy: That was a good suggestion. Enough of my PRs have been merged that I can now comfortably commit to a git-sha, so let's merge this sucker.

PR #795, opened 2020-05-11
Update the old pkg.julialang.org and juliaobserver.com URLs

ViralBShah:
cc @aviks @StefanKarpinski @djsegal

### DIFF
--- a/index.html
+++ b/index.html
@@ -414,7 +414,7 @@ end
       <div class="row">
         <div class="col-12" style="text-align: center">
           <a class="btn btn-sm btn-outline-danger" href="https://juliahub.com">JuliaHub: Package Docs and Code Search</a>
-          <a class="btn btn-sm btn-outline-danger" href="https://juliapackages.com/">Julia Observer</a>
+          <a class="btn btn-sm btn-outline-danger" href="https://juliaobserver.com/">Julia Observer</a>
         </div>
       </div>
 
--- a/index.html
+++ b/index.html
@@ -413,7 +413,7 @@ end
 
       <div class="row">
         <div class="col-12" style="text-align: center">
-          <a class="btn btn-sm btn-outline-danger" href="https://juliahub.com">JuliaHub: Package Docs and Code Search</a>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -413,8 +413,8 @@ <h2 class="lead secondary-heading">Packages</h2>
 
       <div class="row">
         <div class="col-12" style="text-align: center">
-          <a class="btn btn-sm btn-outline-danger" href="https://pkg.julialang.org">Package Docs</a>
-          <a class="btn btn-sm btn-outline-danger" href="https://juliaobserver.com/">Julia Observer</a>
+          <a class="btn btn-sm btn-outline-danger" href="https://juliahub.com">JuliaHub: Package Docs and Code Search</a>
+          <a class="btn btn-sm btn-outline-danger" href="https://juliapackages.com/">Julia Observer</a>
```
StefanKarpinski, line 417: @djsegal: Not sure what you want to call this link—calling it "Julia Packages" is kind of confusing.

ViralBShah: I see juliaobserver.com is still running. It would have been ideal to shut the old one down and rename the new website to JuliaObserver.com.

ViralBShah:
@logankilpatrick Is it possible to see the deployment preview for this PR? I feel the two buttons may be of different sizes and look strange.

logankilpatrick:
Https://julialang.netlify.app

ViralBShah:
That doesn't have this PR on it.

logankilpatrick:
Yeah, our current config does not handle multiple previews at once. So the only test would be locally or merge those other PR's.

index.html
djsegal: I'm ok just saying juliapackages.com is Julia Observer in these docs

What's preventing the migration is that JP doesn't have any cron jobs (whereas JO does)

ViralBShah: Should we then link to juliaobserver.com? If it doesn't have a cron job, won't it fall behind quickly?

djsegal: I'd probably agree with that. The post about shutting down JO was supposed to be an april fools day prank

// JP is still technically in beta till cron jobs get finished

ViralBShah: What? I thought that was real!

Issue #791, opened 2020-05-08
More/better splash images


cormullion:
It was only a matter of time before people (on Slack/Zulip) decided that the main splash image wasn't exciting enough (although calling my subtly-wrought sine curves in

<img width="251" alt="Screenshot 2020-05-08 at 10 40 40" src="https://user-images.githubusercontent.com/52289/81393496-5fc1cb00-9118-11ea-9ee2-0ceff7f71203.png"> 

"shitty" was a bit harsh, I think. :) 

It might be cool to have a set of alternating splash images. I've made a proof of concept, which you can find at:

https://julialangblogmirror.netlify.app

<img width="400" alt="Screenshot 2020-05-08 at 11 05 05 copy" src="https://user-images.githubusercontent.com/52289/81395699-0196e700-911c-11ea-8223-6a64fa002aa8.png">

The basic idea is that there is a bank of 10 (or however many) images, and a new image is chosen each day (in sequence). For testing purposes I changed "days" to "seconds", so for now you're effectively getting a new image each time you refresh the page. But probably getting a new image each day would be better (and faster).

For the moment, I've added 9 more of my own Julia-made images more or less at random, but obviously it would be better if images from other Julian graphics maestros (eg the Makie folks) were included. The main criteria are that the PNG file is about 2000w by 525h and generally of a darker/more saturated look, so that the foreground text stands out as it should.

I know little about web development, so I don't know how much of what I've done is good enough for the official site. JavaScript is odd. :)

tlienart:
That's awesome, and I guess we could just have this circulating automatically. 

One issue to bear in mind is that these images are pretty effing heavy & when getting to the front page on a poor connection leads to seeing a big white rectangle for ±0.5sec before getting the image. 

For instance this image: https://julialangblogmirror.netlify.app/assets/images/backsplash/backsplash-5.png is almost 1Mb

I'm pretty sure you generated them so you should be able to get them in SVG and get the weight to < 100kb right?

logankilpatrick:
Looks awesome @cormullion I never thought the sin curve was bad :)

cormullion:
I did run the images through a PNG-crusher, but I suppose the more detailed ones are bound to be larger. I think imposing a size limit is probably a good idea (although I normally hate that...:).

abhishalya:
This is awesome!

jessebett:
Sorry for the excessively harsh critique of your sine curves @cormullion  <3

[...]

PR #796, opened 2020-05-11
make font color of .main-heading black

kolokd:
white text on a light background makes the text unreadable. as the other text on the page has black font color on white background, making this also black increase consistency.

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -440,7 +440,6 @@ img {
 }
 
 .page-heading .main-heading {
-  color: #fff;
   font-weight: 100;
 }
 .page-heading .secondary-heading {


tlienart:
could you post a before/after screenshot?

kolokd:
oh, my bad. i have set ublock to Block media elements larger than 30 KB. When the page loads because of the background image is 55.3 KB it blocked form loading. so i didn't see the background.

kolokd:
i think the background image should be optimized.

tlienart:
yes definitely, see also #791 (I wholly agree that images should be small, ideally SVG < 100kb; your limit is probably a bit harsh in that respect).

kolokd:
yes, it is harsh. but i have to pay for every bytes.

Issue #789, opened 2020-05-08
Put RSS link only on some specific pages


tlienart:
On Slack @jessebett mentioned that it'd be better not to have the RSS button on every blog page as it is now. 

Assuming we'd only want it on the `/blog/` page, the change to bring in `_layout/head.html` is to change those lines: https://github.com/JuliaLang/www.julialang.org/blob/c891e474a2bc8a3c41d22104649f70ff65bad339/_layout/head.html#L55-L57 for

```html
{{ispage /blog/index.html}}
<a type="application/rss+xml" href="https://julialang.org/feed.xml">
  <i style="color:#e2802f" class="fa fa-rss-square"></i>
</a>
{end}}
```

Would be good to get some feedback first?

zdroid:
That's a good idea

ViralBShah:
I don't necessarily agree with that. I think the feed button should be on every blog page, but it could be put in a different place where it's a bit more subtle. 

However, if the choice is between having it the way it is right now and not having it at all on every blog post, I personally would prefer the former.

zdroid:
@ViralBShah How about this?
![Screenshot from 2020-05-13 10-55-24](https://user-images.githubusercontent.com/2725611/81792389-464dc400-9508-11ea-800e-9916325c4739.png)

![Screenshot from 2020-05-13 10-53-32](https://user-images.githubusercontent.com/2725611/81792193-01c22880-9508-11ea-8eb8-b5fd7973a9d9.png)
It's not distracting, but also allows you to subscribe on every page.

tlienart:
@zdroid I think that's better than what we have now, could you open a PR? there may still be discussion on whether we want it on all pages or not but in the meantime I reckon it's a good change

zdroid:
@jessebett What do you think about the new design? Is removing the RSS icon still necessary?

jessebett:
Hey, 

I don't think removing the RSS icon is necessary. My main gripe was how prominent it was and its juxtaposition with the title of the blog.

Having it inline but to the right may be a good solution, though I don't know how this will look on a narrow window, will it look the same as before?

ViralBShah:
Looks fine. You can try it out - it's live.

zdroid:
@jessebett Is this good to be closed now?

ViralBShah:
I'm going to close it. Can be reopened if necessary.

PR #801, opened 2020-05-13
Move RSS-icon to the right

zdroid:
Also adds a `.rss-icon` class since it's easier to keep track of styles that way.

Source: https://github.com/JuliaLang/www.julialang.org/issues/789#issuecomment-627848279


tlienart:
lgtm, thanks!

ViralBShah:
Nice!

Issue #800, opened 2020-05-13
Frontpage re-design


tlienart:
Following #791 and a discussion on Slack, here are some actionable things we could work on to make some pages more ergonomic and lighter to load.

### Front page specific

* [ ] put the tools (IDE/Debugger tools) on a separate page,  access it via a new button  next to "Try Julia in the Browser"
* [ ] at the moment the buttons "Code examples" and "Try julia in the browser" lead to  the same REPL monster that takes forever to load, not a good sign, only one of the two should, the other one should show actual code directly (like it did before with the rosetta code)
* [ ] remove the Environment stuff and their heavy gifs and move it to the Community page 
* [ ] update the splash image (see #791) making sure that background images there are <100kb. Ideally generate SVG files with programs like Luxor etc and cycle through them
* [ ] make the youtube video link small (_see Go front page_)
* [ ] make the "recent blog post" be automatically generated from the last 3-5 blog posts

### Community page

The community page is a bit of a mess, too much information and not structured  enough, here are some points:

* [ ] add a table of content to help navigate the page
* [ ] begin the page with clear links to "where can I find a package that does X" + key Julia org. 
* [ ] have some examples with key packages, how to do X with Y (ft packages that have > 500 stars)
* [ ] remove the list of meetups and only have a link to the meetup aggregator

Thoughts & actionable points are welcome. Please bear in mind that at some point someone will have to do it so if you want us to do large changes (e.g. change framework) please provide a full working POC otherwise we will likely not do it.

zdroid:
Why can't recent blog post be just... last blog post?

cormullion:
Although I'm a big fan personally, I'm starting to think that having a slideshow/art-gallery isn't entirely a good idea. It can become too much a matter of personal opinion and taste, and the people who comment may be outliers (be they minimalists or lovers of the baroque), some want a Julia set, others hate background graphics... 

The most economical SVGs I can make are linear gradients - you can easily get this beauty for 727 bytes:

<img width="576" alt="Screenshot 2020-05-13 at 09 52 05" src="https://user-images.githubusercontent.com/52289/81792071-7349a900-94ff-11ea-8979-ca7233fad53e.png">

but as soon as I try to add a handful of straight lines it creeps up to 30K or so. 

<img width="576" alt="Screenshot 2020-05-13 at 09 55 10" src="https://user-images.githubusercontent.com/52289/81792435-f834c280-94ff-11ea-94dd-47c6990a02b2.png">

[...]

PR #802, opened 2020-05-13
Drop JuliaCon container

zdroid:
I know that golang has video display on its frontpage, but it's not really that appealing. I doubt anyone comes to the frontpage to see the videos, and if you actually want to see them, you're better off opening them in the new tab. Also, embedding YouTube videos slows down the page load substantially.

Source: https://github.com/JuliaLang/www.julialang.org/issues/800#issuecomment-627886783


ViralBShah:
Why did we drop this? We get a huge amount of video views from this.

ViralBShah:
Let's revert this. For major changes it would be good to let the PRs be open some time to gather feedback.

tlienart:
sorry for the over-enthusiastic merge, I'll bring it back

zdroid:
@ViralBShah Right, I agree. I thought it's not really that prominent since it's not on the top of the page. I'll try to find another way for that section to be redesigned.

logankilpatrick:
~ 33% of all our youtube traffic comes from external sources and ~40% of those external sources are the website.

ViralBShah:
Thank you. There was a PR that tried to load the video async.

PR #803, opened 2020-05-13
Use full width for blog posts on mobile

zdroid:
Looks much better this way, especially on mobile devices. I'm wondering why such change was implemented in the first place.

### DIFF
--- a/_layout/head.html
+++ b/_layout/head.html
@@ -25,14 +25,12 @@
     .main pre {
   	  margin-left: auto;
   	  margin-right: auto;
-  	  width: 90%;
     }
-    .main { width: 75%; font-size: 100%; }
+    .main { width: 100%; font-size: 100%; }
     .main code { font-size: 90%; }
     .main pre code { font-size: 90%; }
-    .container.blog-title { width: 75%; }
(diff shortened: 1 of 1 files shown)


tlienart:
could you post a before/after screenshot?

zdroid:
Screenshot from 2020-05-13 15-30-38
Screenshot from 2020-05-13 15-30-19

tlienart:
IIRC @Keno had asked specifically that the blog posts be narrower so that they'd be easier to read, e.g. posts with a fair bit of text: https://julialang.org/blog/2020/05/rr/

zdroid:
Oh, makes sense. In that case, pre element should be contracted only.

ViralBShah:
Full page width is not easy to read...

ViralBShah:
Yes that would make sense. Full page width on mobile is actually what you want.

Keno:
That just seems like a problem with the zoom setting. Potentially caused by the new floating button.

zdroid:
@tlienart That page is broken on mobile nevertheless.

tlienart:
thanks that looks good :+1:

PR #804, opened 2020-05-13
reverses #802

tlienart:
I'll just merge this directly as it just reverts a recent change. We can discuss design further in #800


ViralBShah:
Thank you

PR #808, opened 2020-05-14
Align navbar items better

zdroid:
Before:
Screenshot from 2020-05-14 14-44-27

After:
Screenshot from 2020-05-14 14-56-30

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -17,12 +17,16 @@
 #main-menu li.nav-item.active {
   border-bottom: 2px solid #CA3C32;
 }
-#logo {
-  padding: 0 20px 0 0;
+.navbar-brand {
+  margin-left: 0.25rem;
+  margin-right: 0;
+  padding-bottom: 1rem;
 }
-#logo img {
-  width: 85px;
-  height: 55px;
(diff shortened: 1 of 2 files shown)


tlienart:
nice! how about padding a little bit left and right so  that julia logo / sponsor are not directly against the boundary?

zdroid:
The inner or the outer (navbar's) boundary?

zdroid:
Also the problem with using 1/9th size is that isn't scalable (in case more items are added). It also isn't guaranted to work due to different text lengths of nav items.

tlienart:
fair enough for the 1/9; looking at your second screenshot there's just more whitespace between the logo and download, it might look more balanced if some of that whitespace was left of the logo (resp. right for sponsor). 
I'm not desperate for that change if you don't think it makes things more balanced then ignore and we can just merge this

zdroid:
You were right, looks better now. I also styled the logo to style better vertically.
Screenshot from 2020-05-14 18-02-59

tlienart:
Awesome thanks, let's merge this it seems like a net improvement and we can always get back to it if needed

PR #809, opened 2020-05-14
Allow JuliaCon alert to be closed

zdroid:
Looks like this:

Screenshot from 2020-05-14 16-24-38

Once closed, the preference is remembered in localStorage.

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -512,6 +512,7 @@ blockquote p {
   padding-right: 0;
 }
 #top-alert .alert {
+  border-radius: 0;
   margin-bottom: 0.2rem;
 }
 #top-alert p {
--- a/index.html
+++ b/index.html
@@ -9,9 +9,13 @@
 
 <body>
   <div class="container-fluid" id="top-alert">
-    <div class="alert alert-dark mb-0" role="alert">
+    <div class="alert alert-dark alert-dismissible mb-0" role="alert">
       <p class="text-center">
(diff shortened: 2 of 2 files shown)


abhishalya:
One one side people want to completely remove js from site. :P

ViralBShah:
Sounds like a good idea.

zdroid:
@abhishalya This actually can be done with minimal JS (no jQuery, too). I used Bootstrap's in-built methods, but it's as easy as without them. Or alternatively, BS could be built to use `alert.js` only.

ViralBShah:
@tlienart Ok to merge?

tlienart:
Yep that’s great :)

ViralBShah:
When we change the alert text with a new message, will it show up again (with the option to cancel)?

zdroid:
@ViralBShah When a new alert is added, adding `localStorage.removeItem("closed");` or just changing the local storage key name (from `closed` to `closed1` for example) does the thing.

Although it'd be good to promote that way, it gets to annoying after some time. Imagine a new user rethinking whether they should try Julia or not, opening the front page several times to check what does Julia offer. And every time they do that, they'd see an annoying alert on top of the page, blocking the way down. Not really the best experience.

An alert that can be closed seems like a reasonable compromise.

ViralBShah:
Can we add some comments to the page, so that the person editing the alert knows about this? Otherwise, people will edit the alert text and have no idea how to address this.

I should point out that of all the complaints we receive for the website, this is not one. Anyways, let's see how it goes.

zdroid:
@ViralBShah That should be fixed in #820. :smile:

PR #810, opened 2020-05-14
Redesign page heading

zdroid:
There are some apparent problems with the front page's heading, the most visible thing on it. It gives the feel of some unmaintained, poorly-designed website. The problems are:

1. Lack of Julia's description
2. Poorly spaced buttons
3. Bad button styles (e.g. green button's border)

I recalled the phrase "We are greedy: we want more" and decided to redesign the page heading completely.

Before:
Screenshot from 2020-05-14 18-43-40
After:
Screenshot from 2020-05-14 19-03-59

Some of @cormullion's SVG banners also style very nicely in this layout. Examples:
Screenshot from 2020-05-14 19-16-14
Screenshot from 2020-05-14 19-16-01
Screenshot from 2020-05-14 19-15-39

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -400,10 +400,12 @@ h4.card-title {
 .btn-primary {
   background-color: #4D64AE;
 }
-.btn-success,
-.btn-primary {
-  border: 1px solid rgba(0,0,0, 0.25);
+.page-heading .btn-success,
+.page-heading .btn-primary {
+  border-color: transparent;
+  box-shadow: 0px 0px 0px 3px #0a0a0a;
 }
+
 
 .link {
   color: #4D64AE;
@@ -438,7 +440,7 @@ img.ecosystem-image {
 .page-heading {
   color: #fff;
   padding: 3rem;
-  background-image: url('../assets/infra/splash-img.png');
+  background-image: url('https://raw.githubusercontent.com/cormullion/graphics/ed1bbef52ae2081a5a192f23d35a0cb9b38b8d30/svg-banners/backsplash-3.svg');
   background-repeat: repeat-x;
   background-size: auto 100%;
 }
@@ -453,6 +455,11 @@ img.ecosystem-image {
   margin-bottom: 2rem;
 }
 
+.page-heading .main-heading,
+.page-heading .secondary-heading {
+  text-shadow: 3px 3px #0a0a0a;
+}
+
 .page-heading p {
   margin-bottom: 0;
 }
@@ -460,6 +467,7 @@ img.ecosystem-image {
 .page-heading span.btn {
   padding-left: 0;
   padding-right: 0;
+  vertical-align: middle;
 }
 
 img {
--- a/_css/app.css
+++ b/_css/app.css
@@ -64,12 +64,6 @@
   margin-bottom: 50px;
 }
 
-.page-heading a {
-  /* font-family: "Futura LT Medium"; */
-  font-family: "Roboto", sans-serif;
-  font-weight: 400;
-}
-
 .extra-link {
   font-size: 12px;
   text-decoration: none !important;
@@ -292,10 +286,6 @@
   font-family: "Roboto", sans-serif;
 }
 
-.btn-success {
-	background-color: #008a00;
-}
-
 .btn-outline-primary {
 	color: #0072ee;
 }
@@ -404,10 +394,17 @@ h4.card-title {
   }
 }
 
+.btn-success {
+  background-color: #008a00;
+}
 .btn-primary {
   background-color: #4D64AE;
-  border-color: #4D64AE;
 }
+.btn-success,
+.btn-primary {
+  border: 1px solid rgba(0,0,0, 0.25);
+}
+
 .link {
   color: #4D64AE;
 }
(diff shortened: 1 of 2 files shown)


abhishalya:
Some space between the two buttons? Also how about we add an icon to the download button.

zdroid:
Like this?
Screenshot from 2020-05-14 19-56-17

To implement the icon idea, it may be handy to remove the v1.4 from the button text.

ViralBShah:
I like this! Should we go with the triangles?

zdroid:
@ViralBShah The first of @cormullion's SVG banners that I posted? That one caught my eye too, it's quite attractive.

cormullion:
How does the asymmetry work on mobile or when resized?

ViralBShah:
Do remove the dot at the end of the second line.

zdroid:
@ViralBShah Are you sure about removing the dot? It feels somewhat more serious with a period, at least in my judgement. Go, Kotlin and Python also have a description as a full sentence, with a period.

Also, split buttons with a space in between or not?

@cormullion Pretty good, and definitely better than it is currently.
Screenshot from 2020-05-14 20-34-39

ViralBShah:
I feel it would be better without the dot. I like the space between the two buttons.

cormullion:
By the way - your "some unmaintained, poorly-designed website" comment is a bit unfair in my view; volunteers have worked quite hard on this site. It's obviously easier to tweak the details such as button spacing at a later date. My $0.02.

zdroid:
@cormullion It wasn't meant to belittle the work done on the website. The site obviously has a lot good things, but the first impression is the most important for newcoming users. No matter how much good content the site has, if the first impression is bad, it won't catch their attention. :cry:  That's why I proposed this redesign in the first place.

tlienart:
this is great :+1: we should get more people to chip in, could you post a screenshot in mobile mode?

zdroid:
@tlienart it's above: https://github.com/JuliaLang/www.julialang.org/pull/810#issuecomment-628815221

StefanKarpinski:
I really love this banner:

Screenshot from 2020-05-14 19-16-14

And the new design looks really nice!

zdroid:
Due to the increased interest for that one splash image, I put it on and tweaked the other elements to be styled better (text-shadow, box-shadow). Here it is:

Screenshot from 2020-05-15 00-52-17

The only problem on this page heading is this GitHub star button. It is impossible to style it properly without editing the button-loading script. We can:
1. either keep it like this, even though doesn't have proper box-shadow and vertical alignment (it's a bit skewed towards the top)
2. or replace it with our own custom button.

ViralBShah:
Can we try our own version of the star button?

zdroid:
@ViralBShah I think that will go a bit harder. Another option which I forgot is to move the star button somwhere else, like the Open source section right below.

ViralBShah:
BTW, I am ok with the button styling as is.

zdroid:
I pushed the new version now, in case anyone wants to try it out. It's in the `New background image` commit, so if it gets rejected I can easily go back to the previous commit.

I am personally also fine with the current GitHub button. Its styling may be left for some future PR / issue. If the others agree on this design, we can go with it.

ViralBShah:
Merging. Can always revert if we need to.

_css/app.css
```diff
@@ -438,26 +437,41 @@ img.ecosystem-image {
   margin-right: 20px;
 }
 
-.jumbotron {
-  background-image: url('../assets/infra/splash-img.png');
+.page-heading {
+  color: #fff;
+  padding: 3rem;
+  background-image: url('https://raw.githubusercontent.com/cormullion/graphics/ed1bbef52ae2081a5a192f23d35a0cb9b38b8d30/svg-banners/backsplash-3.svg');
```
cormullion, line 443: Perhaps this should be copied across - the repo here is only temporary...

abhishalya:
Also it seems on mobile the buttons are overlapping each other a bit.

_css/app.css
tlienart: I'm on it , PR incoming

tlienart:
> Also it seems on mobile the buttons are overlapping each other a bit.

@abhishalya might be worth opening another issue with screenshots ?

zdroid:
@tlienart I confirmed it, the fix is coming.

zdroid:
@tlienart @ViralBShah It'd be interesting to check Google Analytics after this was implemented, e.g. for retention rate.

fredrikekre:
The download button has a high saturation color when not hovered and a low saturation color when hovered. The documentation button is the other way around. Maybe high sat on hover for both? Also, the background is kinda dark, is it possible to lightening it up? Maybe remove the kinda thick black lines between the triangles would be one way to do that?

zdroid:
@ViralBShah (or someone else with the access to the Analytics) Has there been any retention rate change for the front page since the past month? I'm just curious if the redesign had any measurable impact.

ViralBShah:
How would one see that in analytics? The website traffic has all kinds of noise based on when articles show up on HN, releases are made, etc.

zdroid:
@ViralBShah Behavior -> Site Content -> All Pages (also accessible via "Page Report"). `/` indicates the index there, as on this picture:
Screenshot from 2020-06-12 01-16-06
Pageviews are noisy, but the other metrics shouldn't be. Avg. time on page is obviously expected to increase, while bounce and exit rates are expected to decrease (although that may be a bit noisy).

ViralBShah:
There's no noticeable change on any of these over the month before.

PR #812, opened 2020-05-14
Color navbar items on hover and select

zdroid:
Currently navbar items are very hard to notice. I set the color of items to match the bottom border's color on hover and select. The default color is inherited from body, i.e. it'll be the same as for the other text on the page.
Looks like this:
Screenshot from 2020-05-14 21-56-32
(Blog is hovered, Learning is active / selected)

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -2,20 +2,28 @@
   background-color: #fff !important;
   padding: 0;
 }
-#main-menu li.nav-item {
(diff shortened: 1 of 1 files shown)


tlienart:
Looks good to me! Thanks for all these improvements :-) let’s see if there’s any more feedback and happy to merge this tomorrow

PR #813, opened 2020-05-14
Move sitemap.xml to /

zdroid:
It is supposed to be in the root by default (unless it's specified differently in robots.txt). Even though https://julialang.org/sitemap.xml somehow exists, it is different from https://julialang.org/assets/sitemap.xml.


ViralBShah:
Sounds good!

PR #815, opened 2020-05-15
Making backsplash image a local asset

tlienart:
This 

* brings the backsplash image here instead of from @cormullion's repo.
* put that image through `svgo` (20% decrease)

No visible change so I'll just merge after checking the preview.


cormullion:
> svgo (20% decrease)

Impressive (and scary if you look at the source!). I was using http://petercollingridge.appspot.com/svg-optimiser but I'll use svgo next time...

PR #816, opened 2020-05-15
Fix mobile display of buttons

zdroid:
Fixes https://github.com/JuliaLang/www.julialang.org/pull/810#issuecomment-629085689.
Screenshot from 2020-05-15 10-24-31

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -461,6 +461,7 @@ img.ecosystem-image {
 }
 
 .page-heading p {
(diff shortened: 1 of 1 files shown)


tlienart:
lgtm, do you think this backsplash is a bit too dark maybe?

zdroid:
Maybe a bit. If the splash image is updated, then button and text styles should be updated too, to agree with the background visually.

PR #817, opened 2020-05-15
Restyle button colors

zdroid:
Before:
Screenshot from 2020-05-15 10-47-50
After:
Screenshot from 2020-05-15 10-47-26

This way the :hover color of buttons doesn't need to be changed, as they both are darker than the normal color. Also added text-shadow because the text on buttons seems a bit unreadable.

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -402,16 +402,11 @@ h4.card-title {
   }
 }
 
-.btn-success {
-  background-color: #008a00;
-}
-.btn-primary {
-  background-color: #4D64AE;
-}
 .page-heading .btn-success,
 .page-heading .btn-primary {
   border-color: transparent;
-  box-shadow: 0px 0px 0px 3px #0a0a0a;
+  box-shadow: 0px 0px 0px 2px #0a0a0a;
+  text-shadow: 1px 1px 2px #0a0a0a;
 }
 
 
@@ -465,7 +460,7 @@ img.ecosystem-image {
 
 .page-heading .main-heading,
 .page-heading .secondary-heading {
-  text-shadow: 3px 3px #0a0a0a;
+  text-shadow: 3px 3px 3px #0a0a0a;
 }
 
 .page-heading p {


zdroid:
Or just use the default Bootstrap colors:
Screenshot from 2020-05-15 11-00-57
(I tweaked text shadow a bit here)

tlienart:
I may be wrong but I think the colours for the button were picked out of the Julia Logo (@cormullion usually knows all these colour code by heart so probably best to ask him).

zdroid:
They were, but they don't agree with the background and hover colors. @cormullion maybe you can try to pair the colors better?

This is what I got after adding whitening the SVG a bit:
Screenshot from 2020-05-15 11-13-11

tlienart:
Maybe let's wait a bit for this one and ask for feedback on Slack or elsewhere?

cormullion:
Ha, these SVG banners were just example placeholders...! :) These aren't the Julia colors at full opacity, because they would be too bright as a background (I like dark backgrounds behind white text, for obvious (and accessibility) reasons).

Screenshot 2020-05-15 at 10 21 57

tlienart:
@zdroid with the lighter background, my opinion is that we can keep the colours of the button as they are with ok contrast, what do you think?

zdroid:
Looks better now (mostly due to the lines between triangles being thinner). The issue with the button colors on hover is still here, though.

Another issue is the contrast. Compare the previous and the current background:
Screenshot from 2020-05-18 11-18-47
Screenshot from 2020-05-18 11-19-34

I also tried out the PNG splash image from https://julialangblogmirror.netlify.app/ that I liked the most (and styled the elements accordingly):
Screenshot from 2020-05-18 11-36-07

It is not boring, not too colorful, not too bright and not too dark. Definitely something I'd look for. :laughing:

cormullion:
@zdroid Those are spiking neurons (https://www.youtube.com/watch?v=ECt6esNtfHE) - without the Julia colouring unfortunately...:)

zdroid:
Interesting. What do you think about using that kind of background, though?

cormullion:
I'm think I'm starting to prefer solid color backgrounds this last few days ...🤣😂😛

zdroid:
@cormullion I've had enough of colors. The site should be black on white and that's it! :rofl:

tlienart:
@zdroid should I just merge this? is this still appropriate? I'm sure it's good and we can revert if people don't like it

zdroid:
I would keep it open, as it isn't "final" yet (regarding the colors). The button colors should correspond to the background color, so if we change the background the button colors should be changed too.

ViralBShah:
I'd rather just merge and move on. Can always change again if we change background colors.

zdroid:
@ViralBShah Before that, I should change a thing or two. Will do.

zdroid:
It is ready now, looks like this:
Screenshot from 2020-05-19 11-56-49

tlienart:
thanks!

PR #819, opened 2020-05-17
Slightly modified backsplash

tlienart:
Courtesy of @cormullion 

**Current**
Screen Shot 2020-05-17 at 11 39 43

**Possible new**
Screen Shot 2020-05-17 at 11 39 49

**My opinion**

I like that the lines between triangles are thinner, it's also a bit less dark.

---

Could I have a 👎 to leave things as they are or  a 👍 to merge this one?


zdroid:
How about a more dark one instead? It gives better contrast to the text.

cormullion:
There are a few comments saying that the current one "is too dark" - perhaps your monitor is very bright? :)

zdroid:
It's actually pretty dark. I am just comparing it to the previous background which was darker and had a bit more contrast.

tlienart:
But, crucially, adjacent triangles are of different colors!!

Anyway it's the kind of stuff where people might never fully agree, would it make sense to run a short poll on discourse or something? (I fear the wave of random unrelated comments we might get)

abhishalya:
I prefer the old one because of the contrast. Maybe we can keep the darkness from old one and reduce the thinkness of lines as per the new one? Maybe we can reach a middle state in terms of darkness too? :P

ViralBShah:
No poll invitations for design stuff please. It's all too subjective. :-)

ViralBShah:
I did think the earlier one was too dark and prefer the lighter colors in the new one.

PR #821, opened 2020-05-17
Move tools into a separate page

zdroid:
As previously discussed. Although this looks like a good start, maybe this page could be repurposed into a "Getting started with Julia" page instead. It's something that the website is lacking, and appropriate editors / IDEs are good content for it.


ViralBShah:
It is a good idea. However these are not only for those getting started - many of the tools we want are for expert users as well. Perhaps we should sketch out what the end state of this looks like and get it done all together?

zdroid:
Your reasoning makes sense, as changing it to `Getting started` is not really a good idea. Is the title `Useful Tools for Julia` good to stay?

ViralBShah:
It would be nice to get to it from the front page. The other thing is - wouldn't this page appear too sparse content-wise?

zdroid:
Compare the page to e.g. https://www.rust-lang.org/tools. The idea is to keep it simple and accessible for the future users.

I added it to the navbar now, so accessing it from the front-page won't be a problem.
Screenshot from 2020-05-29 00-43-54

ViralBShah:
But this one feels incomplete - with no text or explanation. Let's start with this and develop it further. 

Does this also move the Essential Tools? We might as well move them both.

zdroid:
Yes, it moves Essential Tools as well. Let's keep this open for suggestions of the most appropriate text for the  page.

logankilpatrick:
I think something like:

- Editors
- Other tools: Debugger, profiler, etc.
- CI/CD: Travis, Conveyor, GitHub, etc.
- Online tools (maybe): Repl.it, other places you can program in Julia.

zdroid:
Any news on this?

logankilpatrick:
I have nothing to add on my end

zdroid:
I meant, is it acceptable in the form I proposed it? If not, I guess I should close this PR and leave it for some future proposal.

Regarding the @ViralBShah's and your suggestions, I can't implement that myself, but PR is open for edits.

ViralBShah:
I say we leave it open for someone else to work on it in the future. I don't think we should merge it as is. I might come around to working on it in the next few weeks.

zdroid:
I'm closing it for now, but if someone is interested to work on it, the source code is always available.

PR #823, opened 2020-05-17
Remove ecosystems.html

zdroid:
The webpage https://julialang.org/ecosystems/` is probably from the old website, but it isn't referenced in any other page (not even the sitemap). https://www.google.com/search?q=www.julialang.org%2Fecosystems also gives no results (except the www_old website, which is inaccessible). Currently it points to `community/`, but that doesn't make much sense, since the ecosystem is best described by the `Ecosystem` section in `index.html`.

If it doesn't generate any views (check in Google Analytics), then I suggest removing it.

### DIFF
--- a/ecosystems.html
+++ b/ecosystems.html
@@ -1,7 +1,0 @@
-<!-- REDIRECT -->
-<!DOCTYPE html>
-<html>
-  <head>
-    <meta http-equiv="refresh" content="0; url=/community/" />
-  </head>
-</html>


tlienart:
So a few months back when I migrated from Franklin to the current one, I was told to explicitly keep all pre-existing links, for instance the redirects `/ecosystems/` and `/download/` . There's no real value in removing it so why not just leave it there (even if no one ever uses it)

zdroid:
I was thinking if e.g. ecosystem things get moved from the front page, it would be useful to have that page for that purpose instead.

ViralBShah:
Yes, we don't want to break old URLs that people might be linking to.

ViralBShah:
It would be fine to replace this with a new ecosystems page, naturally as part of one PR.

PR #825, opened 2020-05-20
Improve and update Windows specific platform instructions

musm:
Note I accidentally committed this to master and subsequently  reverted the commit, apologies for that.

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -6,47 +6,52 @@ different location on the same computer, or even to a different computer.
 
 ## Windows
 
-Julia is available for Windows 7 and later, both 32 bit and 64 bit.
+Julia is available for Windows 7 and later for both 32 bit and 64 bit versions.
 
+*We highly recommend running Julia using a modern terminal, such as installing the [Windows Terminal from the Microsoft Store](https://aka.ms/terminal).* 
+
+
+### Installation Notes
 @@tight-list
-1.  Download the Windows julia.exe installer for your platform. 32-bit julia works on both x86 and x86\_64. 64-bit julia will only run on 64-bit Windows (x86\_64).
-2.  Run the downloaded program to extract Julia.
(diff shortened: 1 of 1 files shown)


musm:
closes https://github.com/JuliaLang/julia/issues/7267

tlienart:
Cool :+1:

PallHaraldsson:
>Note, the 32-bit Julia binaries work on both 32-bit and 64-bit Windows

Maybe a similar problem exists on Windows, as with with the julia snap (on Ubuntu), that I discovered is 32-bit. It does and should work, until using with 64-bit libraries, e.g. with PyCall.jl. 'm not using Windows so I can't test. I guess if you install 32-bit everything else it would work (also on Ubuntu, just not commonly done).

zdroid:
@PallHaraldsson Did you try just `sudo apt install julia`? It's maintained by Debian developers and it's up to date.

PallHaraldsson:
No, and not i isn't(?).

I did:

```
$ sudo snap install julia --classic
```

it works, except 32-bit, so works until it doesn't, but for me at least on Ubuntu 18.04.4 LTS:

```
$ sudo apt install julia
[..]
Package julia is not available, but is referred to by another package.
[..]
E: Package 'julia' has no installation candidate
```

I don't have a problem, I use the download page.

zdroid:
The package exists for 20.04, 19.10 and 16.04. Maybe that's why you can't find it? I use 20.04 and it works out of the box.

PallHaraldsson:
Like I said, I do no have a problem, and actually use Julia 1.6. Still good to know, and strange 18.04.x isn't supported. Maybe an oversight/bug since older 16.04 is? [Ironically I get "upgrade offer to 19.10, which will soon expire, so I wait, as I know offer to 20.04 will come soon, and I'm not in a hurry,]

https://askubuntu.com/questions/842042/problems-installing-julia-language

and link there to:

https://julialang.org/downloads/platform.html

we might want to resurrect that page, or redirect to some other.

PR #829, opened 2020-05-22
Removed the book "mastering and Julia 1.0"

bjrnfrdnnd:
Book seemingly does not exist

### DIFF
--- a/learning/index.md
+++ b/learning/index.md
@@ -46,7 +46,6 @@ The [Julia Language's YouTube channel](https://www.youtube.com/user/JuliaLanguag
 @@tight-list
 * Tom Kwong. [Hands-on Design Patterns and Best Practices with Julia](https://www.packtpub.com/application-development/hands-design-patterns-julia-10).  Packt Publishing, January 2020.
 * Tanmay Bakshi. [Tanmay Teaches Julia for Beginners: A Springboard to Machine Learning for All Ages](https://www.mheducation.co.uk/tanmay-teaches-julia-for-beginners-a-springboard-to-machine-learning-for-all-ages-9781260456639-emea). McGraw Hill, November 2019.
-* Malcolm Sherrington. [Mastering Julia 1.0](https://www.packtpub.com/application-development/mastering-julia-10). Packt Publishing, November 2019.
 * Antonello Lobianco. [Julia Quick Syntax Reference](https://www.apress.com/gp/book/9781484251898). Apress, November 2019.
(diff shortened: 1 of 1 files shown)


tlienart:
Thanks.

For anyone else stumbling upon this, there  is  https://www.packtpub.com/gb/application-development/mastering-julia  but it's indeed 2015, there does not seem to be a second edition.

ViralBShah:
I am pretty sure a "Mastering Julia 1.0" had been released. @aviks Do you know more? Just curious. Certainly can't see it on the website.

aviks:
No the second edition of this book has not been published, and likely wont be. We should remove all non-1.0 books from the list.

ViralBShah:
That's a good idea - we should remove the old books.

PallHaraldsson:
It seems premature, as I found it here with !.0" at the cover: https://www.goodreads.com/book/show/41838458-mastering-julia-1-0
> Kindle Edition, 463 pages
> Published April 9th 2019 by Packt Publishing

before was published in 2015.

tlienart:
If you find a website where you can buy it I’ll add it back :)

PallHaraldsson:
FYI. Former edition was paperback, and current seems only Kindle (and why was it revoked? it would it get a cover and (empty) goodreads page as meant to be published but never actually was?).

It's very strange, Amazon link at goodreads didn't work (assuming at some point it did), and while I found ASIN, and it's on this website: https://www.synccentric.com/features/upc-asin/?gclid=CjwKCAjwk6P2BRAIEiwAfVJ0rJEnBSbf4JYnSpUnsPew0I1M9Gt9fwInGMMR3EilZ6JdPGcb7EvQYBoCNWkQAvD_BwE

then going to Amazon doesn't work.

https://www.amazon.com/gp/seller/asin-upc-isbn-info.html/ref=sxts_snpl_1_0_dadbf03a-6de4-4724-8cdc-ef3370f16869?pd_rd_w=9gr1v&pf_rd_p=dadbf03a-6de4-4724-8cdc-ef3370f16869&pf_rd_r=X4TESKS87WYC97VR80PK&pd_rd_r=911e97ee-d1f8-4d8f-afa9-f1d0bf3cda42&pd_rd_wg=eqCaQ&qid=1590245182

PallHaraldsson:
It actually at least on German Amazon: https://www.amazon.de/Mastering-Julia-1-0-processing-problems/dp/1788298136

I didn't try to buy it, or see how far I would get. "Kaupt" button where are you...?

tlienart:
I think it would make more sense to just reach out to  the author and ask what's up  than doing archeological work here right? 

(nice find on the German Amazon, though it says: "Derzeit nicht verfügbar. (...)" i.e. not  available + unclear when it'll be)

ViralBShah:
@aviks knows the author, and his comments are based on that.

PR #830, opened 2020-05-24
Update benchmarks.md and fix broken links

haardikdharma10:
3 links in benchmarks.md file were linking to an outdated link. Maybe the reason is that, ' _assets' folder was named as 'assets' earlier. Made the necessary changes.


logankilpatrick:
Thanks!

tlienart:
aah crap no no, there should  not be direct links via `_assets`, everything is copied to  `/assets/`. I'll fix this.

tlienart:
e.g. https://github.com/JuliaLang/www.julialang.org/blob/gh-pages/assets/benchmarks/benchmarks.svg

PR #841, opened 2020-05-28
Restyle page-heading to use full width

zdroid:
I'm opening this just as a proposal which may be liked more than the current look. The current one feels more comfy, while the proposed one may catch more attention.

Before:
Screenshot from 2020-05-29 01-18-05
After:
Screenshot from 2020-05-29 01-17-54

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -446,6 +446,7 @@ img.ecosystem-image {
   background-image: url('../assets/infra/backsplash-min-0.5.svg');
   background-repeat: repeat-x;
   background-size: auto 100%;
+  border-radius: 0;
 }
 
 .page-heading .main-heading {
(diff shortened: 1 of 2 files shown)


ViralBShah:
I prefer the way it is right now - but perhaps because that is what I am used to. Let's see what others feel.

logankilpatrick:
I am not personally a fan but I agree it could be that I’m used to the other setup.

tlienart:
My impression is that that banner _already_ takes quite a lot / too much in terms of grabbing attention so I'm personally not  super keen in making this bigger 😕 .
Btw with a bit more time having passed I'm now also unsure of the colours of that banner (too grabby compared to before which was more sober) but that can be for another discussion.

zdroid:
Thanks for the opinions, closing this one.

PR #847, opened 2020-05-29
Reformat footer styles

zdroid:
A follow up to #845.

It cleans unused footer styles and avoids using the same IDs (`#fmenu` and `#fmhead`) for multiple elements. Also improves the centering:

Screenshot from 2020-05-29 14-10-01
->
Screenshot from 2020-05-29 14-10-22

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -302,37 +302,46 @@
 	color: #d91b35;
 }
 
-.footer-menu {
-  background-color: #394059;
-  color: #fff;
-  /* font-family: "Futura LT Medium"; */
-  font-family: "Roboto", sans-serif;
-}
-.footer-menu h6 {
-  /* font-family: "Futura LT Medium"; */
-  font-family: "Roboto", sans-serif;
-  font-weight: 500;
-  font-size: 14px;
-}
-.footer-menu ul.nav li a {
-  padding: 3px 0;
-  /* font-family: "Futura LT Medium"; */
-  font-family: "Roboto", sans-serif;
-  font-size: 13px;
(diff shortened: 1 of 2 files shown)


tlienart:
nice. Note that back when Logan was working on this we had an issue with where the CSS was located. I do not know what caused this but it seems to still be an issue. So here you've put the footer stuff in the `app.css` and I think it won't work on the main site (even if it works locally). This is why the  style was in the footer file. 

Anyway feel free to merge this but be aware that the footer style might not be applied properly.

zdroid:
@tlienart What didn't work in it exactly? Some styles didn't work for me initially as well, since they were superseded by the ones in `app.css`. I removed those extraneous styles and then it displayed fine.

tlienart:
does julialang.netlify.app have the style expected? in that case you may have fixed the issue

tlienart:
let's just see..

tlienart:
@zdroid unfortunately I was right: https://julialang.org/index.html would you mind opening a fix PR pasting the footer style in the  `_layout/foot_general.html` for now? 

(PS: I don't think it's a  simple problem to figure out, it may  be some clash on gh-pages, I'm not sure but for now let's have something that works, if you have an idea we can experiment elsewhere)

tlienart:
ok nevermind, I was too quick to criticise, it's actually  fine (thanks a lot!) I reacted before the changes had a chance to propagate

zdroid:
@tlienart The issue seems to be that `css/app.css` takes some extra time to be updated. But it looks right now. :smile:

PR #849, opened 2020-05-29
Fix the responsive problems of video frame

zdroid:
cc @abhishalya


zdroid:
@tlienart @ViralBShah This should be merged asap to fix the mobile display.

PR #862, opened 2020-06-13
Setup for the new community page

logankilpatrick:
WIP PR for the new community page.


community/index.html
```diff
@@ -0,0 +1,110 @@
...
+        </div>
+      </div>
+
+      <br>
+
+    </div>
+  </div>
+
+</body>
+</html>
```
zdroid, line 110:
```suggestion
~~~
```

community/index.html
```diff
@@ -0,0 +1,110 @@
+<!doctype html>
+<html lang="en">
+
+<body>
+
```
zdroid, line 5:
```suggestion
~~~
```

community/index.html
zdroid: To escape HTML inside Markdown.

logankilpatrick:
> This should be `index.md` instead of `index.html` to avoid code duplication.

I will switch this at the end before we merge this in.

logankilpatrick:
@ViralBShah thoughts: https://julialang.netlify.app/community/

ViralBShah:
Let's bring the diversity text, the survey results, and all the official channels with the exact wording. The other channels and meetups and all can go away from that page.

logankilpatrick:
It's worth noting that with the amount of HTML being embedded in this markdown file, it sort of defeats the purpose and actually makes it harder to refactor later on. We should likely switch this index to be HTML and not markdown.

zdroid:
@logankilpatrick If you do that, you will have to re-write the layout structure all by yourself. This gives the room to errors. On the other side, in Markdown you just need to add `~~~` twice. It's how Franklin works, but I'll see if it can be improved.

tlienart:
well there's a bunch of recurring environments which could just be wrapped in a command which would plug in the relevant html. The advantage of doing this kind of stuff is that you're left with a lot less HTML which can help newcomers edit the actual content.

logankilpatrick:
> @logankilpatrick If you do that, you will have to re-write the layout structure all by yourself. This gives the room to errors. On the other side, in Markdown you just need to add `~~~` twice. It's how Franklin works, but I'll see if it can be improved.

Right, but there's no syntax highlighting or proper formatting when you put the HTML into the "~~~" which itself can lead to errors. Either way, I am fine with this for now just wanted to bring up the point.

zdroid:
@logankilpatrick There is, at least in VSCode. I'll see how this can be improved in Franklin anyway.

fredrikekre:
May I suggest merging the  `Community Standards` and `Diversity and Inclusion` sections under the same heading? It looks a bit weird with two headings with just a sentence in each. Seems like all the text fits under a `Community Standards` heading. Maybe I also suggest that the `Official Channels` section moves to the top? That is what I would expect most people are looking for when going to the `Community` section.

zdroid:
Just to expand what @fredrikekre proposed, I'd propose the following order:
Official Channels
Julia User and Developer Survey (because it's currently ongoing)
Events
Community Standards
Localization

ViralBShah:
The survey will be up for only another week or two.

zdroid:
Then that section can be put on top or in alert... and removed or moved down afterwards. But it's still important currently, isn't it?

ViralBShah:
The survey is already in the alert.

logankilpatrick:
I personally like the standard and diversity at the top. I think it sets the right tone.

ViralBShah:
Yes, that does set the right tone, but we can combine the two as @fredrikekre suggested.

fredrikekre:
> I personally like the standard and diversity at the top. I think it sets the right tone.

I disagree, it feels like you are landing on a page of "rules". Most websites has terms of service and similar things in the footer. Looking at other programming language, Rust has it at the bottom of their Community page, and I suggest we do the same.

ViralBShah:
But these are not fineprint rules - that you just click accept on without reading. It is a coherent set of points that is fairly easy to read and understand, and that we are serious about. That is why it doesn't belong in terms of service - like in a footer.

Maybe Rust should put them on top too?

fredrikekre:
I know they are not, and I didn't say they were. I just expressed how it feels to me seeing that as the first thing on that page.

logankilpatrick:
What is the best way to combine the two together?

fredrikekre:
> What is the best way to combine the two together?

Just remove the "Diversity and Inclusion" header and let that sentence be part of the "Community Standards" was my suggestion.

ViralBShah:
Done.

ViralBShah:
There's "Talk to us" on the landing page and "Official channels" on the Community page. It's a bit redundant. Thoughts?

logankilpatrick:
I am not against it

inkydragon:
Maybe we should keep ”Julia GitHub Groups“ section?

I think it's useful for users to check whether a GitHub organization exists for a specific domain

https://github.com/JuliaLang/www.julialang.org/blob/8212c6f91e03f9e1a64a1f43334e4a62343db7d0/community/old_index.md#julia-github-groups

ViralBShah:
Yes good point. We should keep a list of active groups here. The long tail list could perhaps live on the other page.

PR #863, opened 2020-06-14
Restyle hackathon codeblock

zdroid:
It was an unreadable codeblock, now it looks like this:

Screenshot from 2020-06-15 01-30-05

Blockquote style change in `app.css` is harmless, since no file uses `blockquote` (check `grep -R blockquote`). Without it, there are no line breaks inside blockquote.


logankilpatrick:
Thanks!

PR #865, opened 2020-06-15
Drop platform.js

zdroid:
I grepped the repository for any actual use of `platform.js`'s interface i.e. `platform.`, `require(['platform']` and `function(platform)` (as from https://github.com/bestiejs/platform.js/) and it produced no results.

Even if the script really is needed for some page, it should be scoped appropriately and not loaded in every page's footer. It is unnecessary 12kB more to load.


ViralBShah:
Nice!

PR #867, opened 2020-06-15
[WIP] Re-design of the Downloads Page

logankilpatrick:
This will switch the overall page to be inline with the design we want across the website. 

See a preview at: https://julialang.netlify.app/downloads/


ViralBShah:
Look at 1.5-beta. We need to incoproate MUSL binaries on Linux too.

I also think we should not list FreeBSD under Linux. Maybe just make it its own box.

ViralBShah:
Will we then do this for all the 3 release types?

logankilpatrick:
That’s the plan, just wanted initial feedback before I did it all.

giordano:
A link to the build instructions for the "source code" link? https://github.com/JuliaLang/julia/tree/master/doc/build

zdroid:
It looks like whitespace is missing between the sections, as everything is so tightly packed. You can add it with `.mb-{1,2,3,4,5}` classes from Bootstrap (`.mb-3` should be okay).

giordano:
Does the Apple logo look a bit stretched only to me?

zdroid:
@giordano Linux, FreeBSD and Apple logos on the page are not 1:1 originally, so when they get resized to 1:1, they look bad.

ViralBShah:
Any thoughts on how to incorporate the LTS, stable, and pre-releases in this design? I don't think we want the OS logos to be replicated. So perhaps the logos stay on top, and then each OS gets three rows for each of the different releases. 

This would make it OS-centric rather than release-centric, which seems ok.

logankilpatrick:
Yeah, I am looking at it right now with logos for each section and it's a bit too much.

logankilpatrick:
Maybe we only need logos for the Stable release section as that is likely to be where most folks are going anyway.

logankilpatrick:
Not sure we want it this bear bones but I actually like this design: 
Screen Shot 2020-06-17 at 7 12 34 PM

downloads/index.md
```diff
@@ -161,56 +161,140 @@ Checksums for this release are available in both, [MD5](https://julialang-s3.jul
...
+  border: 2px solid #dddddd;
+  margin: 3px;
+  border-radius: 8px;
+}
+
+</style>
+
+<div class="grid-container">
+<!-- five columns -->
+	<div class="platform">
```
zdroid, line 205: Are tabs here intentional? Many HTML files in the website have this problem, of mixing tabs and spaces.

logankilpatrick: This code was graciously donated to me by @cormullion (not pointing fingers just highlighting that other people may use spaces while I use tabs most of the time) but we can clean all these up after the content and design are set.

ViralBShah: Yeah, we should indent it to be clean.

logankilpatrick:
In the interest of not getting bogged down on this, I am going to move ahead with the logo download section for the stable release and then leave the other 2 the way they are now. I will also clean up the bottom section of the page to make some of those parts live in a subpage.

downloads/index.md
cormullion: I plead guilty - my HTML formatting tool (BBEdit/TIdy) inserts tabs by default. 🤣

downloads/index.md
```diff
@@ -138,7 +138,7 @@ Checksums for this release are available in both [MD5](https://julialang-s3.juli
   <!-- source -->
 
   <div class="downloads">
-  <p><a href="https://github.com/JuliaLang/julia/tree/v1.3.1">GitHub</a> </p>
+  <p><a href="https://github.com/JuliaLang/julia/tree/v1.4.2">GitHub</a> </p>
```
tlienart, line 141: should use `{{stable_release}}` for maintainability

ViralBShah:
@logankilpatrick close or keep open?

PR #871, opened 2020-06-17
Add Zulip to Official Channels Page.

logankilpatrick:


ViralBShah:
Edited it a bit. Deleted IRC. Do we monitor Pinterest, Instagram and Facebook? I feel like LinkedIn and Twitter are the only ones - but that is perhaps because I don't see the other channels.

zdroid:
@ViralBShah Instagram not really, since the picture-only format isn't that appropriate for a programming language. (However it still can work, but I've seen no such activity). Facebook has a group with 3k members, that's pretty active.

PR #874, opened 2020-06-17
Remove unavailable video

giordano:
The [Intro to Julia Data Frames](https://www.youtube.com/watch?v=0-Xsts_s5W) is unavailable for me, is it only me? CC: @bkamins

### DIFF
--- a/learning/index.md
+++ b/learning/index.md
@@ -17,7 +17,6 @@ Jupyter notebooks are available for the following YouTube tutorials over at the 
 * [Intro to Julia (version 1.0)](https://www.youtube.com/watch?v=8h8rQyEpiZA&amp;t) , by Jane Herriman
 * [Intro to Julia for data science](https://youtu.be/SLE0vz85Rqo), by Huda Nassar
 * [Intro to the Queryverse, a Julia data science stack](https://www.youtube.com/watch?v=OFPNph-WxLM), by David Anthoff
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Link does not work for me either. However, there’s an updated tutorial on YouTube right now.

ViralBShah:
We probably should only advertise JuliaAcademy, and standardize around that.

logankilpatrick:
Yeah, I’ll fix that up when I re-do the learning page.

bkamins:
This is the best approach. I do not know what was the link (I do not recall providing it). Anyway it was for sure an outdated video. JuliaAcademy is I guess a good reference as it is curated (and I can update materials after each major release there). So probably it is best not to give links to YouTube at all but to pages that are guaranteed to have correct links?

ViralBShah:
Yes. People will still find the videos on youtube. But we should advertise the academy. Probably in the video descriptions on youtube, we can say they are part of a series on JuliaAcademy.

BTW, we should also include that information on DataFrames.jl (in the README and docs) as a pointer.

bkamins:
> BTW, we should also include that information on DataFrames.jl (in the README and docs) as a pointer.

@logankilpatrick - can you please make a PR there the way you feel appropriate?

logankilpatrick:
Yes!

logankilpatrick:
PR is open!

PR #877, opened 2020-06-18
[WIP] Re-designed Julia Learning Page

logankilpatrick:
This is an overhaul of the Julia Learning page. The wording on some of the sections is sparse so feel free to suggest things.

### DIFF
--- a/learning/books.md
+++ b/learning/books.md
@@ -1,0 +1,39 @@
+## Books
+
+@@tight-list
+* Zacharias Voulgaris. [Julia for Machine Learning](https://technicspub.com/julia/). Technics Publications, June 2020.
+* Tom Kwong. [Hands-on Design Patterns and Best Practices with Julia](https://www.packtpub.com/application-development/hands-design-patterns-julia-10).  Packt Publishing, January 2020.
+* Tanmay Bakshi. [Tanmay Teaches Julia for Beginners: A Springboard to Machine Learning for All Ages](https://www.mheducation.co.uk/tanmay-teaches-julia-for-beginners-a-springboard-to-machine-learning-for-all-ages-9781260456639-emea). McGraw Hill, November 2019.
+* Antonello Lobianco. [Julia Quick Syntax Reference](https://www.apress.com/gp/book/9781484251898). Apress, November 2019.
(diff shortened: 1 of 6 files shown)


logankilpatrick:
https://julialang.netlify.app/learning/

zdroid:
Those red sections look a bit messy with different top and bottom whitespace:

Screenshot from 2020-06-18 09-40-34

ViralBShah:
Apart from the whitespace issues, I have some content thoughts.

- [ ] I think we should show all the JuliaAcademy classes, and some of the books on this page with icons/pictures.
- [ ] For Youtube, we should mention all the other non-teaching materials - perhaps that is for the community page.
- [ ] For the universities teaching Julia, maybe we should add this image

ViralBShah:
Screen Shot 2020-06-18 at 9 48 07 AM

ViralBShah:
Screen Shot 2020-06-18 at 9 48 19 AM

ViralBShah:
I can probably update the books to add a couple of new ones. I am guessing if we are going to do this, I can share the powerpoint slides, so that we can use the individual images in that to put a set of tiles together.

logankilpatrick:
@ViralBShah I agree with all those suggestions. Adding some images will liven things up. Feel free to post the images here or send them to me on Slack and I'll get them in there unless the ones above are sufficient ^

zdroid:
@ViralBShah It does look very good, but it can't be expanded further, which is a bit of a liability. It also won't display nicely on mobiles, since the image has fixed size and cannot be broken into layout components.

ViralBShah:
I think with books, the images are easy to expand. With universities, not so much. I guess we could just show this list as a collection of some universities, followed by a larger list.

zdroid:
@ViralBShah That's right, I agree with you that not all universities should be presented on the prominent pages. Maybe a separate page for that would work well enough.

But I still don't see much merit of using an image. It is not a suitable form of communication, to merge multiple images into a larger one, when they can be broken down and made into a layout. Such image will be displayed very bad on smaller screens for example.

ViralBShah:
I did not suggest using the image... Just shared it as an example of what I have. I have shared the presentation with the images with Logan.

PR #883, opened 2020-06-23
legal/data: page about Pkg telemetry data

StefanKarpinski:


legal/data/index.md
```diff
@@ -0,0 +1,147 @@
...
+```text
+Julia-Pkg-Protocol: 1.0
+Julia-Version: 1.6.0-DEV.282
+Julia-System: x86_64-apple-darwin14-libgfortran5
+Julia-CI-Variables: APPVEYOR=n;CI=n;CIRCLECI=n;CONTINUOUS_INTEGRATION=n;GITHUB_ACTIONS=n;GITLAB_CI=n;JULIA_CI=n;TF_BUILD=n;TRAVIS=n
+Julia-HyperLogLog: 405,2
+Julia-Interactive: true
+```
+
+The `Julia-Client-UUID` and `Julia-Project-Hash` headers are omitted. If you want to only opt out of sending the project hash header, you can leave `client_uuid` as (or delete it entirely and a new one will be generated) and set `secret_salt = falsse`, which results in this telemetry info:
```
wildart, line 129: `secret_salt = false`

legal/data/index.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Data Analysis & Retention
+
+Since the first version of Julia which connects to package servers by default is Julia 1.5, which has not yet been released, we have not yet begun analyzing the data sent by clients. Once we start to do this, details of this analysis will be provided here, including open source code which can be inspected and critiqued. The fully anonymous aggregate output of this analysis will be made publicly available and. We will also establish a data retention policy: all individual requested data collected will be kept for no more than a fixed period of time for analysis and then will be securely deleted.
```
giordano, line 147: > will be made publicly available and.

Missing something after "and" or just an extra "and"?

legal/data/index.md
giordano, line 129:
```suggestion
The `Julia-Client-UUID` and `Julia-Project-Hash` headers are omitted. If you want to only opt out of sending the project hash header, you can leave `client_uuid` as (or delete it entirely and a new one will be generated) and set `secret_salt = false`, which results in this telemetry info:
```

legal/data/index.md
```diff
@@ -0,0 +1,147 @@
...
+GITLAB_CI=n;JULIA_CI=n;TF_BUILD=n;TRAVIS=n
+```
+This header gives an indicator for each of a set of environment variables that are commonly set on automated continuous integration (CI) systems, such as AppVeyor, Azure Pipelines, CircleCI, GitHub Actions, GitLab, and Travis CI. On user systems, theses will typically not be set. When analyzing requests to package servers, these indictors help to distinguish request from automated CI systems from real users. Automated CI systems can initiate a lot of requests and tend to be ephemeral, so they may have a different client UUID each time they run, which means that they can significantly pollute the statistics gleaned from package server request records. Due to the large volume of requests that can be initiated by CI systems, it is also often necessary to handle these systems specially to reduce service costs, e.g. by configuring CI processes to use a package server that is co-located with the CI service. Without the ability to identify such CI requests, it would be hard or impossible to identify and mitigate these costly automated requests to package servers.
```
simonbyrne, line 79: typo
```suggestion
This header gives an indicator for each of a set of environment variables that are commonly set on automated continuous integration (CI) systems, such as AppVeyor, Azure Pipelines, CircleCI, GitHub Actions, GitLab, and Travis CI. On user systems, these will typically not be set. When analyzing requests to package servers, these indictors help to distinguish request from automated CI systems from real users. Automated CI systems can initiate a lot of requests and tend to be ephemeral, so they may have a different client UUID each time they run, which means that they can significantly pollute the statistics gleaned from package server request records. Due to the large volume of requests that can be initiated by CI systems, it is also often necessary to handle these systems specially to reduce service costs, e.g. by configuring CI processes to use a package server that is co-located with the CI service. Without the ability to identify such CI requests, it would be hard or impossible to identify and mitigate these costly automated requests to package servers.
```

StefanKarpinski: GitHub isn't highlighting the typo for me and I can't spot it. I also made a bunch of edits so I may have fidex it already. Where was the typo?

StefanKarpinski:
There's probably more typos to fix and improvements to make, but this is a good start. Thanks for the feedback, all!

legal/data/index.md
simonbyrne: "theses"

PR #889, opened 2020-06-29
Adding Arch Linux

fusion809:
Hi,

I noticed that the "Platform Specific Instructions for Official Binaries" did not mention Arch Linux, even though Arch Linux has had a Julia package in its official repositories for quite some time now. So in this commit I am merely adding Arch Linux to this package with instructions on how to install its Julia package. 

Thanks for your time.

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -170,6 +170,16 @@ Recent Debian/ubuntu distributions include their own build of Julia, which can b
 sudo apt install julia
 ```
 
+## Arch Linux
+The Arch User Repository has [a package for Julia](https://aur.archlinux.org/packages/julia-bin) that is built from the official binaries of Julia. To install it run:
+
+```
+sudo pacman -S base-devel git
+git clone https://aur.archlinux.org/julia-bin.git
+cd julia-bin
+makepkg -si
+```
+
 ## FreeBSD Ports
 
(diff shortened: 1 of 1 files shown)


downloads/platform.md
```diff
@@ -170,6 +170,13 @@ Recent Debian/ubuntu distributions include their own build of Julia, which can b
 sudo apt install julia
 ```
 
+## Arch Linux
+Arch Linux has its [own build of Julia](https://www.archlinux.org/packages/community/x86_64/julia/) in its community repository. To install it run:
+
+```
+sudo pacman -S julia
```
giordano, line 177: Please, no: https://github.com/JuliaLang/www.julialang.org/pull/702#discussion_r402695166

fusion809: Odd, I've never had package-related issues on Arch. I've had plenty on NixOS, but never Arch. If I change it to installing `aur/julia-bin` will that be satisfactory?

ViralBShah: Does `aur/julia-bin` use the official pre-compiled binaries? If so, yes that would be acceptable to add.

giordano:
Yes, that's what I use

downloads/platform.md
fusion809: Sure does and I've now made the change.

ViralBShah:
@giordano Can we merge this?

giordano:
Yes, should be fine

PR #897, opened 2020-07-06
fix typos in "put-this-in-your-pipe.md"

rfourquet:


blog/2013/04/put-this-in-your-pipe.md
```diff
@@ -199,8 +199,8 @@ Hello
 Process(`echo Hello`, ProcessExited(0))
 ```
 
-The `"Hello\n"` after the `readall` command is a returned value, whereas the `Hello` after the `run` command is printed output.
-The `Process(`echo Hello`, ProcessExited(0))` is the value returned by `run`.
+The `"Hello"` after the `readchomp` command is a returned value, whereas the `Hello` after the `run` command is printed output.
+The ``Process(`echo Hello`, ProcessExited(0))`` is the value returned by `run`.
```
rfourquet, line 203: I'm not certain this is the right fix.

tlienart: should be good

PR #900, opened 2020-07-06
Over 3500 packages

PallHaraldsson:
I saw over 3500 stated at Juliacomputing.com so I guess ok here. Before they didn't want to include with JLL packages, and without there are 3440, and with 3883 including the 16 waiting in the pipeline to be registered.

I find it likely "Julia has been downloaded over 13 million times" to be outdated. No July newsletter yet, this was in older.

### DIFF
--- a/index.html
+++ b/index.html
@@ -403,7 +403,7 @@ end
 
       <br>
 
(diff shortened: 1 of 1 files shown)


zdroid:
By the next month that'll be over 4000.

PR #901, opened 2020-07-10
Add portable builds for Windows

musm:
Fixes https://github.com/JuliaLang/julia/issues/32158
Note only applies to newer builds since it depends on https://github.com/JuliaLang/julia/pull/35098

I'm not sure how much it matters but these are tarballs, which on newer Windows versions `tar` is shipped and is in the system path. 

Perhaps longer term we can instead offer zip builds?

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -154,9 +154,14 @@ Checksums for this release are available in both, [MD5](https://julialang-s3.jul
 <table class="downloads table table-hover  table-bordered">
   <tbody>
     <tr>
-      <th> Windows Self-Extracting Archive (.exe) <a href="/downloads/platform/#windows">[help]</a></th>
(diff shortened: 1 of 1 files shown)


ViralBShah:
Note that post 1.5 release, the downloads template will need to be updated.

ViralBShah:
Now that I see it, I wonder if we could have just had 64-bit (Installer) and 64-bit (Portable) in the Windows row, just like we have multiple options for Linux.

musm:
Maybe. It might also be a bit more confusing on which one to use/click ?

We should probably put a recommended text on the installer

musm:
You mean something like this:

image

musm:
image

ViralBShah:
Yes, that's what I was imagining. I committed something.

PR #905, opened 2020-07-12
Footer: Add a link to `status.julialang.org`

DilumAluthge:


logankilpatrick:
The status page is awesome!

DilumAluthge:
Yeah @fredrikekre and @staticfloat did good work.

staticfloat:
It's all @fredrikekre, I just set things up once he'd already gone through the work of figuring out what we want.

DilumAluthge:
Anyway, anyone object to adding this link to the footer?

I put it under "About", but we can put it under a different heading if people prefer.

ViralBShah:
How about in Downloads? Maybe call it `PkgServer Status` or `Pkg status` is that is too long?

DilumAluthge:
Sure! Take a look.

PR #906, opened 2020-07-12
A better way of handling fork PRs vs non-fork PRs

DilumAluthge:
Currently, we just use `continue-on-error: true` statements to prevent the build for erroring on PRs made from forks.

This is a rather rough approach. We can be more precise.

The best way to do this is to explicitly check whether or not the PR is made from a fork. We use this conditional:
```
github.repository == github.event.pull_request.head.repo.full_name
```

If the PR is made from a fork, then this conditional evaluates to `false`. If the PR is NOT made from a fork, then this conditional evaluates to `true`.

Therefore, we only push PR previews if the PR is not made from the fork.

Everything else is the same for both fork and non-fork PRs.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -15,7 +15,7 @@ jobs:
         with:
           persist-credentials: false
       - name: Install SSH Client
-        continue-on-error: true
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Closing in favor of #907

Issue #903, opened 2020-07-10
PowerPC is incorrectly formatted in the Tiers on downloads page


ViralBShah:
PowerPC should like the x86-64 musl entry above. It is showing up in the first two columns instead of the last two. It is unclear why this is happening. The page source looks fine to me.

![Screen Shot 2020-07-10 at 5 43 29 PM](https://user-images.githubusercontent.com/744411/87205209-ea6ba580-c2d4-11ea-87b6-a2c3394d2343.png)

ViralBShah:
@tlienart Might you know why this is happening?

tlienart:
I’ll have a look today, thanks for the ping, maybe @zdroid has an idea too

PR #911, opened 2020-07-15
Downloads: PowerPC entry (erroneous HTML)

tlienart:
ref: #903 

it was an issue with `rowspan` for Linux being set to `7` and now that there's an extra line (powerpc) it should be `8`.

Screenshot 2020-07-15 at 11 42 53


ViralBShah:
Thank you! I did not realize the number of rows had to be captured right.

PR #916, opened 2020-07-19
Update index.md with new Julia book published

ordovician:
Adding a line for Julia book I published on Leanpub this year.


ViralBShah:
I thought this book is not yet complete. Perhaps we should indicate that it is 40% complete.

Or have a separate section on Books under preparation.

ordovician:
Yes one could do that. Although I don't expect it to remain at 40% very long. The book is basically done. I am just reading over chapters and extra time before adding them and allowing some time in between if case I get feedback.

ViralBShah:
Please resubmit when ready.

Issue #590, opened 2020-01-16
Add upgrade Julia version instructions


logankilpatrick:
Per here: https://discourse.julialang.org/t/a-proper-introduction-on-how-to-update-julia-somewhere-accessible/28166

We should add some sort of “official” upgrading Julia version instructions.

ViralBShah:
Generally, while you may want to upgrade - as in replace a patch release, you don't want to do that for minor and major releases. For those, you may want multiple versions of Julia around and we prepare the directory names and install paths carefully for that to work.

logankilpatrick:
I am not sure I follow. What does "you don't want to do that for minor and major releases"  mean? I don't want to update my Julia version?

ViralBShah:
A Julia 1.2 install should not be replaced with 1.3, for example but 1.2 can be replaced with 1.2.1.

logankilpatrick:
What do you mean by "should not be replaced"? Are you saying the user should keep 1.2 and 1.3 installed?. I have like 8 versions of Julia on my computer but I still want to keep updating to the different versions. 

I want to get something on this front going since it's a common question and the process is not easy.

ViralBShah:
Let's continue this in #926

PR #926, opened 2020-07-28
advertise jill.py in Downloads as a command-line installation tool

johnnychen94:
[jill.py](https://github.com/johnnychen94/jill.py) is a command-line tool (written in Python) that downloads and extracts official binaries, and makes symlinks for all platforms. This could serve as a recommended alternatives to package managers such as `apt`, `brew` and `choco`

There are many sub-commands, but the most useful one is `install`:

`jill install [VERSION] [--install_dir INSTALL_DIR] [--symlink_dir SYMLINK_DIR] [--upstream UPSTREAM] [--confirm]`

demo:

Notes:

* `jill` host a [list of binary mirrors](https://github.com/johnnychen94/jill.py/blob/master/jill/config/sources.json), so it doesn't always download Julia from official buckets. There's a gpg verification after downloading so I think it's safe to do so. To ensure downloading from official buckets, `jill install --upstream Official` can be used.
* This package is mostly developed and maintained by myself, I'm not very sure if this is appropriate to advertise this unofficial tool. I could move this project to JuliaLang and make modifications if it is desired.

cc: @staticfloat

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -56,6 +56,9 @@ Windows 7 / Windows Server 2012 users also need to install:
 
 Uninstallation is preferably performed by using the Windows uninstaller. The directory in `%HOME%/.julia` can then be deleted if you want to remove all traces of Julia (this includes user installed packages).
 
+## Cross-platform installer
+
+[Jill.py](https://github.com/johnnychen94/jill.py) is a community-maintained command-line tool that automates the installation workflow for all platforms. After installing this using `pip install jill -U`, you can then use `jill install` to install the current stable release, and `jill install latest` to install the nightly builds.
 
 ## macOS
 
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -56,9 +56,6 @@ Windows 7 / Windows Server 2012 users also need to install:
 
 Uninstallation is preferably performed by using the Windows uninstaller. The directory in `%HOME%/.julia` can then be deleted if you want to remove all traces of Julia (this includes user installed packages).
 
-## Cross-platform installer
-
-[Jill.py](https://github.com/johnnychen94/jill.py) is a community-maintained command-line tool that automates the installation workflow for all platforms. After installing this using `pip install jill -U`, you can then use `jill install` to install the current stable release, and `jill install latest` to install the nightly builds.
 
 ## macOS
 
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Does anyone have thoughts on this? @staticfloat @ViralBShah

fredrikekre:
I commented here: https://github.com/JuliaLang/www.julialang.org/pull/928#issuecomment-667078709

logankilpatrick:
Thanks @fredrikekre ! I did not realize that PR had the bulk of the discussion on this.

johnnychen94:
If there are concerns that [/downloads](https://www.julialang.org/downloads/) is too prominent, we could move it to [/downloads/platform](https://www.julialang.org/downloads/platform/). If there are concerns about its being non-official, I'm happy to move it to Julialang and let people here watch and maintain.

If this is about Python dependency, I don't think it's hard nowadays to get a Python executable compared to brew, choco. You only need to install `jill.py` once and it fits all Julia releases, including the future releases if there is no breaking change in how we install Julia. One feature of this is that it could be [used together with `cron`](https://github.com/johnnychen94/jill.py#example-with-cron) to get it updated for both stable and nightly builds.

But to clarify, I don't mind at all whether this tool gets advertised here, it's just that I regularly see terminal users complain about the missing installation tool and I thought perhaps I should advertise it so that more people would know about its existence and take a try. Yes, there are many of them available on each platform, just that there isn't a unified and recommended one yet.

ViralBShah:
We should start with `/downloads/platform` for now, and if it starts getting widespread adoption and maintenance, can always become more mainstream.

johnnychen94:
I'm not sure what's the opinion on this PR, I've moved the words to `/downloads/platform` as suggested by @ViralBShah

johnnychen94:
> https://github.com/JuliaLang/www.julialang.org/issues/590
> Generally, while you may want to upgrade - as in replace a patch release, you don't want to do that for minor and major releases.

`jill.py` basically follows this idea. Assume there's `v1.0.3` and `v1.4.0` installed

* `jill install 1.5` will update symlinks `julia` and `julia-1` to julia 1.5 exec, and create a symlink `julia-1.5`
* `jill install 1.5 --upgrade` will also `cp -r $JULIA_DEPOT_PATH/environments/v1.4 $JULIA_DEPOT_PATH/environments/v1.5`
* `jill install 1.3` will only create one symlink `julia-1.3` because it's older than the current julia version
* `jill install 1.4` will create and update symlinks `julia`, `julia-1` and `julia-1.4`

`jill install` gets you the latest stable version, `jill install 1` gets you the latest `1.y.z` version, and `jill install 1.5` gets you the latest `1.5.z` version. 

RC and beta is not included, to install these test versions, one has to specify the full version, e.g., `jill install 1.5.0-rc1`

ViralBShah:
Quick question. I am a bit confused since I have read different versions of this. Does it have to be a python script, or is there a shell script version as well?

I don't have a problem with it being python. Just making sure I understand.

johnnychen94:
[jill.sh](https://github.com/abelsiqueira/jill) written by @abelsiqueira is originally a Linux bash script that installs Julia, then I added macOS support. I want to add Windows support and other fancy features, then I made a [jill.py](https://github.com/johnnychen94/jill.py) fork which covers all jill.sh functionality.

ViralBShah:
I am personally ok with merging this. Does jill.py work on windows? Let's move it to the section after Linux and FreeBSD.

johnnychen94:
> Does jill.py work on windows?

Yes, it does.

PR #927, opened 2020-07-28
Post links for the zip instead of tarballs for windows

musm:
We did some work to convert these to zips on the buildbots since tarballs are not common for Windows users


ViralBShah:
Nice!

musm:
I think we should add the extension to links so people know ones an exe and the other a zip ?

ViralBShah:
Should be ok to do that.

PR #928, opened 2020-07-28
Simplify Linux installation to copy-paste

ChrisRackauckas:


barche:
The downside is that this will have to be updated every time the Julia recommended version changes.

downloads/platform.md
```diff
@@ -76,9 +76,16 @@ You can uninstall Julia by deleting Julia.app and the packages directory in `~/.
...
 ## Linux and FreeBSD
 
-It is strongly recommended that the official generic binaries from the downloads page be used to install Julia on Linux and FreeBSD.
+It is strongly recommended that the official generic binaries from the downloads page be used to install Julia on Linux and FreeBSD. The following
+command downloads the latest Julia version and makes a symlink to allow the `julia` command to find the binary:
 
+```
+wget https://julialang-s3.julialang.org/bin/linux/x64/1.4/julia-1.4.2-linux-x86_64.tar.gz
+tar -xvzf julia-1.4.2-linux-x86\_64.tar.gz
+ln -s ./julia-1.4.2/bin/julia ~/.local/bin/
```
rfourquet, line 85: This results in a broken symlink, maybe
```suggestion
ln -s $PWD/julia-1.4.2/bin/julia ~/.local/bin/
```

maxmouchet:
`$PWD` should be quoted, otherwise this fail if there is a space in the path:
```bash
"$PWD"/julia-1.4.2/bin/julia
```

downloads/platform.md
```diff
@@ -76,9 +76,16 @@ You can uninstall Julia by deleting Julia.app and the packages directory in `~/.
...
 ## Linux and FreeBSD
 
-It is strongly recommended that the official generic binaries from the downloads page be used to install Julia on Linux and FreeBSD.
+It is strongly recommended that the official generic binaries from the downloads page be used to install Julia on Linux and FreeBSD. The following
+command downloads the latest Julia version and makes a symlink to allow the `julia` command to find the binary:
 
+```
+wget https://julialang-s3.julialang.org/bin/linux/x64/1.4/julia-1.4.2-linux-x86_64.tar.gz
+tar -xvzf julia-1.4.2-linux-x86\_64.tar.gz
+ln -s "$PWD"/julia-1.4.2/bin/julia ~/.local/bin/
```
KristofferC, line 85: Is this in `PATH` by default?

StefanKarpinski: It's kind of weird that these instructions are just run in some unspecified directory. Presumably the user's home directory? Maybe a little info about where to run this?

ChrisRackauckas: is there a place we should download it to?

StefanKarpinski: Not really, but maybe a word about "pick a place where you want the Julia directory to live—this can be your home directory or maybe a `~/dev` directory if you prefer that.

ChrisRackauckas: When you `apt-get` or install a Windows binary, it just goes somewhere and that's where it is. What's a good default for us to do here? Anyone who cares where it goes will know what to do: the instructions are for the other audience.

ViralBShah:
@tlienart We have a way to make this generic, IIRC. So that it doesn't need changing with every release.

downloads/platform.md
ViralBShah: Why not say that run the command from the directory you want it in?

downloads/platform.md
```diff
@@ -76,9 +76,16 @@ You can uninstall Julia by deleting Julia.app and the packages directory in `~/.
 
 ## Linux and FreeBSD
 
-It is strongly recommended that the official generic binaries from the downloads page be used to install Julia on Linux and FreeBSD.
+It is strongly recommended that the official generic binaries from the downloads page be used to install Julia on Linux and FreeBSD. The following
+command downloads the latest Julia version and makes a symlink to allow the `julia` command to find the binary:
```
ararslan, line 80: The given command doesn't actually download the latest Julia version (it happens to download the most recent 1.4 release currently, but will soon be out of date). Might be worth emphasizing that it's an example for a particular version.

tlienart:
@ViralBShah im lacking context here but if you open an issue explaining what needs to be done & assign it to me I'll get it done asap

ViralBShah:
Sorry - if you look at the PR, it has hard-coded Julia versions, which will have to be changed every time there's a new release. I thought onthe downloads page we have variables that we can use to refer to the latest stable release.

logankilpatrick:
Yes we do, though not sure if they will work in this context.

tlienart:
Ah right in the code you mean? Yes yes that can be done; ok I'll try that today

johnnychen94:
Except for the python dependency, I still believe #926  `jill install [VERSION] [--install_dir INSTALL_DIR] [--symlink_dir SYMLINK_DIR]` provides better and generic instruction to this. And `jill` works for Windows, too.

downloads/platform.md
zdroid, line 85:
```suggestion
wget https://julialang-s3.julialang.org/bin/linux/x64/1.4/julia-{{ stable_release }}-linux-x86_64.tar.gz
tar -xvzf julia-{{ stable_release }}-linux-x86\_64.tar.gz
ln -s "$PWD"/julia-{{ stable_release }}/bin/julia ~/.local/bin/
```

zdroid:
@ViralBShah Look at [`config.md`](https://github.com/JuliaLang/www.julialang.org/blob/master/config.md). `stable_release` is defined there. I suggested the appropriate changes to fix that.

downloads/platform.md
tlienart: I think let's leave that to another PR as the comment above wrt Jill is more important

tlienart:
In light of  https://github.com/JuliaLang/www.julialang.org/pull/928#issuecomment-665530657 should we close this PR and open another one with the one-liner from `jill`?
 
PS: this is maybe an unrelated concern but would it be a security concern that the "blessed" download way would be dependent upon an external repo not part of the `JuliaLang` org?

fredrikekre:
> PS: this is maybe an unrelated concern but would it be a security concern that the "blessed" download way would be dependent upon an external repo not part of the JuliaLang org?

Yes, I don't think we should list that in such a prominent place. In particular it doesn't look good that you first have to install python and then a python package to finally install Julia.

Maybe we could instead support and maintain something like Rust's rustup.sh.

abelsiqueira:
> Yes, I don't think we should list that in such a prominent place. In particular it doesn't look good that you first have to install python and then a python package to finally install Julia.
> 
> Maybe we could instead support and maintain something like Rust's rustup.sh.

Just pointing out that Jill is a bash script, so there is no need to install python before: https://github.com/abelsiqueira/jill (maybe I should officially name it Jill.sh?)

If you prefer I can transfer ownership to JuliaLang.

johnnychen94:
> https://github.com/JuliaLang/www.julialang.org/pull/926#issue-457522244
> I could move this project to JuliaLang and make modifications if it is desired.

Install Julia is super easy. `jill.*` doesn't need to be the first choices for people to install Julia, they're just so handy when terminal users want a simple and reliable copy&paste solution. `jill.*` are made for these users.

For `jill.py`, there’s a very good chance Python is already installed for terminal users. How could `pip install jill` be problematic while `curl <args> <rustup-url> | bash` isn't? Perhaps I just need to rename it as `juliaup`?

> Maybe we could instead support and maintain something like Rust's rustup.sh.

Maybe, if there's someone in Julialang willing to and he happens to have some time.

ViralBShah:
I would be happy to move `jill` to JuliaLang. If we are to include it here though, it will need to be attended to, which shouldn't be too difficult. Good to know what @StefanKarpinski  @staticfloat @fredrikekre @KristofferC think.

ViralBShah:
Not sure what to do here.

StefanKarpinski:
I'd be happy to look at jill and decide if we should include it in a more official capacity, but blocking better Linux install instructions on that seems kind of unnecessary. How about we improve the Linux install instructions and then separately consider something like that?

ViralBShah:
I think the easiest way to get this merged is to ask people to download and untar in their home directory and then run with `./julia-<release>/julia`. This is literally how I run it on every single machine. Later, `jill` can be used to put all the versions in a single directory and make it easy to pick one  and so on.

abelsiqueira:
I agree with @StefanKarpinski. I started a discussion on https://github.com/abelsiqueira/jill/issues/33.

ViralBShah:
I am merging it right now as it is a significant improvement. Can always make it better.

downloads/platform.md
```diff
@@ -76,9 +76,16 @@ You can uninstall Julia by deleting Julia.app and the packages directory in `~/.
...
 
+```
+wget https://julialang-s3.julialang.org/bin/linux/x64/1.4/julia-{{ stable_release }}-linux-x86_64.tar.gz
```
johnnychen94, line 83: `1.4` here should be replaced with some placeholder

fredrikekre: `{{ stable_release_short }}`

KristofferC:
It seems strange to merge a "copy paste" solution that doesn't work and has open unaddressed review comments.

PR #931, opened 2020-07-29
RFC: Put the musl/glibc remark on the download page

thofma:
I love the remark about musl/glibc on the downloads/platform page, but I think it is a bit too hidden. This PR tries to put it directly on the downloads page. I tried a good old footenote, but I don' t know if I am doing this right or if it is the right approach.

I just wanted to get this moving, since I don't think a user should be required to know what glibc or musl is in order to download julialang. If a user has to leave the downloads page in order to google for more information for his system, the installation is too complicated :)

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -215,7 +215,7 @@ Checksums for this release are available in both, [MD5](https://julialang-s3.jul
 
 {{end}} <!-- upcoming_release -->
 
-(*) Most users would prefer the `glibc` version of the distribution unless you know that your system uses `musl`.
+(*) Most Linux users should use the glibc binaries unless you know that your system uses musl as its libc.
 
 ## Older Releases
 
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -172,7 +172,7 @@ Checksums for this release are available in both, [MD5](https://julialang-s3.jul
     <tr>
       <th> Generic Linux on x86 <a href="/downloads/platform/#linux_and_freebsd">[help]</a></th>
       <td colspan="3">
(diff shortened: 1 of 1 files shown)


downloads/index.md
```diff
@@ -215,6 +215,7 @@ Checksums for this release are available in both, [MD5](https://julialang-s3.jul
 
 {{end}} <!-- upcoming_release -->
 
+(*) Most users would prefer the `glibc` version of the distribution unless you know that your system uses `musl`.
```
ararslan, line 218:
```suggestion
(*) Most Linux users should use the glibc binaries unless you know that your system uses musl as its libc.
```

downloads/index.md
```diff
@@ -172,7 +172,7 @@ Checksums for this release are available in both, [MD5](https://julialang-s3.jul
     <tr>
       <th> Generic Linux on x86 <a href="/downloads/platform/#linux_and_freebsd">[help]</a></th>
       <td colspan="3">
-        <a href="https://julialang-s3.julialang.org/bin/linux/x64/{{upcoming_release_short}}/julia-{{upcoming_release}}-linux-x86_64.tar.gz">64-bit (glibc)</a>
+        <a href="https://julialang-s3.julialang.org/bin/linux/x64/{{upcoming_release_short}}/julia-{{upcoming_release}}-linux-x86_64.tar.gz">64-bit (glibc*)</a>
```
ararslan, line 175: I would put the asterisk on musl rather than glibc, since that's the one with the caveat.

ararslan:
Requesting a review from @tlienart to ensure that the syntax for footnotes works properly here and there isn't something better we should be doing. Thanks for this!

tlienart:
Maybe slightly prettier:

Screenshot 2020-07-30 at 11 53 02

also the initial `[1]` is a separate link so that when you click on it you jump to the footnote.

thofma:
Thanks, this looks awesome!

PR #936, opened 2020-08-01
Make the URL for upcoming release a bit nicer.

fredrikekre:


fredrikekre:
https://www.julialang.org/downloads/#upcoming_release_vupcoming_release_upcoming_release_date to https://www.julialang.org/downloads/#upcoming_release

PR #937, opened 2020-08-02
Set stable release to 1.5

ararslan:
Not ready just yet.


ViralBShah:
We have to move the current download block for 1.4 to old releases. Unfortunately we don't have automation there.

ararslan:
Good call, thanks Viral! Should be all good now.

PR #941, opened 2020-08-03
Fix stable release download block

### LINKED ISSUE #939: Add back links to portable builds on Windows
I don't see which PR removed them, but we should add them back.

### LINKED ISSUE #940: Revert or fix one of the recent PR to Downloads
It seems that a  bunch of things disappeared from the downloads page in a recent PR:

* windows portable links -- PR #901 but then #939 
* musm/glibc linux footnote -- PR #931 but now I can just see the footnote but not the reference to  the footnote (nor the download link with musm/glibc differentiation) 

was that on  purpose? 

cc @musm @ViralBShah @logankilpatrick

fredrikekre:
Fix stable release download block to include musl + portable win install, fixes #939, fixes #940.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -38,8 +38,8 @@ Checksums for this release are available in both [MD5](https://julialang-s3.juli
   <tbody>
     <tr>
       <th> Windows <a href="/downloads/platform/#windows">[help]</a></th>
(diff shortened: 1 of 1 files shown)


tlienart:
Thanks!

PR #942, opened 2020-08-03
Make HTML redirect a helper function

zdroid:
All links are identical except the one in `manual.html`. It used to be https://docs.julialang.org/en/v1/, but I changed it to https://docs.julialang.org for automatic redirect. That will make the link work for v2 too, I guess?

Idea from [sciml.ai](https://github.com/SciML/sciml.ai/blob/master/utils.jl#L68).


ViralBShah:
See if it all works when deployed?

PR #944, opened 2020-08-03
Center GitHub star button vertically

zdroid:
Before:
Annotation 2020-08-03 201545

After:
Annotation 2020-08-03 201544

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -463,6 +463,7 @@ img.ecosystem-image {
 }
 
 .page-heading span.btn {
+  padding-top: 18px;
   padding-left: 0;
   padding-right: 0;
   vertical-align: middle;


logankilpatrick:
I'm having a really hard time seeing the difference. Can you point out what has changed in the images?

zdroid:
@logankilpatrick You must be blessed not to have OCD like me... :laughing:  It's the star button's position in the right, it was moved ~5-10px lower.

giordano:
> I'm having a really hard time seeing the difference.

Open the two images in two tabs and switch between them.

> Can you point out what has changed in the images?

I think the GitHub icon is now centred compared to the "Download" and "Documentation" buttons

ViralBShah:
I like it better this way too.

PR #950, opened 2020-08-07
Properly scale images on Learning page

zdroid:
Using "center" class conflicts with defined width, while fixed width and height are bad for mobile devices. Both can be replaced with "img-fluid" class which scales the images appropriately.


logankilpatrick:
Thanks!

PR #953, opened 2020-08-10
Switch Slack links to not use HTTPS and use http instead.

logankilpatrick:


KristofferC:
I think it would be good if PRs like this also state the reason for the change.

ViralBShah:
This is now just a redirect url, which doesn't support https. We are using a different invite flow than before that is completely on slack.

Issue #956, opened 2020-08-15
{{stable_release}} substitutes with patch number but patch number is not in path on MacOS


pkalikman:
As a result this instruction

`ln -s /Applications/Julia-1.5.0.app/Contents/Resources/julia/bin/julia /usr/local/bin/julia`

on 

https://julialang.org/downloads/platform/#macos

is incorrect--the right path is `.../Julia-1.5.app/...`, at least on my MacOS Catalina machine. 

Apologies that I don't know how to format that token or I'd submit the PR to fix myself :)

fredrikekre:
You can probably just change `{{ stable_release }}` to `{{ stable_release_short }}` in https://github.com/JuliaLang/www.julialang.org/blob/master/downloads/platform.md.

pkalikman:
Thanks! Will try this.

PR #957, opened 2020-08-15
Fix path reference issue on MacOS

pkalikman:
stable_release substitutes with patch # so this instruction

`ln -s /Applications/Julia-1.5.0.app/Contents/Resources/julia/bin/julia /usr/local/bin/julia`

on https://julialang.org/downloads/platform/#macos is incorrect--the right path is `.../Julia-1.5.app/...`, at least on my MacOS Catalina machine.

Kudos @fredrikekre for solution.

https://github.com/JuliaLang/www.julialang.org/issues/956

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -67,10 +67,10 @@ Navigate to `/usr/local/bin` and remove the `julia` file. Then type the followin
 
 
 ```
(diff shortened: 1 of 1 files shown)


ViralBShah:
Thanks for the PR

PR #961, opened 2020-08-21
Update oldest supported version of macOS to 10.9

vtjnash:
In preparation for build system simplifications in https://github.com/JuliaLang/julia/pull/37121, since we aren't really testing back this far anyways


ViralBShah:
I am considering removing the stuff about 10.6. 

Also maybe this should just be a draft PR until 1.6 is released since the current release is not affected. Probably should be reverted.

PR #966, opened 2020-08-23
Update Julia in a Nutshell

ViralBShah:
Reworking of "Julia in a Nutshell" points. Multiple Dispatch is more prominent.

### DIFF
--- a/index.html
+++ b/index.html
@@ -79,22 +79,22 @@
       <!-- 3 -->
       <div class="col-lg-4 col-md-6 feature">
         <h3>Multiple Dispatch</h3>
-        <p>Julia uses <a href="https://docs.julialang.org/en/v1/manual/methods/">multiple dispatch</a> as a paradigm, making it easy to express many object-oriented and <a href="https://docs.julialang.org/en/v1/manual/metaprogramming/">functional</a> programming patterns. The talk on the [Unreasonable Effectiveness of Multiple Dispatch](https://www.youtube.com/watch?v=kc9HwsxE1OY) explains why it works so well.
+        <p>Julia uses <a href="https://docs.julialang.org/en/v1/manual/methods/">multiple dispatch</a> as a paradigm, making it easy to express many object-oriented and <a href="https://docs.julialang.org/en/v1/manual/metaprogramming/">functional</a> programming patterns. The talk on the <a href="https://www.youtube.com/watch?v=kc9HwsxE1OY">Unreasonable Effectiveness of Multiple Dispatch</a> explains why it works so well.
       </div>
       <!-- 4 -->
       <div class="col-lg-4 col-md-6 feature">
         <h3>General</h3>
-        Julia provides <a href="https://docs.julialang.org/en/v1/manual/networking-and-streams/">asynchronous I/O</a>, <a href="https://github.com/JuliaDebug/Debugger.jl">debugging</a>, <a href="https://docs.julialang.org/en/v1/stdlib/Logging/">logging</a>, <a href="https://docs.julialang.org/en/v1/manual/profile/">profiling</a>, a <a href="https://docs.julialang.org/en/v1/stdlib/Pkg/index.html">package manager</a>, and more. One can build entire [Applications and Microservices](https://www.youtube.com/watch?v=uLhXgt_gKJc) in Julia.</p>
+        Julia provides <a href="https://docs.julialang.org/en/v1/manual/networking-and-streams/">asynchronous I/O</a>, <a href="https://github.com/JuliaDebug/Debugger.jl">debugging</a>, <a href="https://docs.julialang.org/en/v1/stdlib/Logging/">logging</a>, <a href="https://docs.julialang.org/en/v1/manual/profile/">profiling</a>, a <a href="https://docs.julialang.org/en/v1/stdlib/Pkg/index.html">package manager</a>, and more. One can build entire <a href="https://www.youtube.com/watch?v=uLhXgt_gKJc">Applications and Microservices</a> in Julia.</p>
       </div>
       <!-- 5 -->
       <div class="col-lg-4 col-md-6 feature">
         <h3>Easy to use</h3>
-        <p>Julia has high-level syntax, making it an accessible language for programmers from any background or experience level. <a href="https://julialang.org/learning/code-examples/">Try some Julia code right in the browser</a> to get a feel for the language.</p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -66,35 +66,35 @@ <h2 class="lead secondary-heading">
...
-        <h3>General</h3>
-        <p>Julia uses <a href="https://docs.julialang.org/en/v1/manual/methods/">multiple dispatch</a> as a paradigm, making it easy to express many object-oriented and <a href="https://docs.julialang.org/en/v1/manual/metaprogramming/">functional</a> programming patterns.
-        It provides <a href="https://docs.julialang.org/en/v1/manual/networking-and-streams/">asynchronous I/O</a>, <a href="https://github.com/JuliaDebug/Debugger.jl">debugging</a>, <a href="https://docs.julialang.org/en/v1/stdlib/Logging/">logging</a>, <a href="https://docs.julialang.org/en/v1/manual/profile/">profiling</a>, a <a href="https://docs.julialang.org/en/v1/stdlib/Pkg/index.html">package manager</a>, and more.</p>
+        <h3>Multiple Dispatch</h3>
+        <p>Julia uses <a href="https://docs.julialang.org/en/v1/manual/methods/">multiple dispatch</a> as a paradigm, making it easy to express many object-oriented and <a href="https://docs.julialang.org/en/v1/manual/metaprogramming/">functional</a> programming patterns. The talk on the <a href="https://www.youtube.com/watch?v=kc9HwsxE1OY">Unreasonable Effectiveness of Multiple Dispatch</a> explains why it works so well.
```
JeffBezanson, line 87: This should use the "functions" chapter as the link for "functional" instead of metaprogramming. Macros are not really related to functional programming.

index.html
```diff
@@ -66,35 +66,35 @@ <h2 class="lead secondary-heading">
...
       </div>
       <!-- 5 -->
       <div class="col-lg-4 col-md-6 feature">
-        <h3>Easy to use</h3>
-        <p>Julia has high-level syntax, making it an accessible language for programmers from any background or experience level. <a href="https://julialang.org/learning/code-examples/">Try some Julia code right in the browser</a> to get a feel for the language.</p>
+        <h3>General</h3>
+        Julia provides <a href="https://docs.julialang.org/en/v1/manual/networking-and-streams/">asynchronous I/O</a>, <a href="https://github.com/JuliaDebug/Debugger.jl">debugging</a>, <a href="https://docs.julialang.org/en/v1/stdlib/Logging/">logging</a>, <a href="https://docs.julialang.org/en/v1/manual/profile/">profiling</a>, a <a href="https://docs.julialang.org/en/v1/stdlib/Pkg/index.html">package manager</a>, and more. One can build entire <a href="https://www.youtube.com/watch?v=uLhXgt_gKJc">Applications and Microservices</a> in Julia.</p>
```
JeffBezanson, line 92: Could mention macros here with the link to "metaprogramming" if you want.

index.html
```diff
@@ -66,35 +66,35 @@ <h2 class="lead secondary-heading">
...
-        <h3>Julia is fast!</h3>
+        <h3>Fast</h3>
         <p>Julia was designed from the beginning for <a href="/benchmarks/">high performance</a>.
         Julia programs compile to efficient native code for <a href="/downloads/#support_tiers">multiple platforms</a> via LLVM.</p>
       </div>
       <!-- 2 -->
       <div class="col-lg-4 col-md-6 feature">
         <h3>Dynamic</h3>
-        <p>Julia is <a href="https://docs.julialang.org/en/v1/manual/types/">dynamically typed</a>, feels like a scripting language, and has good support for <a href="https://docs.julialang.org/en/v1/stdlib/REPL/">interactive</a> use.</p>
+        <p>Julia is <a href="https://docs.julialang.org/en/v1/manual/types/">dynamically typed</a>, feels like a scripting language, and has good support for <a href="https://docs.julialang.org/en/v1/stdlib/REPL/">interactive</a> use. Julia has a <a href="https://docs.julialang.org/en/v1/manual/types/">rich language of descriptive datatypes</a>, and type declarations can be used to clarify and solidify programs.</p>
```
JeffBezanson, line 76: It doesn't really make sense to mention type declarations under "dynamic". I guess it could go under "fast" or "general".

JeffBezanson: Possible solution: rename section to "Dynamic and optionally typed"?

index.html
JeffBezanson: To make the section name an adjective like the others, could rename it "Flexible"?

fredrikekre: Or "Composable"?

index.html
ViralBShah: That would be too long a name for the page design.

ViralBShah: I am going to leave out the types and optional typing. That seems like too much detail at this stage of someone's Julia journey.

index.html
```diff
@@ -66,35 +66,35 @@ <h2 class="lead secondary-heading">
...
         <h3>Dynamic</h3>
         <p>Julia is <a href="https://docs.julialang.org/en/v1/manual/types/">dynamically typed</a>, feels like a scripting language, and has good support for <a href="https://docs.julialang.org/en/v1/stdlib/REPL/">interactive</a> use.</p>
+</p>
       </div>
       <!-- 3 -->
       <div class="col-lg-4 col-md-6 feature">
-        <h3>Optionally typed</h3>
-        <p>Julia has a <a href="https://docs.julialang.org/en/v1/manual/types/">rich language of descriptive datatypes</a>, and type declarations can be used to clarify and solidify programs.</p>
+        <h3>Reproducible</h3>
+        <p><a href="https://julialang.github.io/Pkg.jl/v1/environments/">Immutable environments</a> let you recreate the same Julia environment every time, across platforms, with <a href="https://binarybuilder.org">pre-built binaries</a>.</p>
```
JeffBezanson, line 82: I wonder if "immutable environments" is the right term here. You can modify an environment (add/remove dependencies to it). Maybe "environments and immutable package versions"? Or "manifests"? It's also a bit circular (environments let you recreate environments). Maybe "Environments let you recreate an identical software stack every time, ..."? cc @StefanKarpinski

ViralBShah: Coming to think about it, for this position - I don't like having the word "immutable" at all. Maybe something simpler?

StefanKarpinski: "reproducible"?

fredrikekre: I was also thinking that, but that is the also the header. Maybe that is OK though.

ViralBShah: I think using the same words is good to reinforce the point.

PR #974, opened 2020-08-26
Simplify date calculations

zdroid:


logankilpatrick:
Is this even being used anymore @tlienart given we use franklin now?

zdroid:
Yep, it's `{{blogposts}}` function used in `blog/index.md`.

tlienart:
yeah these two changes are much better than what I'd put there

PR #976, opened 2020-08-29
Over 4200 packages

PallHaraldsson:


PallHaraldsson:
I noticed that third page with info that tends to be forgotten... I updated only 3000->4000, 4200 almost are registered 8or if you count the queue too), probably by the time you read this.

fredrikekre:
Can we just say "thousands of packages" and never have to update this again?

ViralBShah:
We really don't need to update this that often. We'll update it when it goes over 5,000. Or around next JuliaCon, whichever comes first.

PR #982, opened 2020-09-11
Julia: A fresh approach to computing

tisztamo:
Removing a single word to completely change the first impression of the landing user.

Based on https://discourse.julialang.org/t/future-directions-of-julia/46269/30

### DIFF
--- a/index.html
+++ b/index.html
@@ -33,7 +33,7 @@
         The Julia Language
       </h1>
       <h2 class="secondary-heading">
-        A fresh approach to technical computing.
+        A fresh approach to computing.
       </h2>
 
       <p>


ViralBShah:
We've avoided doing that because it is a bit misleading. Happy to leave it open to reconsider.

@JeffBezanson @StefanKarpinski Any thoughts?

StefanKarpinski:
Not sure. Let's see what @JeffBezanson thinks. If we're going to change this, I wonder if a different slogan altogether would be better.

fredrikekre:
Could also just remove it? This is the only place where this slogan is still used still AFAIK. It is removed from the startup banner for example.

ViralBShah:
Could go fully minimal and just say "Julia"

StefanKarpinski:
Yeah, that seems good. No need for a slogan anymore really.

JeffBezanson:
I agree we can just remove the slogan. Then we could also make the top banner use a bit less vertical space; it seems very large to me for something that has little content.

PR #983, opened 2020-09-11
Removed slogan + shallow banner

tisztamo:
As discussed in #982

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -62,7 +62,7 @@
   /* font-family: "Futura LT Medium"; */
   font-family: "Roboto", sans-serif;
   font-weight: 100;
-  font-size: 48px;
+  font-size: 46px;
 }
 .secondary-heading {
   /* font-family: "Futura LT Medium"; */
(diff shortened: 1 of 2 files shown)


ViralBShah:
Merging to see how it feels.

fredrikekre:
Looks better with `The Julia Language` IMO. Looks a bit short now, and a bit weird with the logo in the header just there, so it reads like "Julia Julia".

Screenshot from 2020-09-12 00-22-28

---

Screenshot from 2020-09-12 00-22-52

artkuo:
It would be nice to have "language" appear somewhere in top half of landing page. Right now, the first few lines say `julia`, `Julia`, `Julia in a Nutshell`, followed by "Julia was designed to be..." and "Julia programs compile..." The naive reader still doesn't know what it is, and it's only implicit that it's a language. It's not mentioned that it's great for technical computing and data science (among others). Perhaps the previous slogan is overly specific or stale, but _some_ slogan would be beneficial, and should be relatively high up.

fredrikekre:
Perhaps `The Julia Programming Language`?
Screenshot from 2020-09-12 00-30-22

logankilpatrick:
Done @fredrikekre

ViralBShah:
Yeah, just "Julia" was not working. "The Julia Programming Language" is much nicer.

PR #997, opened 2020-09-22
Fix URL in MIT announcement banner

SaschaMann:
The link in the banner on https://julialang.org/ leads to a 404 page: https://computationalthinking.mit.edu/fall20/

The correct URL is https://computationalthinking.mit.edu/Fall20/ with an upper case F, I think.

(This may be a mistake on their end too, I don't know)


logankilpatrick:
Lol, they switched it twice in two days, my guess is the first time broke people linking to it so they reverted.

logankilpatrick:
Either way, thank you!

PR #998, opened 2020-09-23
Use Franklin 0.10

tlienart:
updating the version in the deploy script; note that the updates from 0.9.16 -> 0.10 do not affect this website in any way

* drop of support for an old file structure (pre Franklin 0.6...)
* addition of a `@delay` macro which allows `hfun` to have pages wait for a complete full pass before being executed (relevant for `hfun` which rely e.g. on the list of page tags which is only complete after a full pass over all pages). 

so this should just be good to merge

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Install Julia
         uses: julia-actions/setup-julia@v1
         with:
-          version: 1.5.0
+          version: 1.5
       # This ensures that NodeJS and Franklin are loaded then it installs
       # highlight.js which is needed for the prerendering step.
       # Then the environment is activated and instantiated to install all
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Thank you!

fredrikekre:
https://julialang.org/downloads/#current_stable_release doesn't put that header at the top of the browser. Was that caused by this PR? Most other links, e.g. https://julialang.org/downloads/#download_julia works

tlienart:
ah that's an interesting one, no it's not caused by this PR. it's because the header for `current_stable_release` and, further down `long_term_support` are entered as raw HTML and don't have an anchor (if you click on `Long-term support (LTS) release: v1.0.5 (Sep 9, 2019)` nothing will happen whereas if you click on `Older Releases` just below it, the page will move to it)

if you check https://github.com/JuliaLang/www.julialang.org/blob/master/downloads/index.md you'll see: 

```html
<h2><a href="#current_stable_release">Current stable release: v{{stable_release}} ({{stable_release_date}})</a></h2>
```

which is entered as raw html. it should be like this:

```html
<h2 id=older_releases ><a href="#older_releases">Older Releases</a></h2>
```

note the `id` tag. I'll open a PR.

PR #1000, opened 2020-09-23
fix header anchor on downloads page (issue mentioned in #998)

tlienart:
adds proper anchors to raw html titles on the downloads page (so that when you click on them the page moves to them)

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -27,7 +27,7 @@ Different OSes and architectures have varying [tiers of support](/downloads/#cur
 ---
 
 ~~~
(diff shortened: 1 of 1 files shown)


fredrikekre:
Should they be quoted? Doesn't seem to b e needed but most `id` entries are I think.

fredrikekre (approved):

tlienart:
not when they don't have spaces but it's true that it could be quoted for consistency (though on that page everything is unquoted, this is probably an inconsistency in Franklin which I could fix though it doesn't have an impact bc all generated `id` are without whitespace).

PR #1002, opened 2020-09-24
Add blogpost about the rr memory debugging saga

Keno:


Keno:
Direct link: https://julialang.netlify.app/previews/pr1002/blog/2020/09/rr-memory-magic/

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+
+In this post I want to tell the story of issue [#37524](https://github.com/JuliaLang/julia/issues/37524), the most unusual of the several rr traces we received. As spoiled in the title of this blog post, the eventual issue turned
+out to be faulty memory. However, by popular demand, and to dispel any rumors of the use of black magic, I thought
+it might be worth stepping through the analysis of this issue step by step.
+
+When debugging an rr trace, the first thing to do is usually to just replay it to the point of the crash to look around (sometimes the crash is obvious from the state of the program at the crash location, other times, more history is required). In this case the initial investigation was done by my colleague
+[Tim Besard](https://github.com/maleadt). Downloading the trace and starting the replay he would have seen the following. Note that this recording is sped up about 20x (it runs for about two minutes real time) and lightly edited at the end to focus on only the relevant output.
+
+~~~
+  <asciinema-player author="Keno Fischer"
```
fredrikekre, line 29: This player expands on the page when you press play which is slightly annoying but I don't know if there is anything to do about that.

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+                    theme="asciinema"
+                    title="The initial replay"
+                    src="/assets/blog/2020-09-23-rr-memory/rrmemory.cast">
+  </asciinema-player>
+~~~
+
+So clearly the user who submitted the bug report was doing some sort of episodic procedure (maybe an ML training process), and a couple thousand iterations in, something went wrong. The issue reported the following output:
+```
+signal (11): Segmentation fault
+in expression starting at /home/user/RL-Carla/rl-carla/proofofconcepts/crashes/julia-crash.jl:76
```
fredrikekre, line 40: Should these stacktraces be cleaned from the username of the issue submitter?

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+This is the point where Tim asked me to take over, since that really isn't supposed to happen.
+Now, the error message here may look a bit scary, but let's parse it out part by part. The actual error message
+is:
+
+```raw
+Replaying `SIGNAL: SIGSEGV(det)': expecting tracee signal or trap, but instead at
+ `write' (ticks: 144040993972)
+```
+
+What does this mean? First a quick refresher about how rr works: Basically, it records any modifications made to the process memory by the kernel (or any other source of non-determinism). Then, between such modification it relies on the determinism of the processor to produce a bitwise-identical output state, given an input memory state.
```
fredrikekre, line 87:
```suggestion
What does this mean? First a quick refresher about how `rr` works: Basically, it records any modifications made to the process memory by the kernel (or any other source of non-determinism). Then, between such modification it relies on the determinism of the processor to produce a bitwise-identical output state, given an input memory state.
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+
+Now, what happened here is that rr expected to get a deterministic (`(det)`) segfault (i.e. one caused by the execution of an instruction, as opposed to being asynchronously sent by some other process e.g. using `kill -SIGSEGV`). However, instead of seeing this segfault, we ended up somewhere else. In this case (`but instead at
```
fredrikekre, line 89:
```suggestion
Now, what happened here is that `rr` expected to get a deterministic (`(det)`) segfault (i.e. one caused by the execution of an instruction, as opposed to being asynchronously sent by some other process e.g. using `kill -SIGSEGV`). However, instead of seeing this segfault, we ended up somewhere else. In this case (`but instead at
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+ 'write'`), during the replay, we instead tried to execute a write system call. We've already seen the output behavior of the program (basically one write for every episode), so a reasonable guess is that during replay, it instead sucessfully finished the episode and reached the next `write` system call (which would have been the first
+ execution point at which the `rr` supervisor regains control from the tracee). In rr parlance, we call this a "divergence": A situation in which the execution path taken during replay differs from what actually happened during recording. Divergences are generally caused by rr bugs resulting in unobserved memory modifications (e.g. an imperfect model of the kernel's memory modification behavior), but there are more subtle issues that may be responsible as well (such as CPU microarchitecture bugs). I thus turned to trying to debug this divergence. There are a few common candidates for divergences that I looked for first. In no particular order of preference, they are:
```
fredrikekre, line 91:
```suggestion
 execution point at which the `rr` supervisor regains control from the tracee). In rr parlance, we call this a "divergence": A situation in which the execution path taken during replay differs from what actually happened during recording. Divergences are generally caused by `rr` bugs resulting in unobserved memory modifications (e.g. an imperfect model of the kernel's memory modification behavior), but there are more subtle issues that may be responsible as well (such as CPU microarchitecture bugs). I thus turned to trying to debug this divergence. There are a few common candidates for divergences that I looked for first. In no particular order of preference, they are:
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+
+- Unusual system calls that may not have gotten enough testing in rr, thus making it likely that rr's model of the kernel behavior is incomplete
```
fredrikekre, line 93:
```suggestion
- Unusual system calls that may not have gotten enough testing in `rr`, thus making it likely that `rr`'s model of the kernel behavior is incomplete
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
...h explaining some terminology. While rr does keep track of elapsed wall clock time (`real_time` above), that notion of "time" is entirely informational. Instead, rather than dividing time into hours, seconds and minutes, `rr` divides time into `events`, `tid` (+ serial, not shown and only relevant when tids are re-used) and `ticks`. `events` are a global (over the tree of recorded processes) linear ordering of events that cause external modification to the process' memory (e.g. system calls or signals), whereas `ticks` are a per-task (think per-thread or per-`tid`) measure of forward progress. The exact metric depends on the CPU microarchitecture and what measurement hardware is available. Valid choices are e.g. `Number of retired instructions` or `Number of retired conditional branches`, but in theory any stable, reliable count of forward progress is sufficient, as long as it uniquely identifies (potentially in conjunction with the register state) a particular point in the execution (e.g. number of retired instructions works trivially and retired conditional branches mostly work, because that count in conjunction with the instruction pointer forms a monotonically increasing pair).
```
fredrikekre, line 136:
```suggestion
Let's start with explaining some terminology. While `rr` does keep track of elapsed wall clock time (`real_time` above), that notion of "time" is entirely informational. Instead, rather than dividing time into hours, seconds and minutes, `rr` divides time into `events`, `tid` (+ serial, not shown and only relevant when tids are re-used) and `ticks`. `events` are a global (over the tree of recorded processes) linear ordering of events that cause external modification to the process' memory (e.g. system calls or signals), whereas `ticks` are a per-task (think per-thread or per-`tid`) measure of forward progress. The exact metric depends on the CPU microarchitecture and what measurement hardware is available. Valid choices are e.g. `Number of retired instructions` or `Number of retired conditional branches`, but in theory any stable, reliable count of forward progress is sufficient, as long as it uniquely identifies (potentially in conjunction with the register state) a particular point in the execution (e.g. number of retired instructions works trivially and retired conditional branches mostly work, because that count in conjunction with the instruction pointer forms a monotonically increasing pair).
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+        return tt->instance;
+    jl_ptls_t ptls = jl_get_ptls_states();
+    jl_value_t *jv = jl_gc_alloc(ptls, jl_datatype_size(tt), tt);
+    for (i = 0; i < nargs; i++)
+        set_nth_field(tt, (void*)jv, i, args[i]);
+    return jv;
+}
+```
+
+So we're just getting the type pointer from some array. We can find out which one:
```
staticfloat, line 318: Instantly lost me.  😂 

I'm not sure what the C code is supposed to illustrate; if it's supposed to communicate to us that the code is straightforward, it may not be doing its job.  Perhaps you can shrink it down to 2-3 lines?  I can read a backtrace, so I know that it's focused on `jl_svecset`, of which there is only one in the above function, but there's a lot going on in that function and none of it is commented, so the casual reader will most likely either completely skip this or be confused by it.

It's also not clear what "getting the type pointer from some array" means.  If you could unpack that a bit, (e.g. "getting a pointer to the type information associated with this array, a common operation done by the Julia runtime anytime it needs to access the type information of an object") it would also be helpful.

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+```
+(rr) disas 0x7f049e2ba8e3
+Dump of assembler code for function jl_apply_tuple_type:
+[snip]
+   0x00007f049e2ba8e0 <+96>:  mov    (%rax),%r10
+   0x00007f049e2ba8e3 <+99>:  mov    -0x8(%r10),%r9
+[snip]
+```
+
+Alright, so that's a load from `r10`. Luckily `r10` is recorded in the trace: `r10:0x1007f0490b784d0`. A seasoned debugger will quickly recognize this pointer as too wide. In particular, this pointer has the lowest bit of its high byte set. These bits are mostly unused. Sometimes unused pointer bits are used for extra information, but at least julia itself only uses the unused low bits for this purpose, not the high bits. That said, this trace included a significant number of external native libraries (including a full python environment), so it's certainly possible that some external library would have used such pointer tagging techniques. Let's do some more investigation. The first thing I tried to do was take a look at the pointer with the high bit manually cleared to see if it was a valid julia object at all or just some junk that somehow ended up being loaded from:
```
fredrikekre, line 172:
```suggestion
Alright, so that's a load from `r10`. Luckily `r10` is recorded in the trace: `r10:0x1007f0490b784d0`. A seasoned debugger will quickly recognize this pointer as too wide. In particular, this pointer has the lowest bit of its high byte set. These bits are mostly unused. Sometimes unused pointer bits are used for extra information, but at least Julia itself only uses the unused low bits for this purpose, not the high bits. That said, this trace included a significant number of external native libraries (including a full Python environment), so it's certainly possible that some external library would have used such pointer tagging techniques. Let's do some more investigation. The first thing I tried to do was take a look at the pointer with the high bit manually cleared to see if it was a valid julia object at all or just some junk that somehow ended up being loaded from:
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+
+```
+(rr) p jl_(0x7f0490b784d0)
+Array{Any, 1}
+$1 = void
+(rr) 
+```
+
+Ok, so this is actually a pointer to a julia object (the type `Array{Any, 1}`) with a corrupted high bit. Let's gather some more information. We're still 10000 ticks away from the place where the segfault was supposed to occur, but let's see what's currently happening in the process:
```
fredrikekre, line 181:
```suggestion
Ok, so this is actually a pointer to a Julia object (the type `Array{Any, 1}`) with a corrupted high bit. Let's gather some more information. We're still 10000 ticks away from the place where the segfault was supposed to occur, but let's see what's currently happening in the process:
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+
+```
+(rr) b *0x7f049e2ba8e3
+Breakpoint 1 at 0x7f049e2ba8e3: file /buildworker/worker/package_linux64/build/src/julia.h, line 1222.
+(rr) c
+Continuing.
+[FATAL /home/keno/rr/src/ReplaySession.cc:636:check_pending_sig()] 
+```
+
+we just get the divergence again, so it never actually reached this execution point, but diverged earlier (but only happened to crash there). Can we find out what happened? Looking back at the assembly, we see that r10 was loaded from
```
fredrikekre, line 226:
```suggestion
we just get the divergence again, so it never actually reached this execution point, but diverged earlier (but only happened to crash there). Can we find out what happened? Looking back at the assembly, we see that `r10` was loaded from
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+`rax`, which we also have: rax:0x436487c0. What's there currently (remember we're still 10000 ticks before the crash):
```
fredrikekre, line 227:
```suggestion
`rax`, which we also have: `rax:0x436487c0`. What's there currently (remember we're still 10000 ticks before the crash):
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+unexpected that messes with it:
+
+```
+(rr) p &((uint64_t**)0x7f0478f5aa10)[-1]
+$9 = (void **) 0x7f0478f5aa08
+(rr) p jl_((uint64_t)*$9 & ~0x3)
+Array{Any, 1}
+```
+
+(note that we masked out the low bits, which julia does use as mentioned above - and
```
staticfloat, line 356: I would explicitly tie together the chain of events so far, as I think you might lose some people here.  Maybe something like:

> (jnote we masked out the low bits, which julia does use as mentioned above - and if we looked at the assembly code, it does the same. To recap: The original value we were investigating, `0x1007f0490b784d0` was found to be a pointer to a Julia type object, with a bit flipped.  Looking at where that pointer was used, we found an array that points to that type object, whose address is `0x7f0478f5aa08`.  We have now found the object that holds the pointer whose bit got flipped.

Or something like that.

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+    at /buildworker/worker/package_linux64/build/src/jltypes.c:560
+560 in /buildworker/worker/package_linux64/build/src/jltypes.c
+(rr) 
+Continuing.
+[FATAL /home/keno/rr/src/ReplaySession.cc:636:check_pending_sig()] 
+ (task 19490 (rec:7760) at time 51911)
+ -> Assertion `false' failed to hold. Replaying `SIGNAL: SIGSEGV(det)': expecting tracee signal or trap, but instead at `write' (ticks: 144040993972)
+```
+
+We see some accesses from the type cache, and then the divergence. What about in the other direction? (Note that the divergence crashed the rr process, so we have to first get back to the same execution point - luckily since this is a recording, everything is stable, so we can just `seek-ticks` right back to where we were).
```
fredrikekre, line 263:
```suggestion
We see some accesses from the type cache, and then the divergence. What about in the other direction? (Note that the divergence crashed the `rr` process, so we have to first get back to the same execution point - luckily since this is a recording, everything is stable, so we can just `seek-ticks` right back to where we were).
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+(rr) bt
+#0  jl_svecset (x=0x7f0490b784d0 <jl_system_image_data+16912>, i=<optimized out>, t=<optimized out>)
+    at /buildworker/worker/package_linux64/build/src/julia.h:840
+#1  jl_f_tuple (F=<optimized out>, args=0x42f54590, nargs=46200)
+    at /buildworker/worker/package_linux64/build/src/builtins.c:728
+```
+
+For reference (from the 1.5 release branch):
+
+```
```
fredrikekre, line 287:
```suggestion
```c
```

blog/2020/09/rr-memory-magic.md
fredrikekre, line 356:
```suggestion
(note that we masked out the low bits, which Julia does use as mentioned above - and
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+tested, upstreamed, etc.), we managed to remotely and conclusively diagnose that a reported
+crash was caused by faulty memory on the users machine.  Even though the rr trace did not
+contain the actual cause of the crash, it was nevertheless
+an invaluable tool to analyze this issue. The fact that the stray bit flip was missing
+from the recording already excluded 99% of possible cases for the crash, and some
+careful analysis of the trace gave enough clues to be able to eliminate most of the
+remaining. To quote Sherlock Holmes "When you have eliminated the impossible,
+whatever remains, however improbable, must be the truth". Without rr, it may often
+be tempting to blame unexplained crashes on bad hardware, cosmic rays or gremlins
+living under the floors. Here, we were able to fairly convincingly conclude that
```
staticfloat, line 500:
```suggestion
living under the floor. Here, we were able to fairly convincingly conclude that
```

staticfloat (commented):
This was really fun!  Great work!

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+then `0x5c` address bits set, we can probably make the assumption that the bit that was flipped
+was that of the array itself (whose address ended in `0x8`, i.e. having the high dword at `0xc`),
+rather than the copy of the type tag in the svec (whose address ended in `0x0`), but none of the
+detected failures exactly match the address offset of one of the two addresses we identified as
+potential targets for the bit flip (though the test in the screenshot was not run to completion,
+so maybe it would have found one). Further analysis of the failure would require physical access
+to the DRAM and perhaps an electron microscope, but we need not go there, because we've already
+determined this is not a Julia bug and the user will probably be replacing their machine.
+
+# Conclusion
```
fredrikekre, line 488:
```suggestion
## Conclusion
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+
+With about an hour's work (plus 2-3 hours to add the `seek-to-ticks` functionality written,
+tested, upstreamed, etc.), we managed to remotely and conclusively diagnose that a reported
+crash was caused by faulty memory on the users machine.  Even though the rr trace did not
```
fredrikekre, line 492:
```suggestion
crash was caused by faulty memory on the users machine.  Even though the `rr` trace did not
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+contain the actual cause of the crash, it was nevertheless
+an invaluable tool to analyze this issue. The fact that the stray bit flip was missing
+from the recording already excluded 99% of possible cases for the crash, and some
+careful analysis of the trace gave enough clues to be able to eliminate most of the
+remaining. To quote Sherlock Holmes "When you have eliminated the impossible,
+whatever remains, however improbable, must be the truth". Without rr, it may often
```
fredrikekre, line 498:
```suggestion
remaining. To quote Sherlock Holmes
> When you have eliminated the impossible, whatever remains, however improbable, must be the truth.
Without `rr`, it may often
```

blog/2020/09/rr-memory-magic.md
```diff
@@ -0,0 +1,503 @@
...
+
+One good way to investigate a divergence is to poke around in memory a bit prior to when one suspects the divergence to
+have occurred and see what kind of memory modifications could have made the difference. Usually, I just do this by going to the prior event, but in our case, there are 15 million conditional branches between the last two events, which is quite a bit of distance. Instead I hacked up a [quick command](https://github.com/mozilla/rr/pull/2658) that lets us seek directly to a particular tick count. In this case, I arbitrarily chose `143998750000`, about 10000 ticks prior to when the segfault occurred during the recording. Let's go there:
+
+```
```
fredrikekre, line 143: I think this will highlight as Julia code, but it looks pretty good anyway for the `rr` traces.

fredrikekre (commented):
Mostly formatting and some typos.

StefanKarpinski:
A few edits:

> However, instead of seeing this segfault, we ended up somewhere else. In this case (but instead at ‘write’), during the replay, we instead tried to execute a write system call

would be clearer as
> However, instead of seeing this segfault, we ended up somewhere else: instead at 'write'. In this case, during the replay, rr found itself trying to execute a valid write system call rather than segfaulting

> Ok, so this is actually a pointer to a julia object (the type Array{Any, 1}) with a corrupted high bit.

maybe:
> Barring a massive coincidence, this is an otherwise-valid pointer to a julia object (the type Array{Any, 1}) with a corrupted high bit.

I’m a bit lost in the middle part where you’re hunting down the precise divergence point—maybe a bit more explanation of each step and why you’re doing it.

> memory on the users machine

should be:

> memory on the user’s machine

great post, especially the conclusion!

blog/2020/09/rr-memory-magic.md
Keno: I like floors, indicating the abstract concept of living under peoples' floors rather than a specific floor.

blog/2020/09/rr-memory-magic.md
Keno: This is an implementation detail, so the style guide suggests julia (the program/implementation) rather than Julia (the abstract language). I'll code-format it.

blog/2020/09/rr-memory-magic.md
Keno: Fair. The trace is public, so anybody can see it, but I'll strip it from the blogpost

blog/2020/09/rr-memory-magic.md
Keno: yep

blog/2020/09/rr-memory-magic.md
Keno: I'll use bash highlighting.

Keno:
Thanks for the feedback. Everyone. Merged. Of course as always, additional feedback can happen via comment here or direct PR.

PR #1004, opened 2020-09-25
Add relevant image preview for social sharing of new rr post

Keno:


Keno:
@tlienart This ends up with two `og:image` tags on the page. What's the correct way to ensure only one shows up?

tlienart:
This is controlled in the utils.jl file, I'll check in a couple of hours when I have a laptop!

tlienart:
Ah ok got it.

### why there's a problem

* The `layout/meta.html` has this line: https://github.com/JuliaLang/www.julialang.org/blob/c93f98d28156a5e264f16ca325a34c3d5d563359/_layout/meta.html#L7 which is added to *every* page
* Your page has an additional meta specification which is added at another point

the result is a duplicate `og:image` as you point out. 

I'll clean this up in a separate PR.

tlienart:
now superseded by #1005 ; closing here.

PR #1005, opened 2020-09-25
fixing meta tags + hfun

tlienart:
supersedes #1004 

* improves the `hfun_meta` so that it has default values that are applied on all pages for `og:xxx` fields which can be overwritten by the user
* simplify the way `meta` tags are specified + adjust the definitions wherever they were used

Note: there's a comment about `pubdate` to leave that definition out of the `+++` block for now, this is a small bug in Franklin which I'll fix within the hour.

### DIFF
--- a/utils.jl
+++ b/utils.jl
@@ -13,13 +13,13 @@ A full example can be found in `blog/2020/05/rr.md`.
 """
 function hfun_meta()
     p = "property"
-    # default og properties, can be over-written by the user
+    # default og properties, can be overwritten by the user
     ogdflt = (
         title = (p, "og:title", "The Julia Language"),
         image = (p, "og:image", "/assets/images/julia-open-graph.png"),
         descr = (p, "og:description", "Official website for the Julia programming language"),
         )
-    # check what the user has provided (if anything) and override defaults
+    # check what the user has provided (if anything) use defaults otherwise
     meta = locvar(:meta)
     if !isnothing(meta)
         for c in keys(ogdflt)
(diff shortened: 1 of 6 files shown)


tlienart:
You can see in the source of https://julialang.netlify.app/previews/pr1005/blog/2020/09/rr-memory-magic/

that 

```html
<meta property="og:image" content="https://julialang.org/assets/blog/2020-09-23-rr-memory/faulty-mem.jpg">
```

(and there's only one).

utils.jl
```diff
@@ -4,21 +4,33 @@ using Dates
...
-    m = locvar(:meta)
-    io = IOBuffer()
-    for tuple in locvar(:meta)
-        write(io, "<meta ")
-        for (prop, val) in tuple
-            write(io, "$prop=\"$val\" ")
+    p = "property"
+    # default og properties, can be overwritten by the user
+    ogdflt = (
+        title = (p, "og:title", "The Julia Language"),
```
Keno, line 18: On Blog pages, can we default this to the post title, e.g. here's what it currently looks like on facebook:

Screen Shot 2020-09-25 at 7 17 15 AM

Rather than the "The Julia Language" can we do the post title for og:title if we're on the blog?

Keno:
I'll go ahead and merge this, since it's better than nothing. We can fix the title separately.

utils.jl
tlienart: yes, I'll open another PR with that

PR #1007, opened 2020-09-25
Use page title and descr as default `og:title` `og:description` if they are defined

tlienart:
* Use the title (or if not defined, the first header) as `og:title` 
* Use the rss description as `og:description` 

And if they aren't defined fall back on respectively "The Julia Language" and "Official website for ...". 

Follow up from #1005


Keno:
Works!

PR #1019, opened 2020-09-27
Update index.md

wchoston:
Changed 'has' to 'have' to match 'members'.


logankilpatrick:
Thank you! Good catch.

PR #1021, opened 2020-09-29
[WIP] Add new D&I Page

logankilpatrick:
This is a WIP update to our D&I page based on current initiatives by folks in the community.


ViralBShah:
I think there might be a couple of julialang blogs also that @xorJane wrote post JuliaCon with some stats. Also can link to the community surveys that provide some data.

ViralBShah:
I wonder if we should include a chart from Google analytics data on % by gender over time.

SaschaMann:
The sub-pages are all fairly short, I wonder if it might be better to embed them rather than link to them? Makes it easier to get an overview imo.

logankilpatrick:
> The sub-pages are all fairly short, I wonder if it might be better to embed them rather than link to them? Makes it easier to get an overview imo.

The sub-pages have to be built out. They are blank right now.

PR #1024, opened 2020-09-30
added function to show time zone according to user local timezone

devil-cyber:
hey maintainer need some feedback on my code

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -100,12 +100,12 @@
        <div class="col-lg-4 col-md-3 language-features"><hr/></div>
      </div>
 
-     <br>
+     <br>    
 
      <p>The Julia Community has a shared calander for all upcoming global events. If you are an event organizer, please <a href="mailto:logan@julialang.org">email us</a>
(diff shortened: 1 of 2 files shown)


ViralBShah:
This may be for @tlienart to review.

tlienart:
Hello! 

A few points from important to not important

**Important**

* I don't think this will do what you want it to do, the code will be evaluated on the build system which means that the only zone you will ever get out of this that will be displayed on the website is the time zone of the GitHub build system. I imagine that's not what you want.
* I think if what you want is to show the timezone of the user, you'd need to recover that info from the browser, doesn't google calendar do that automatically when it's on UTC though? 

**Less important**

* The spirit of your code works well, btw `localzone()` never returns an empty string, it's either a valid timezone or an error is raised, so you could just have written `hfun_timezone() = localzone.name`
* You added a lib to `utils.jl` (`TimeZones`), in order for the site to build properly, you would need to add that dependency to the project as well (update the `Project.toml`).

devil-cyber:
Ok thanks for the feedback I will try to improve this.

PR #1031, opened 2020-10-08
Update classes.md

valenmgama:
Include Marzieh Khakifirooz course centered on JuMP Linear Optimization.


logankilpatrick:
Thank you!

PR #1040, opened 2020-10-14
add Julia High Performance, 2nd Edition

aviks:
I just realised that this book had not been added to the list.


logankilpatrick:
Thank you!

PR #1042, opened 2020-10-15
Add Julia teaching activities at the University of Basel

frankschae:
We are currently using Julia in the "Classical Mechanics" course by Prof. Christoph Bruder at the University of Basel.


logankilpatrick:
Thank you!

PR #1049, opened 2020-10-20
Update Emacs logo to current one

taingra:
New Emacs logo taken from the official GNU website https://www.gnu.org/software/emacs/


logankilpatrick:
Thank you!

PR #1053, opened 2020-10-21
Pin python version in github-action

tlienart:
(should be safe to merge)

This pins the python version to `3.8`, python is only used in a very minor way on the website (for the magnification) so that won't change anything, however it will prevent potential CI errors with python 3.9 in the eventuality where someone were to use PyPlot and require matplotlib for instance.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Install python
         uses: actions/setup-python@v1
         with:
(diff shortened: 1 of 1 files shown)


DilumAluthge:
@tlienart @logankilpatrick Is this good to merge?

logankilpatrick:
Yes!

logankilpatrick:
Thanks for reminding me.

PR #1054, opened 2020-10-23
specify google group

apparluk:
Specify use of a Google Group.

### DIFF
--- a/community/stewards.md
+++ b/community/stewards.md
@@ -6,7 +6,7 @@ The spirit of community is crucial to free/open-source development. In the large
 
 ## Stewards of the Community Standards
 
-All members of the Julia community are expected to abide by the community standards and encourage others to do so. To ensure that some group has the obligation and authority to cope with conflicts, a group known as the “Julia Stewards” has been created. The current membership of this group is:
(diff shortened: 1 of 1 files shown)


timholy:
I think this kind of missed the point. `group` is being used in two ways, "group of people" and "Google Group." The former is more important---how you email a collection of people is a pretty minor detail.

But it might be worth disambiguating these; want to try a different approach? I might use the word "committee" instead of "group" to refer to the "group of people."

logankilpatrick:
Hey, not sure as to the context of this PR. Feel free to elaborate more if need be.

apparluk:
It's a liability issue, who has access to what information and where is it stored.

It's the similar with the "non-steward members" content.

Consult with counsel.

logankilpatrick:
Thanks for the heads up.

apparluk:
membership lists 7

therefore quorum

would be better than

3+ disinterested stewards (possibly including non-steward members)

apparluk:
@logankilpatrick 

If the developers don't want to inform the community of the use of 3rd Party servers in processing complaints that is their own business.

redacted content as appears at the base of e-mails sent in response to complaints

`You received this message because you are subscribed to the Google Groups "Julia Stewards" group.`

PR #1058, opened 2020-10-26
Fix PR comments

DilumAluthge:
For posterity, fixing this is done by:
1. Go to the repo for the action in question: https://github.com/thollander/actions-comment-pull-request
2. Go to the commits: https://github.com/thollander/actions-comment-pull-request/commits/master
3. Get the latest commit

@logankilpatrick Note that previews will not be created for this PR, since it is from a fork. I'd recommend that copy the change from this PR, and then make your own PR that is not from a fork; that way we can confirm that this fixes the PR previews.


logankilpatrick:
"Hey Siri, what does posterity mean" 

Thank you nonetheless!

logankilpatrick:
Thank you again!

PR #1061, opened 2020-10-26
Over 4,500 package

PallHaraldsson:
[skip ci]

### DIFF
--- a/index.html
+++ b/index.html
@@ -366,7 +366,7 @@ end
 
       <br>
 
-      <p>Julia has been downloaded over 17 million times and the Julia community has registered <a href="https://juliahub.com">over 4,000 Julia packages</a> for community use.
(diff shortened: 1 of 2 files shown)


giordano:
I thought [it was said](https://github.com/JuliaLang/www.julialang.org/pull/976#issuecomment-683336791)

> We'll update it when it goes over 5,000. Or around next JuliaCon, whichever comes first.

?

PallHaraldsson:
Yes, actually I updated that file, when I meant update the front-page...but I do not see a good reason not to keep in sync.

You can merge the latter commit or just both.What was said didn't apply to the front page.

ViralBShah:
No it is not misleading. It is not necessary for this page to be in sync with Julia Computing.

PR #1066, opened 2020-10-29
Fix placeholder expansions in downloads/platform, fixes #1065.

### LINKED ISSUE #1065: Downloads page instructions broken (template rendering error?) for MacOS & Linux Official Binaries
When visiting the page https://julialang.org/downloads/platform/, the instructions for downloading the binaries do not seem to be properly substituting in the versions for `{{stable_release_short}}` or `{{ stable_release }}`. This makes it real difficult to figure out the URLs to download.

This appears to affect both the macOS and Linux&FreeBSD instructions.

As a guess, is the template not evaluating those because it is within a code block?

Edit: I was able to find the correct values in this file, https://github.com/JuliaLang/www.julialang.org/blob/master/config.md, which at time of writing are 1.5 and 1.5.2, respectively.

fredrikekre:


tlienart:
You can get this with ```` ```shell ```` ? Why go for raw html?

fredrikekre:
The problem is not the highlighting, it is that `{{...}}` are not expanded in code.

tlienart:
ah of course, I'll merge :+1:

downloads/platform.md
```diff
@@ -65,10 +65,9 @@ To start running Julia from the Terminal, you can do the following:
 
 Navigate to `/usr/local/bin` and remove the `julia` file. Then type the following command:
 
-
-```
-ln -s /Applications/Julia-{{stable_release_short}}.app/Contents/Resources/julia/bin/julia /usr/local/bin/julia
-```
+~~~
+<pre><code class="language-shell">ln -s /Applications/Julia-{{stable_release_short}}.app/Contents/Resources/julia/bin/julia /usr/local/bin/julia</code></pre>
```
tlienart, line 69: small nit, I think it should be

```
classs="lang-shell hljs"
```

to ensure highlight picks it up

tlienart (changes requested):

downloads/platform.md
tlienart: it probably doesn't matter I don't think it'll be highlighted anyway

fredrikekre: I believe that is the post-pre-rendering version, before that it is just `language-XXX`, no?

tlienart: yes

fredrikekre: Anyway, seems to have succesfully deployed and prerendered into:
```
<code class="shell hljs">wget https://julialang-s3.julialang.org/bin/linux/x64/1.5/julia-1.5.2-linux-x86_64.tar.gz
tar zxvf julia-1.5.2-linux-x86_64.tar.gz</code>
```

PR #1067, opened 2020-10-30
adding explicit language for shell blocks

tlienart:
Un-qualified code blocks default to `Julia` as a language which means that something like

```
export PATH=...
```

will have the `export` be treated as a keyword. This is very minor but the PR ensures these blocks are treated as shell language.


fredrikekre:
For that particular example the Julia highlighting is pretty good actually; it highlights the `export` and also the interpolation, but yea, this is more correct.

PR #1072, opened 2020-11-08
the legal data page is no longer needed

StefanKarpinski:
There are no longer any Julia versions in use that send any data that requires this legal data notice, so I'm taking it down. We should, however, have some kind of legal notice here, but that's work for another PR.


ViralBShah:
I suggest removing it altogether for now. When we have a legal page can always add it back.

StefanKarpinski:
Will it break the formatting of the footer if I delete a row from one of the columns?

logankilpatrick:
Check the PR preview link above, looks good to me.

ViralBShah:
No the footer formatting won't break.

StefanKarpinski:
I left the placeholder `/legal/` page and removed the link to it. Makes it easier to put something there.

PR #1075, opened 2020-11-09
JSoC projects for DeepChem.jl

rkurchin:
Proposing these additions by PR per @ChrisRackauckas 's suggestion.

### DIFF
--- a/jsoc/gsoc/deepchem.md
+++ b/jsoc/gsoc/deepchem.md
@@ -1,0 +1,22 @@
+# DeepChem.jl development projects – Summer of Code
+
+## Towards DeepChem.jl: Combining Machine Learning with Chemical Knowledge
+
+We have been developing the AtomicGraphNets.jl package, which began modestly as a Julia port of [CGCNN](https://github.com/txie-93/cgcnn), but now has plans to expand to a variety of more advanced graph-based methods for state-of-the-art ML performance making predictions on atomic systems. In support of this package, we are also developing ChemistryFeaturization.jl, which contains functions for building and featurizing atomic graphs from a variety of standard input files. ChemistryFeaturization will eventually form the bedrock of a DeepChem.jl umbrella organization to host a Julia-based port of the popular [Deepchem](http://deepchem.io) Python package.
+
+Some of the features we're excited about working on include:
+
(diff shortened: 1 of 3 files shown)


ChrisRackauckas (approved):

timholy:
Excited to see this! Where is AtomicGraphNets.jl? I am starting to build related infrastructure, including the ability to query and analyze structures on PubChem for structural diversity, and fit ligand-binding data to a model of the binding pocket.

rkurchin:
Oops sorry, should maybe have included that link! The packages aren't registered yet (I want to at least have some skeletal docs before doing so, and those are still a work in progress), but you can find them here: [ChemistryFeaturization](https://github.com/aced-differentiate/ChemistryFeaturization.jl), [AtomicGraphNets](https://github.com/aced-differentiate/AtomicGraphNets.jl)

timholy:
Do you already have a plan for handling chirality? That's one of the early things I was expecting to have to tackle.

rkurchin:
Ooh, fun question! I haven't thought much about it because for the near term at least, we're mostly interested in inorganic solids, where those effects don't tend to be important (except maybe sometimes in magnetic behavior). But I'm deliberately trying to put the code through its paces on different types of problems (e.g. playing with QM9 right now), so while I can't promise to give it high priority myself, I'm certainly open to having discussions about how these things could/should be handled and being cognizant not to inadvertently close off important pathways in this regard...

If you'd be interested in co-mentoring something aimed in that direction, I'd definitely be open to it! Happy to chat more  about this (chirality) or any other mutually interesting directions!

timholy:
Sounds awesome. It's great to know that others are interested in related areas, and maybe even good that we have different foci since it may help increase the generality of tools. In the short term my main plan is to work on a package, PubChemCrawler.jl, for searching and analyzing structures from PubChem, and to make necessary modifications to OpenSMILEs to parse and represent chirality. The things closer to your interest in learning models will be focused on pretty specific structures (steroids) but there may nevertheless be some ways in which we can synergize.

WRT JSoC, I may mentor JuliaImages projects so I don't know what I can promise, but certainly if you get a good candidate ping me and I can check availability.

PR #1078, opened 2020-11-11
Add queryverse.

ma-laforge:
Should we first ask owners if they are ok with this?

I really think this it should be added to the list of organizations.


logankilpatrick:
No worries, if they don't want to be shown, they can remove their org. Thanks again!

PR #1083, opened 2020-11-13
 Add a few more organizations.

ma-laforge:
Also create new categories to keep Scientific more manageable:
 - Astronomy/Space
 - Physics/Quantum mechanics

### DIFF
--- a/community/organizations.md
+++ b/community/organizations.md
@@ -5,26 +5,26 @@ The following is a non-comprehensive list of Julia GitHub groups grouped by doma
 ### Julia
 
 @@tight-list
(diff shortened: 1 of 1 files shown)


ViralBShah:
Are all the orgs that you are adding reasonably active?

ma-laforge:
Well, 1 or 2 of them seem less actively maintained. I was hoping they could still act as a nucleation site to get people to contribute.

**=> Yes, I suppose it might be good to have another page with only high-activity orgs.**
But if there are only a few that aren't, highlighting them on this page might be good as well. That way people can redistribute code to/from them, and possibly dismantle them if they are no longer needed.

PR #1084, opened 2020-11-15
Fix broken link to MLJ projects, caps error in projects.md

jpnski:
Following the current link to the MLJ projects (https://julialang.org/jsoc/gsoc/mlj/) from the Projects page results in 404. I believe the correct link it should point to is: https://julialang.org/jsoc/gsoc/MLJ/

Updated projects.md to reflect this difference in capitalization to successfully link to the MLJ projects page.


logankilpatrick:
Thank you! Good catch.

PR #1090, opened 2020-11-23
Add DFTK JSOC project

mfherbst:
Pining @antoine-levitt if he has any comments.

Also I was not too sure where to put the project info. There is a science.md, but then many of the projects are kind of related to science. Happy for feed back if I should place the project somewhere else.

### DIFF
--- a/jsoc/gsoc/dftk.md
+++ b/jsoc/gsoc/dftk.md
@@ -1,0 +1,60 @@
+# DFTK.jl development projects – Summer of Code
+
+## Automatic differentiation in density-functional theory
+
+Density-functional theory (DFT) is probably the most widespread method
+for simulating the quantum-chemical behaviour of electrons in matter
+and applications cover a wide range of fields such as materials
+research, chemistry or pharmacy. For aspects like designing the
+batteries, catalysts or drugs of tomorrow DFT is nowadays a key
+building block of the ongoing research. The aim to tackle even larger
+and more involved systems with DFT, however, keeps posing novel
+challenges with respect to physical models, reliability and
+performance. For tackling these aspects in the multidisciplinary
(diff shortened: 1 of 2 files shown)


antoine-levitt:
LGTM!

ViralBShah:
@logankilpatrick @tlienart How do we fix the failing checks?

mfherbst:
Thanks @ViralBShah for the merge!

tlienart:
@ViralBShah seems like GithubAction deprecated `set-env`: 

> Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

looks like they will fully disable that; it's actually @DilumAluthge who knows this stuff (the lines https://github.com/JuliaLang/www.julialang.org/blob/ce5d82154987f397797b1132ae5ae4ee824d1cea/.github/workflows/deploy.yml#L23-L28), Dilum would you have an idea how to bypass this short of using the `ALLOW_UNSECURE_COMMANDS` which doesn't sound too good? thanks 🙌

giordano:
Do you basically want to set a variable in a job and use its value in subsequent jobs?  You can do it with [`fromJSON`](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#fromjson)

DilumAluthge:
Yeah gimme one sec

DilumAluthge:
@tlienart @ViralBShah https://github.com/JuliaLang/www.julialang.org/pull/1091

Issue #1070, opened 2020-11-08
Links in the diversity section goes to empty pages or nowhere


KristofferC:
The links here https://julialang.org/diversity/current/ leads to empty pages.

In https://julialang.org/diversity/ideas/, the link to ASL interpretation goes nowhere.

Perhaps they should just be listed and not have links until there is any content for them.

logankilpatrick:
Was hoping folks would have had the chance to update the details for their projects by now but if these don't get updated soon, we can revert to just a list.

ViralBShah:
We should not set up empty links with the hope that someone will populate them. Since people get busy and such, it is usually better to get the content ready and up first, and then set up a link.

logankilpatrick:
That's fair, will update them to a normal list this weekend.

PR #1097, opened 2020-12-01
remove some links from diversity pages that lead nowhere

KristofferC:
Fixes https://github.com/JuliaLang/www.julialang.org/issues/1070.

I think the list of the active projects would have more impact if they had a bit more context to them. Right now, it isn't clear if it is just a wish list or something that is actually ongoing. For example, the "Translation of Julia materials into Spanish" could have a link to where that material is.

### DIFF
--- a/diversity/current/JuliaES.md
+++ b/diversity/current/JuliaES.md
@@ -1,1 +1,0 @@
-# Julia ES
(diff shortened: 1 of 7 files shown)


logankilpatrick:
Thank you!

ViralBShah:
We should absolutely not have dead links in the main parts of the website. Can't do much about old blogs and such, but the core website should not have broken links.

logankilpatrick:
They weren't dead persay, just pages with no content because I put the hooks in for folks to update their respective projects but that never got done and I forgot to remove the links (so thanks for this Pr). Will probably get these pages up and running soon once folks have the energy to write about their work (early 2021).

PR #1098, opened 2020-12-01
use a Project + Manifest to record the dependencies used

KristofferC:
Without using a Project, new versions of dependencies that might be buggy can make the site act wierdly. Better to just record them and do a manual check when updating dependnecies.


ViralBShah:
@KristofferC Can you merge if this works to your satisfaction?

logankilpatrick:
I was going to test locally but if it works for @KristofferC then it should be good to go

Issue #1009, opened 2020-09-25
Add "about the author" section for blogs?


Keno:
On multi-author blogs, I think it's common to have a short "about the author" blurb at the end, e.g. (just taking the first one I found via HN):

https://blogs.embarcadero.com/bold-for-delphi-is-open-source/ 

<img width="766" alt="Screen Shot 2020-09-25 at 9 36 55 AM" src="https://user-images.githubusercontent.com/1291671/94273722-c7838180-ff12-11ea-9117-0a2c21f00725.png">

I think that would be a good addition to the blog.

logankilpatrick:
This would be great to have. @tlienart, I am guessing the right way to approach this would be in Franklin?

tlienart:
this can be done without any modifications in Franklin; the stuff that's needed is the content: the bio and photo; if you get me a few of those I can do a poc

logankilpatrick:
Great, perhaps we can just add @Keno 's name to his most recent blog post along with his Twitter profile image. Alternatively you can use my name for one of the recent blog posts I've done and my GitHub profile image.

ViralBShah:
I don't think this is really necessary for now.

PR #1100, opened 2020-12-01
adding author blurb #1009

tlienart:
cc @logankilpatrick @Keno 
context: #1009 

what it looks like:

Screen Shot 2020-12-01 at 20 54 04

this is just a draft but should be pretty easy to extend; I'm a bit squeezed these days so maybe @logankilpatrick if you'd like to take this on? what would need to be done:

1. possibly a touch of CSS adjustments to make sure this looks 'right'
1. adjusting the `hfun_about_the_author` in `utils.jl` (e.g. if you want the possibility for more social images or modifying the text a bit
1. square (width=height) mugshots should be added in `_assets/authors/...` and referred to as the `author_img` page  variable
1.  author blurb is the `author_blurb` page variable

A note on the last two points is that it will be more maintainable to have a dictionary of author img+blurbs; this could be declared in `utils.jl` as well; and then you'd do `AUTHOR_BLURBS.get(locvar(:author), "")` so that there's only one place where this would need to be maintained.


logankilpatrick:
@Keno Mind adding in your image to this PR? Or is it otherwise okay to add your image?

PR #1101, opened 2020-12-02
downloads: Add link to versions.json

SaschaMann:
This file contains info on all published Julia versions (see https://github.com/JuliaLang/julia/issues/33817 for more context).
It's currently not easy to find but may be useful for people who build Julia tooling.

Generation script: https://github.com/SaschaMann/Julia-versions.json/blob/main/build_json_map.jl (written by `@staticfloat`)
Repo that builds and deploys it: https://github.com/SaschaMann/Julia-versions.json

---

I don't really know where this belongs on the page, so I just added it to the end. Let me know if I should move it elsewhere.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -349,3 +349,9 @@ The platforms currently supported by Julia are listed below. They are divided in
 </ul>
 ~~~
 @@ @@
+
+
+
+## JSON release feed
+
+The info above is also available as a [JSON file](https://julialang-s3.julialang.org/bin/versions.json). The file is updated once every 24 hours.
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -354,4 +354,4 @@ The platforms currently supported by Julia are listed below. They are divided in
 
 ## JSON release feed
 
-The info above is also available as a [JSON file](https://julialang-s3.julialang.org/bin/versions.json). The file is updated once every 24 hours.
+The info above is also available as a [JSON file](https://julialang-s3.julialang.org/bin/versions.json) ([schema](https://github.com/SaschaMann/Julia-versions.json/blob/main/schema.json)). The file is updated once every 24 hours.


logankilpatrick:
Nice! This looks helpful.

ViralBShah:
If this is an official sort of thing, should it go into some JuliaLang repo?

logankilpatrick:
It's pulling from S3 so it's getting the "official" data. Would make sense to have it in the Julia org IMO.

logankilpatrick:
Only incentive to leave it outside the JuliaLang org is if we are crunched for GitHub action CI minutes.

SaschaMann:
> If this is an official sort of thing, should it go into some JuliaLang repo?

I'm happy to transfer the repo to an org. One thing to consider is that access to the secrets in the repos should be somewhat limited (see https://github.com/JuliaLang/julia/issues/33817#issuecomment-713017624), that's why it probably shouldn't live in julialang/julia for example.

Since setup-julia for GitHub Actions and potentially other tools to install Julia pull data from it, modifying it could allow someone to modify it to point it at a malicious download URL instead (there are precautions one can take in the tool itself, but it's imo best to not rely on those exclusively). The more accounts have access to it, the larger the attack surface. I don't know how the JuliaLang org is set up, so I can't tell how many accounts would have access to it. When setting it up, Elliot said he was fine with it living in that repo, that's why it lives there for the time being. Alex also has access to it. (to be clear: this is not about trusting the people who have access to the repo, but rather about every account adding a potential entrypoint for some external actor; and GH unfortunately lacks tools for more fine-grained control for accessing secrets)

SaschaMann:
> Only incentive to leave it outside the JuliaLang org is if we are crunched for GitHub action CI minutes.

Public repos have unlimited minutes, but even then, it only takes ~1.5min a day normally, and ~30 minutes a day in the week following a (pre-)release.

ViralBShah:
@SaschaMann can you figure out where it should live org-wise given the security concerns? Maybe JuliaCI?

DilumAluthge:
JuliaLang and JuliaCI both have quite a few members. Maybe JuliaPackaging?

We could also make a dedicated org with just a few members and put the script there. E.g. JuliaDownloads or JuliaDeployment.

SaschaMann:
> can you figure out where it should live org-wise given the security concerns? Maybe JuliaCI?

> JuliaLang and JuliaCI both have quite a few members.

Members aren't an issue on their own. It depends if the default permission is `read` or `write`, and how many owners/admins there are. This is not something I can figure out from the outside.

If an org has 500 members, but only 5 of them have write access to all repos, it would be easy enough to create a team and only give access to the versions.json-repo to that team (this is how Exercism handles permissions for ~230 people and 300 repos that are continuously deployed with similar security considerations).

`julia-actions` would work for me, since the same security concerns about access apply there and access is limited by default, but if people want to use the file for non-GHA tools, they might not like that approach.

> We could also make a dedicated org with just a few members and put the script there. E.g. JuliaDownloads or JuliaDeployment.

That could work as well. I don't see a huge benefit over a personal repo but I wouldn't mind this option.

DilumAluthge:
In JuliaCI, members have write access by default, so that won't work.

ViralBShah:
In general everything on that downloads page is meant to be officially supported by the community, and hence should have continuity, more than one person knowing what's going on, fixing things while someone else is unavailable etc. So, org makes sense from signalling and group management perspective. 

Practically speaking not much will change. Let me also ping @StefanKarpinski on this.

SaschaMann:
(I added something to my comment after you responded, so extracting it for clarity)

Another option would be uploading the schema to S3 as well. Then there's almost zero reason for anyone to ever access the repo it's built in. The docs, if there will be any extensive ones, could live in a separate repo.

ViralBShah:
Uploading to S3 makes sense, and then it lives with all the other downloads as well. And there is an established mechanism to deal with secrets and such for S3 uploads.

staticfloat:
Yeah, let's just upload the schema to S3.  Sascha, if you don't already have credentials to that S3 directory, I'll generate credentials for you so that you can maintain it.

SaschaMann:
>  Sascha, if you don't already have credentials to that S3 directory, I'll generate credentials for you so that you can maintain it.

I have the credentials to update `versions.json` (iirc you restricted it to that file). Can you extend its permissions to allow uploading the schema file?

I'll setup the CI workflow to deploy it tomorrow.

---

> In general everything on that downloads page is meant to be officially supported by the community, and hence should have continuity, more than one person knowing what's going on, fixing things while someone else is unavailable etc. So, org makes sense from signalling and group management perspective.

Makes sense. Do you think from that perspective it's enough if the docs live in an org, and the "deploy repo" is still in a personal repo? That would save the overhead of a single-repo org. Though I'm still happy to create one or move it if you prefer that.

If so, could you create a repo in JuliaCI for the docs and add me to it? Then I can move all relevant documentation - most of which hasn't been written yet - there.

staticfloat:
> Could you extend its permissions to allow uploading the schema file to `versions-schema.json`?

Done.

SaschaMann:
Thanks! It's deployed: https://julialang-s3.julialang.org/bin/versions-schema.json (via https://github.com/SaschaMann/Julia-versions.json/commit/797e1623674b48f55c2701ace022161371e5f153)

SaschaMann:
Quick follow-up to the discussion above: [GitHub announced a feature today](https://github.blog/2020-12-08-new-from-universe-2020-dark-mode-github-sponsors-for-companies-and-more/#new-to-actions) that would possibly allowing moving the script to JuliaLang/Julia after all. It would require someone to approve a deploy, so it wouldn't be fully automated anymore, but it may be worth it if that means it can be moved. I'll keep an eye on it and try if it out once the feature has been released.

PR #1102, opened 2020-12-02
Explain the vertical axis of the benchmark timing graph

rdeits:
As discussed in https://discourse.julialang.org/t/julia-micro-benchmarks-graph/50005


ViralBShah:
I've been meaning to also add a bunch of other benchmarks to this page.

PR #1103, opened 2020-12-03
downloads: Update link to JSON schema

SaschaMann:
See https://github.com/JuliaLang/www.julialang.org/pull/1101

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -354,4 +354,4 @@ The platforms currently supported by Julia are listed below. They are divided in
 
 ## JSON release feed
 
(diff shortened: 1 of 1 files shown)


ViralBShah:
Should it only have released versions - and not RCs and betas and such? Or is the idea to have every single released binary in such a list?

DilumAluthge:
> Should it only have released versions - and not RCs and betas and such? Or is the idea to have every single released binary in such a list?

I think it's the latter. That way, if e.g. you want to test a RC on GitHub Actions CI, you can do so.

PR #1109, opened 2020-12-21
Fix the welcome message

DilumAluthge:
When GitHub Actions runs on the `pull_request` event on a pull request from a fork, the `GITHUB_TOKEN` has read-only permissions, and thus it will not be able to post a comment. See e.g. https://github.com/JuliaLang/www.julialang.org/pull/1108, where the welcome message failed on commit `60bff9b`, and no comment was posted.

Instead, we should use the `pull_request_target` event, which should have the correct permissions even when the pull request is from a fork.

@SaschaMann Can you take a look at this PR and confirm that I am using `pull_request_target` correctly?

### DIFF
--- a/.github/workflows/greet.yml
+++ b/.github/workflows/greet.yml
@@ -2,7 +2,7 @@ name: Welcome first time contributors
 on:
   issues:
     types: [opened]
-  pull_request:
+  pull_request_target:
     branches:
       - master
 jobs:


logankilpatrick (approved):
Based on my reading here: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target this looks right. Good catch!

logankilpatrick:
@DilumAluthge Feel free to merge this, looks like it will fix it to me.

SaschaMann:
> Can you take a look at this PR and confirm that I am using pull_request_target correctly?

Yes, you are 👍

PR #1110, opened 2020-12-26
Add in auto platform detection on downloads page.

logankilpatrick:


logankilpatrick:
TODO: 
- [ ] Test on mobile platforms
- [ ] Test on Linux
- [ ] Get design feedback
- [ ] Figure out what other minimum info we want on this page

logankilpatrick:
@tlienart does anything here look glaringly wrong? This works locally for me but is not detecting the platform in the build preview now that it's deployed.

tlienart:
I'm not familiar with this so can't help much :) but if it doesn't work in the preview I doubt it'll work live 😕

ViralBShah:
Should we close this?

logankilpatrick:
Sure, I'll come back to it when I have more time.

PallHaraldsson:
I like it, but shouldn't we be emphasizing juliaup (soon)? Is it worth it do do much work for the download page then? I use juliaup on Linux and it seems stable...

PallHaraldsson:
```
ubuntu@ubuntu:~$ julia
Command 'julia' not found, but can be installed with:
sudo snap install julia
```

That's aweful advice, FYI, better:
Command 'julia' not found, but can be installed with:
sudo snap install juliaup
```

or take over the julia package to switch to juliaup, then it could be snap still, because the downloaded julia wouldn't be.

LilithHafner:
This has gone stale enough that it may be worth starting a new PR. Further, once we start recommending juliaup, the explicit download links will become targeted toward folks who opt out of juliaup's automation and those users may also prefer to see a full list of available downloads.

PR #1093, opened 2020-11-28
WIP: add blog post on new latency-reduction tools

timholy:
This focuses on the new Core.Compiler.Timings inference-timing tools, and
the utilities in SnoopCompile for analyzing the results (`@snoopi_deep` and friends). These tools were
introduced by Nathan Daly, who is a co-author of the post. CC @NHDaly

This WIP in part because it depends on quite a few outstanding PRs:

- [x] https://github.com/JuliaLang/julia/pull/38566
- [x] https://github.com/JuliaLang/julia/pull/38596 will force me to re-run all the examples with nicer printing
- [x] https://github.com/timholy/SnoopCompile.jl/pull/152
- [ ] https://github.com/timholy/SnoopCompile.jl/pull/157
- [ ] https://github.com/timholy/SnoopCompile.jl/pull/159
- [ ] https://github.com/JuliaCollections/AbstractTrees.jl/pull/65

Nevertheless it seemed time to post this so that @NHDaly, among others, can collaborate on the writing and so that the DataFrames developers can get a sense for the overall context.

PR #1111, opened 2020-12-26
Add a blog post on precompilation

timholy:
I'm preparing new material on how to reduce latency and improve precompilation in packages, but I realized I would benefit from some introductory material before discussing the tools (e.g., before returning to #1093). Hopefully this post will introduce all the concepts I need.


timholy:
Pinging some participants in https://discourse.julialang.org/t/understanding-and-optimizing-compiler-time-just-a-bit/51759 and https://discourse.julialang.org/t/understanding-precompilation-and-its-limitations-reducing-latency/48294: @tpapp, @chriselrod, @jkrumbiegel, @joa-quim.

tlienart:
Small Franklin note which you could ignore: if you use a triple backtick with a `!` then the code in there gets executed and the output is shown on the page, the interest in doing this is that if ever the output changes slightly, it will be updated automatically. So for instance:

````
```!
c64 = [1.0]
```
````

--> 

Screenshot 2020-12-27 at 12 08 21

**Comment by timholy** [comment]
Thanks for the tip! I'm not sure I'll use it due to changes in formatting and the fact that it seems to need an isolated statement, but I will keep that in mind.

I just pushed an update that may hint at something I'm trying but failing to do: I'd be interested in exploring having "boxes" for different types of content: asides, exercises/quizzes, and perhaps other topics. As you'll see, I pushed a .css file as well as my markdown, but this is me just imitating things I've found on the internet without actually understanding what I'm doing. Is this easy or hard? If the former, any pointers?

**Comment by tlienart** [comment]
You can do this with `@@` in Franklin this will surround content in a div block of a given class (or classes): e.g.

```
@@note
content here
@@
```

Will inject the content in a `<div class="note">...</div>". For several classes use `@@class1,class2,class3 content @@`. I _think_ that should be all you need but don't hesitate to ping me from here :)

File: blog/2020/12/precompile_tutorial.md [hunk 1]
```diff
@@ -0,0 +1,586 @@
...
+## Sources of latency, and reducing it with `precompile`
+
+Most of Julia's latency is due to *code loading* and *compilation*. Julia's dynamic nature also makes it vulnerable to *invalidation* and the subsequent need to recompile previously-compiled code; this topic has been covered in a previous [blog post][invalidation], and that material will not be rehashed here.  In this series, it is assumed that invalidations are not a dominant source of latency.  (You do not need to read the previous blog post to understand this one.)
+
+In very rough terms, `using SomePkg` loads the method definitions, after which calling `SomePkg.f(args...)` forces `SomePkg.f` to be compiled (if it hasn't been already) for the specific types in `args...`.  The primary focus of this series is to explore the opportunity to reduce the cost of compilation.  We'll focus a lot on *precompilation*,
+
+```
+julia> using SomePkg
+[ Info: Precompiling SomePkg [12345678-abcd-9876-efab-1234abcd5e6f]
+```
```
ianshmean, line 26: worth adding something like this?

---
or as of 1.6, at the end of `pkg>` actions
```
(@1.6) pkg> add SomePkg
    Updating registry at `~/.julia/registries/General`
   Resolving package versions...
    Updating `~/.julia/environments/v1.6/Project.toml`
  [12345678] + SomePkg v0.1.0
    Updating `~/.julia/environments/v1.6/Manifest.toml`
  [12345678] + SomePkg v0.1.0
  [abcdefgh] + SomeDep v0.1.0
 Precompilng project...
  Progress [========================================>]  2/2
2 dependencies successfully precompiled in 3 seconds
```

chriselrod:
> While a lot of work has gone into Julia 1.6 to reduce latency, users and developers naturally will want more.

More work, but less latency. ;)

I think it looks good and found it informative/helpful. I have some playing around to do!

timholy:
If you're looking into fixing inference, the upcoming (on `master`) `inference_triggers` framework in SnoopCompile is the bomb :laughing:.

tpapp:
This is very informative, thanks for writing it up! As I am still digesting this, I don't have much to add yet, except that it is unclear how to get a method instance object from the `@snoopi_deep` and similar output that will come in later installments.

blog/2020/12/precompile_tutorial.md
timholy: I added a brief version of that.

timholy:
Boxes are working great, thanks @tlienart!

Barring suggested changes, I'm currently targeting this for a Jan 5 release (just to give people a few days to get back in the swing).

NHDaly (commented):
🌟 This was great! 🌟 

This was _such_ a useful introduction to many of the topic at play, and I think will do a good job of helping people to understand what's going on under the hood. Thanks for a really nice read! :) :)

Thank you for writing this all up so clearly and explaining it so well!

blog/2021/01/precompile_tutorial.md
```diff
@@ -0,0 +1,678 @@
...
+```
+
+might be *dramatically* more complex if you've loaded and used large packages that do a lot of computation.
+@@
+
+@@note
+**Box 3** Generally, the set of backedges is a graph, not a tree: in real code, it's possible for `f` to call itself (e.g., `fibonacci(n) = fibonacci(n-1) + fibonacci(n-2)`), or for `f` to call `g` which calls `f`.
+When following backedges, MethodAnalysis omits `MethodInstances` that appeared previously, thus performing a "search" of the graph.  The results of this search pattern can be visualized as a tree.
+
+Type inference behaves similarly: it caches its results, and thus infers each `MethodInstance` only once. (One wrinkle is [constant propagation](https://en.wikipedia.org/wiki/Constant_folding), which can cause the same `MethodInstance` to be re-inferred for different constant values.)  As a consequence, inference also performs a depth-first search of the call graph.
```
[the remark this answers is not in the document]
timholy: If I were making other changes, I'd do this, but on its own but I think the question of persistence of inference's cache is a bit more detail than readers need.

timholy:
Awesome! Let's write the next one together? I'd like to have one/two on the GUIs, the flamegraph and my (still morphing but mostly stable) "profile-guided despecialization" (PGDS) GUI.

The PGDS stuff gets into complex territory that I'm still figuring out, but probably best just to get out what we know and let the community help figure it out...

PR #1116, opened 2021-01-08
Add download links for v1.6.0-beta1

ararslan:
Including the return of Linux PowerPC. Currently waiting on macOS notarization.


ararslan:
macOS has been notarized and Windows signing has been verified.

PR #1133, opened 2021-01-30
Tweak macOS installation steps

travigd:
Was helping someone debug installing on a Mac and thought the docs could be a bit better. They were a bit confused about the symlink step. Overall, I think we should just be as radically simple and to-the-point as possible here.

Also for uninstallation, can we just say to nuke `~/.julia`? For most users, it probably doesn't make sense to tell them to delete specific subdirectories within.


logankilpatrick:
Thank you for this!

PR #1136, opened 2021-02-04
Update Manifest

IanButterworth:
Franklin was quite a few patches behind "0.10.12" -> "0.10.28"


DilumAluthge (approved):
This should be good to merge, since all of the updates are nonbreaking.

DilumAluthge:
Should be good to go. If for some reason something breaks, we can revert.

PR #1145, opened 2021-02-04
Remove 'soc' directory

aviks:
Seems like this directory is no longer used? It's presence confused me when trying to make changes.


logankilpatrick:
It was there as a re-direct since we initially changed the directories but at this point I think we can nuke it

PR #1159, opened 2021-02-04
Update science.md

mforets:
Removing the proposal because the project has been completed.


logankilpatrick:
Can you remove this file and it's reference from the project page as well?

logankilpatrick:
Thank you!!

PR #1160, opened 2021-02-04
Update some Tooling projects

pfitzseb:


logankilpatrick:
Hey @pfitzseb, google mentioned that they didn't want generic projects that relate to open issue. Can you add a few specific projects here and then leave a section that also says you are "open to other projects" and students can check out the issue tracker for inspiration?

pfitzseb:
Yes, but that'll take a bit more time than just removing the Juno projects. Let's merge this as-is (or without the generic VSCode section) and we'll open another PR with more specific suggestions?

logankilpatrick:
Sounds good!

PR #1161, opened 2021-02-04
Added two time series projects for GSOC with corrections

ababii:
Hi Logan,

Thanks for your feedback on the previous version. I made the changes accordingly. In my view, Julia lacks popular, scalable, and automated time series forecasting functionality. For example, one of the most popular statistical packages in R is the 'forecast' package with 8-12k daily downloads (actually more popular than ML packages like xgboost, randomforest, glmnet, or keras). Adding the time series functionality has the potential to increase significantly the popularity of Julia in academic and business communities. The speed of Julia is a huge advantage here. For instance, I'm working on a project that involves forecasting hundreds of thousands of time series of ticket sales on different routes for a major airline. It takes quite some time just to clean this dataset in R. Julia can do the same job in minutes.

I listed the functions that in my view are important to implement, including scalable seasonal ARIMA models, intermittent demand forecasting, and MIDAS regressions. But I also leave it for students to suggest other popular functions for time series forecasting from other platforms. Please let me know if I can make further improvements to projects and/or to add clarity. I have participated in the 2012 GSOC with R Project, but it was a while ago.

Best wishes,
Andrii

### DIFF
--- a/jsoc/gsoc/pythia.md
+++ b/jsoc/gsoc/pythia.md
@@ -1,0 +1,32 @@
+# Pythia – Summer of Code
+## Machine Learning Time Series Regression
+
+[Pythia](https://github.com/ababii/Pythia.jl) is a package for scalable machine learning time series forecasting and nowcasting in Julia.
+
+The project mentor is [Andrii Babii](https://ababii.bitbucket.io/).
+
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Look good, thanks!

PR #1165, opened 2021-02-06
Stats. e.g. over 5000 packages

PallHaraldsson:
I projected and over 25 million downloads should be true, but I have no proof the less than 800,000 additional downloads for January happened.

Over 5000 packages is for sure true with them 5001. Number 5000 "New package: Proquint v0.1.0 (#29279)", actually a neat idea, and then "DLPipelines v0.1.0 (#29280)".

### DIFF
--- a/index.html
+++ b/index.html
@@ -364,7 +364,7 @@ end
 
       <br>
 
-      <p>Julia has been downloaded over 24 million times and the Julia community has registered <a href="https://juliahub.com">over 5,000 Julia packages</a> for community use.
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Well wait for JC to update the stats on their site sometime this month and then merge (if over 25 mil)

PallHaraldsson:
We could update to over 24 million, to begin with, as that's for sure right, and a huge jump. And 5000 packages.

ViralBShah:
I know we have hit the 25M - so just updating that here. Let's not update until JuliaCon now...

PR #1169, opened 2021-02-08
Remove UCX topic from HPC GSoC

jpsamaroo:
@vchuravy has already implemented functional UCX support


logankilpatrick:
Thank you!

PR #1173, opened 2021-02-09
add two publications and update packages links

kalmarek:
I'm not sure what is the order of articles; in 2019 they seem to follow alphabetic order (by title); there seems to be no order in 2020


logankilpatrick:
Thanks! Yeah, we need to sort out the ordering.

PR #1178, opened 2021-02-13
Update platform.md macos homebrew syntax changes

mecanogrh:
brew cask <command> was deprecated in favor of brew <command> --cask in Homebrew 2.6.0. Starting with 2.7.0 they have been removed in favor of brew <command>


logankilpatrick:
Thank you for this! Good catch.

Issue #1139, opened 2021-02-04
Update ML GSoC Projects 


logankilpatrick:
Hey @maleadt @MikeInnes @DhairyaLGandhi @avik-pal @aviks @findmyway @staticfloat @ChrisRackauckas we need to get the ML projects here:
https://julialang.org/jsoc/gsoc/flux/ updated for GSoC 2021 asap.

There are new requirements for the projects found here: https://google.github.io/gsocguides/mentor/defining-a-project-ideas-list including adding an (easy, medium, hard) rating for each project. Please read through that link and updated the projects as needed.

If you are no long able to mentor the project or the project has been completed, please open a PR to remove it. Thanks all!

aviks:
@DhairyaLGandhi maybe we should remove the model-zoo examples project?

avik-pal:
For the 3D Deep Learning Project: I think we can remove it as a single project. Most of the work was completed by Nirmal during the last GSoC. Some of them are left but would be better if integrated with some of / broken off into the other projects:

1. Differentiable Rendering **[HARD]**: We kind of decided that it is best not to integrate it in Flux3D. Rather any of the differentiable renderers developed can use the primitives provided by Flux3D.
    * **Expected Outcome:** SoftRasterizer/DiB-R based works. We already have RayTracer.jl which is motivated by OpenDR. (Of course, if someone wants to implement NERF - like models they are most welcome to submit a proposal). We would ideally target at least 2 of these models.
    * **Skills:** GPU Programming, Deep Learning, (deep) familiarity with the literature, familiarity with defining (a lot of) Custom Adjoints

2. Core Development **[MEDIUM]**:
    * **Expected Outcome:**
         * Some of the functions require custom adjoints for speedup
         * Functions require GPU kernels. Some of these are of common interest to the community like -- knn, etc.
         * Move DataLoaders and similar functions out of Flux3D and use DataLoaders.jl and similar utilities for these functions
         * Benchmarking with Tensorflow Graphics and Pytorch3D. We already have the scripts for kaolin, need to extend that.
         * Most of these problems are listed as issues in the main repo.
    * **Skills:** GPU Programming, Deep Learning, familiarity with defining (a lot of) Custom Adjoints

[...]

PR #1182, opened 2021-02-16
Update FastAI.jl GSoC

darsnack:
Updates FastAI.jl GSoC project in response to #1139. My only concern with this project is timeline (i.e. will we have enough work for the student when summer rolls around). Previously we were focused on the fast.ai book, so if we can get Jeremy's permission (it's copyrighted) then recreating the book would become part of the scope. In which case, I no longer have timeline concerns cause there will be plenty of work.

cc @lorenzoh


logankilpatrick (approved):

darsnack:
I just want @lorenzoh's feedback before we merge

darsnack (commented):
This is good to go on our end

logankilpatrick:
Thank you all!

lorenzoh:
Thanks @darsnack!

PR #1198, opened 2021-02-24
adding project of optimisation of MLJ

darenasc:
* Adding an entry to the list of MLJ projects "Diagnose and exploit opportunities for speeding up common MLJ workflows."


logankilpatrick:
Looks great, thank you!

PR #1203, opened 2021-03-01
Remove one occurrence of JuliaText in GitHub groups

gdalle:
JuliaText was mentioned twice, I kept the one in the DataScience category


logankilpatrick:
Good catch, thanks!

Issue #1207, opened 2021-03-09
Error in Readme to run project locally


Bhavik-Ardeshna:

Bhavik-Ardeshna:
Clone the repository and `cd` to it. Start julia with `julia --project` and do

```julia
julia> using Pkg; Pkg.instantiate()

julia> using Franklin

julia> newsite("mySite", template="pure-sm")

julia> serve()
→ Initial full pass...
→ evaluating code [...] (learning/code-examples.md)
→ Starting the server...
✓ LiveServer listening on https://localhost:8000/ ...
  (use CTRL+C to shut down)
```

Bhavik-Ardeshna:
```
julia> serve()
 Activating environment at `~/Desktop/Julia/www.julialang.org/Project.toml`
→ Initial full pass...
┌ Franklin Warning: in <jsoc/gsoc/graphs.md>
│ A h-function call '{{fill title}}' (or '{{title}}') has argument 'title' which
│ doesn't match a page variable available in the current scope. It might have
│ been misspelled or should be defined via a '@def title = ...' either
│ locally on the page or globally in 'config.md'.
│ 
│ Relevant pointers:
│ - page variables and {{...}} blocks: https://franklinjl.org/syntax/page-variables/
└
┌ Franklin Warning: in <MLJ.md>
│ Encountered an issue processing 'MLJ.md' in rg/content/jsoc/gsoc.
│ Verify, then re-start the Franklin server.
│ The error is displayed below:
│ Franklin.HTMLBlockError("I found a HBlock that did not match anything, verify '{{< get_param title >}}'")
└
ERROR: Franklin.HTMLBlockError("I found a HBlock that did not match anything, verify '{{< get_param title >}}'")
Stacktrace:
 [1] process_file(::Symbol, ::Pair{String,String}, ::String, ::Vararg{Any,N} where N) at /home/user/.julia/packages/Franklin/X69Nd/src/manager/file_utils.jl:88
 [2] fd_fullpass(::NamedTuple{(:other, :infra, :md, :html, :literate),NTuple{5,Dict{Pair{String,String},Float64}}}) at /home/user/.julia/packages/Franklin/X69Nd/src/manager/franklin.jl:240
[...]

PR #1209, opened 2021-03-09
Remove the graphs.md file from gsoc

KristofferC:
This file is empty and causes warnings when serving the docs.

```
┌ Franklin Warning: in <jsoc/gsoc/graphs.md>
│ A h-function call '{{fill title}}' (or '{{title}}') has argument 'title' which
│ doesn't match a page variable available in the current scope. It might have
│ been misspelled or should be defined via a '@def title = ...' either
│ locally on the page or globally in 'config.md'.
│ 
│ Relevant pointers:
│ - page variables and {{...}} blocks: https://franklinjl.org/syntax/page-variables/
└
```

I don't think it is referenced from anywhere. Ref #1207 (but doesn't fix the error).


logankilpatrick:
Thank you! I was going to leave it but if it's giving warnings we can delete it.

PR #1210, opened 2021-03-09
Remove graph.md and Resolve some typo in MLJ.md

Bhavik-Ardeshna:
#1207 #1209


logankilpatrick:
Thank you for this PR but looks like that page you edited should not be there in the first place. Opened an issue to get it resolved.

PR #1215, opened 2021-03-10
JSoC: Rename "Tooling" to "Web and IDE Development" to make it sound …

fonsp:
…more exciting


fonsp:
Not renaming the page file (`tooling.md`) so that we don't break shared and bookmarked URLs.

logankilpatrick:
Great idea, thanks!

PR #1135, opened 2021-02-03
blog post: Julia 1.6 Highlights

StefanKarpinski:

PR #1236, opened 2021-03-25
Set stable download version to 1.6.0

ararslan:
Also add PowerPC to the stable version section


ararslan:
Should this and #1135 be merged together, perhaps?

logankilpatrick:
Makes sense to me

KristofferC (approved):

Issue #1241, opened 2021-03-25
dark mode julia logo


ViralBShah:
I use a dark mode plugin in my browser, and we have larger issue around respecting dark mode: #775.

While much of the website works fine with the dark mode plugin, it is only the julia logo that blends into the background. How do we use the white julia logo (which we have in the logo repo) when the browser is in dark mode?
<img width="355" alt="Screen Shot 2021-03-25 at 3 34 24 PM" src="https://user-images.githubusercontent.com/744411/112532970-d3e07e00-8d7f-11eb-9d5e-e05ebe62e73b.png">

logankilpatrick:
My understanding is that we would need to implement something in our code (like a dark mode Library or special non-existing Franklin syntax to handle dark mode) or use @cormullion 's black logo with a white outline to make it visible. The latter is probably easier.

ViralBShah:
@tlienart Is this something Franklin should support?

tlienart:
This is not on Franklin's side, as far as I'm aware this should be addressed by either

1. some CSS (suggested by Keno in the other issue)
2. some javascript

Either way it's not on the static side, this should be done when the user loads the page. I do not know precisely which CSS needs to be added so that it works with the different ways user can request dark mode (if it's a browser extension that needs to be detected, that's going to be a mess which requires javascript, if it's a toggle on the website, that also needs javascript). 

I'm not familiar with dark mode stuff so if someone wants to experiment with it and figure out what needs to be done to respect user preferences and it turns out that it does need some specific support on Franklin's side, I'll be glad to lend a hand there

cormullion:
As a temporary measure we could do some white-lining - I'll take a look tomorrow.

PR #1244, opened 2021-03-26
add white line around logo #1241

cormullion:
I put a small white stroke around the Julia logo, which should be invisible for most people, but will show up for those who use dark-mode plugins in their browser.

Artist's impression (I don't know what it would look like with an actual dark-mode plug-in):

Screenshot 2021-03-26 at 11 14 30


logankilpatrick:
Looks great, @ViralBShah can you give it a shot?

ViralBShah:
Not showing up for me yet. Maybe taking some time to propagate...

cormullion:
How the live version looks on my iPad with a wacky dark-mode browser setting.

PNG image 3

Perhaps this will do as a temporary fix.

PR #1251, opened 2021-04-01
Use correct name scheme for Mac release

ageorgou:
Making the reported name match the actual file.

I tried replacing the instances of `<version>` in the instructions with `{{stable_release}}` and `{{stable_release_short}}`, but it doesn't render correctly inside backticks. Is it okay to use a `<code>` block as with the Windows path above?

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -59,7 +59,7 @@ Uninstallation is preferably performed by using the Windows uninstaller. The dir
 
 ## macOS
 
-On macOS, a `julia-<version>-mac64.dmg` file is provided, which contains `Julia-<version>.app`. Installation is the same as any other Mac software: drag the `Julia-<version>.app` to Applications Folder's Shortcut. The Julia download runs on macOS 10.9 Mavericks and later releases. You can build from source for macOS 10.6 Snow Leopard (possibly earlier versions as well) and 32-bit but neither are fully supported.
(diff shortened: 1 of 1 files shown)


downloads/platform.md
```diff
@@ -59,7 +59,7 @@ Uninstallation is preferably performed by using the Windows uninstaller. The dir
 
 ## macOS
 
-On macOS, a `Julia-<version>.dmg` file is provided, which contains `Julia-<version>.app`. Installation is the same as any other Mac software: drag the `Julia-<version>.app` to Applications Folder's Shortcut. The Julia download runs on macOS 10.9 Mavericks and later releases. You can build from source for macOS 10.6 Snow Leopard (possibly earlier versions as well) and 32-bit but neither are fully supported.
+On macOS, a ~~~<code>julia-{{stable_release}}-mac64.dmg</code>~~~ file is provided, which contains ~~~<code>Julia-{{stable_release_short}}.app</code>~~~. Installation is the same as any other Mac software: drag the ~~~<code>Julia-{{stable_release_short}}.app</code>~~~ to Applications Folder's Shortcut. The Julia download runs on macOS 10.9 Mavericks and later releases. You can build from source for macOS 10.6 Snow Leopard (possibly earlier versions as well) and 32-bit but neither are fully supported.
```
fredrikekre, line 62:
```suggestion
On macOS, a ~~~<code>Julia-{{stable_release}}-mac64.dmg</code>~~~ file is provided, which contains ~~~<code>Julia-{{stable_release_short}}.app</code>~~~. Installation is the same as any other Mac software: drag the ~~~<code>Julia-{{stable_release_short}}.app</code>~~~ to Applications Folder's Shortcut. The Julia download runs on macOS 10.9 Mavericks and later releases. You can build from source for macOS 10.6 Snow Leopard (possibly earlier versions as well) and 32-bit but neither are fully supported.
```

giordano: As confusing as it is, dhe `.dmg` is lowercase, the `.app` is uppercase: https://julialang-s3.julialang.org/bin/mac/x64/1.6/julia-1.6.0-mac64.dmg

fredrikekre: Ah, so it was wrong before.

ageorgou: Yes, the capitalisation and missing `-mac64` are what I tried to fix, then thought it should also use the version variables.

fredrikekre: Yea, thanks.

PR #1258, opened 2021-04-05
Anonymize IP addresses 

logankilpatrick:
Add synonymous IP addresses: https://developers.google.com/analytics/devguides/collection/gtagjs/ip-anonymization

Can someone double check that I have this  right?


aviks (commented):
Google help seems to suggest that you should do it like this: 

`ga('set', 'anonymizeIp', true);`

https://developers.google.com/analytics/devguides/collection/analyticsjs/ip-anonymization

ViralBShah:
Can someone explain what we might lose (statistics-wise) if we do this?

logankilpatrick:
Closing for now, I don't remember the motivation for this.

PR #1259, opened 2021-04-06
GSoC: Link to documenter.md

mortenpi:
It looks like a link to the new `documenter.md` page was missed in f05366c636639e7a35c6723846ede5a5c86d03d7.


logankilpatrick:
Good catch, thanks!

PR #1261, opened 2021-04-08
Add single page for all GSoC projects to improve searchability

logankilpatrick:

PR #1266, opened 2021-04-11
Add all projects page

tlienart:
Context: #1261 ; I think it's a bit ugly but if it's just for the purpose of being able to search a page for keywords, it does the job (it gets H titles wrong etc etc though but that core purpose is met).

cc @logankilpatrick 

Note: ignore the RSS stuff, that's from a recent Franklin release and should not affect julialang.org in any meaningful way.


logankilpatrick:
Wonderful, this does the job! Thank you very much. Side note, was I missing something with the inclusion of files like I was trying before? Or would have not have worked?

tlienart:
No it's a bit tricky and something I should fix in Franklin, the inclusions are relative to the __site folder, not the projects folder, so here you were trying to include .md files which were not in the __site folder.

There's two ways (currently) to do this: 1. Is to use the exact path (which is annoying) and 2. Put stuff in _assets as then they will be copied over into the __site folder as they are

I know this is awkward; the initial raison d'être for the command dates back a bit and made several assumptions as to why someone would use this, I need to revisit this and adjust it, there's already an open issue for it I think :)

logankilpatrick:
Makes sense :), thank you again!

PR #1267, opened 2021-04-11
Fix typo for NumFocus

kimlaberinto:


logankilpatrick:
I need to sleep more :) thank you!

Issue #1268, opened 2021-04-12
Broken links to https://docs.julialang.org/en/latest/manual/...


rikhuijzer:
It appears that https://docs.julialang.org/ doesn't have a `latest` tag any more. Still, this repository contains 10 links to that version of the docs, which all show a 404 when opening the link:

```
$ rg 'latest/manual'
www.julialang.org/blog/2018/07/iterators-in-julia-0.7.md
412:[^defaults]: In Julia, this actually defines two methods of `iterate`, as described in the [Julia docs](https://docs.julialang.org/en/latest/manual/methods/#Note-on-Optional-and-keyword-Arguments-1).
416:[^slurpsplat]: Slurping refers to  how using `args...` in a function definition "slurps" up the trailing arguments, and splatting is the inverse operation. The [Julia docs](https://docs.julialang.org/en/latest/manual/functions/#Varargs-Functions-1) say more on this.

www.julialang.org/blog/2013/04/distributed-numerical-optimization.md
133:macro creates a [task](https://docs.julialang.org/en/latest/manual/control-flow/#man-tasks)

www.julialang.org/blog/2018/06/missing.md
234:See the [manual](https://docs.julialang.org/en/latest/manual/missing/) for more details

www.julialang.org/blog/2018/05/extensible-broadcast-fusion.md
227:[Broadcasting]: https://docs.julialang.org/en/latest/manual/arrays/#Broadcasting-1
233:[documented and available to packages]: https://docs.julialang.org/en/latest/manual/interfaces/#man-interfaces-broadcasting-1

www.julialang.org/blog/2017/01/moredots.md
537:lower case, and then we replace any sequence of whitespace (the [regular expression](https://docs.julialang.org/en/latest/manual/strings.html#Regular-Expressions-1)

www.julialang.org/blog/2018/03/pifonts.md
19:Julia embraces the Unicode standard [enthusiastically](https://docs.julialang.org/en/latest/manual/unicode-input/), so it's very easy to use the appropriate Greek (and other Unicode) letters in your code. In the REPL, for example, type `\pi TAB` to insert the Unicode character `U+03C0`:

www.julialang.org/blog/2020/08/invalidations.md
445:[performance tips]: https://docs.julialang.org/en/latest/manual/performance-tips/

www.julialang.org/blog/2016/02/iteration.md
32:[manual](https://docs.julialang.org/en/latest/manual/arrays/#Iteration-1).
[...]

PR #1269, opened 2021-04-12
Fix broken links to latest docs

rikhuijzer:
Fixes https://github.com/JuliaLang/www.julialang.org/issues/1268.

In the linked issue, Fredrik Ekre has a good point:

> I guess the links that have corresponding sections in the current manual can be updated, but I guess in general blogposts will always be outdated after some time.

So, that is why I have fixed the links only for blog posts published after 2017. That means that posts which are still actual like [invalidations](https://julialang.org/blog/2020/08/invalidations/) now have their link fixed. For older posts, I guess that the reader won't be surprised by broken links.


logankilpatrick:
Thanks for this!

PR #1285, opened 2021-04-25
LBT blog post

ViralBShah:
A very early start to the LBT blog post. Needs more work, but I figured I would get it started.

### DIFF
--- a/blog/2021/05/julia-1.7-blas-lapack-trampoline.md
+++ b/blog/2021/05/julia-1.7-blas-lapack-trampoline.md
@@ -1,0 +1,23 @@
+@def title = "BLAS and LAPACK trampolines in Julia 1.7"
+@def authors = "Elliot Saba, Viral Shah"
+@def published = "5 May 2021"
+@def rss_pubdate = Date(2021, 5, 5)
+@def rss = """"BLAS and LAPACK trampolines in Julia 1.7"""
+
+The BLAS and LAPACK libraries have been a corner stone of technical computing for at least 20 years now, if not more! The APIs in BLAS and LAPACK have been so successful, that hardware vendors provide their own implementations of these APIs in order to extract the highest performance from their hardware. While Julia has always supported using your own BLAS, it has historically required either building Julia from source, or rebuilding the system image (as was the case with MKL.jl).
+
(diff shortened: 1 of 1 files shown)


ViralBShah:
Possible things to add:

- [ ] List of different BLAS and LAPACK we already support: OpenBLAS, MKL, Accelerate
- [ ] LBT features and auto-detection of LP64 / ILP64 / etc.
- [ ] Wishlist to support BLIS and maybe even NVBLAS.
- [ ] Roadmap to link various Yggdrasil libraries against LBT. Perhaps these can even be made to work with older Julia releases - just have to initialize LBT (which is currently done in the LinearAlgebra module).

giordano:
For usage outside of Julia, specifically in HPC environments, it may be worth mentioning that libblastrampoline is [available in Spack](https://github.com/spack/spack/blob/0e43f5299152271adee6de144562ae08087cdc58/var/spack/repos/builtin/packages/libblastrampoline/package.py)

giordano:
I just created two _simple_ (understatement) packages:

* [`OpenBLAS32.jl`](https://github.com/giordano/OpenBLAS32.jl)
* [`OpenBLASHighCoreCount.jl`](https://github.com/giordano/OpenBLASHighCoreCount.jl)

to forward libblastrampoline to the corresponding JLL packages.

I'm not sure it's worth registering them --- I don't think any package should use them directly, only user code ---, but I figured I'd mention them here for feedback and because they may be relevant for the post.

ViralBShah:
I was thinking we just update various things that need to link to BLAS to link directly to LBT and put a julia_version 1.7 requirement in Yggdrasil.

giordano:
> I was thinking we just update various things that need to link to BLAS to link directly to LBT and put a julia_version 1.7 requirement in Yggdrasil.

That's part of the story, only for binary packages.  With [`OpenBLASHighCoreCount.jl`](https://github.com/giordano/OpenBLASHighCoreCount.jl) for example you can replace default OpenBLAS with one which can use up to 128 threads (default can get up to 32), much like with `MKL.jl` you can replace OpenBLAS with MKL.

ViralBShah:
Any takers for completing this post (and also adding the relevant JuliaCon videos)? @giordano @staticfloat?

PR #1288, opened 2021-05-01
add macports installation guide

queensferryme:


StefanKarpinski:
I guess the question is: is the MacPorts package for Julia functional enough that we want to officially recommend it?

queensferryme:
Well, I can't say for sure that the Julia port is or will be free of bug, but at least it has been actively maintained ([commits](https://github.com/macports/macports-ports/commits/master/lang/julia/Portfile)) and used ([stats](https://ports.macports.org/port/julia/stats?days=30&days_ago=0)). I myself have been using the Julia port for machine learning (with Flux.jl) and it has been working great.

The Julia port can be tested with `sudo port test julia` (which internally runs `make test`), and most of the test passed (except one, see https://github.com/macports/macports-ports/pull/10250#issue-593502637). So I believe it is functional enough.

Anyway, I am just making a recommendation here and it is up to you to decide whether to accept MacPorts into the official documentation.

DilumAluthge:
Does MacPorts build Julia from source, or do they just ship the official binaries from JuliaLang.org?

We often find users on Discourse that are having problems because they are not using the official binaries, problems which resolve immediately when the users switch to the official binaries.

queensferryme:
Yes, MacPorts builds Julia from source.

queensferryme:
Understandable, I will close this PR.

PR #1296, opened 2021-06-04
Update tutorials.md

semmottos:
The link goes to the pdf for the current Julia version in Github. I think there's no need to mention the Julia version at all.


logankilpatrick:
Agreed, thanks!

PR #1307, opened 2021-06-28
benchmarks: Fix formatting issue

rht:
Make sure the sentence "Note that the Julia results depicted above do not include compile time" starts in a new line.

Currently it looks like this:
2021-06-27-205604_1161x566_scrot


logankilpatrick:
Thank you! I meant to fix this before.

Issue #1235, opened 2021-03-25
Updates to the Julia CoC


logankilpatrick:
Hey @JuliaLang/stewards, 

I think it would be worth iteratively working on making the CoC more robust. In particular, I think we should update the text to include the mention of the need to make things open as well as add a diversity statement. 

I like the NumPy openness statement: 
> Be open. We invite anyone to participate in our community. We prefer to use public methods of communication for project-related messages, unless discussing something sensitive. This applies to messages for help or project-related support, too; not only is a public support request much more likely to result in an answer to a question, it also ensures that any inadvertent mistakes in answering are more easily detected and corrected.

As an OSS project, it makes sense to state this explicitly. 

We should also add a dedicated diversity section on the CoC re-affirming our commitment to making the Julia ecosystem a place where everyone is welcome. Right now, the "inclusion" statement is bundled with the "be respectful" section which I think should be split.

jiahao:
Thank you for taking the lead on this language change. Cross-referencing [a relevant Slack thread](https://julialang.slack.com/archives/C6FGJ8REC/p1616620365067300) for context.

Summary of issues discussed
- Pile-on behavior should probably be discouraged.
- A "no politics" rule is not neutral, but rather reinforces the status quo, silences the marginalized, and makes it harder to advocate for positive change.
- The existence of marginalized people (e.g. nonbinary, trans, ethnic minorities) is inherently politicised.
- Diversity and inclusion of people means it's necessary to call out behaviors and statements to the contrary. (c.f. Popper's paradox of tolerance)
- A healthy community has to balance needs for robust communication, offering support, and be willing to improve, adopt best practices, and experiment with new ideas to advance diversity, equity and inclusion goals.

StarTrooper08:
One thing came into my mind .
That whenever we make issue and after the issue is been solved it should be closed .
So that it won't make any confusion for the new contributors and also experienced contributers who want to contribute.

So what I want to ask is that can we add this point in the CoC :

### After the issue gets resolved author should closed the issue .


Please do have feedback whether this point will work.

logankilpatrick:
> One thing came into my mind .
> That whenever we make issue and after the issue is been solved it should be closed .
> So that it won't make any confusion for the new contributors and also experienced contributers who want to contribute.
> 
> So what I want to ask is that can we add this point in the CoC :
> 
> - After the issue gets resolved author should closed the issue .
> 
> 
> Please do have feedback whether this point will work.

[...]

PR #1312, opened 2021-07-03
Add "Be Open" section to CoC

logankilpatrick:
Inspired by: https://numpy.org/code-of-conduct/ and based on the conversation here: https://github.com/JuliaLang/www.julialang.org/issues/1235 

I think codifying this in the DNA of the community will help the sustainability of the ecosystem and language. CC: @JuliaLang/stewards Should we ask for public comments on this like @Keno did on the Researchers section?

### DIFF
--- a/community/standards.md
+++ b/community/standards.md
@@ -18,6 +18,10 @@ If you believe one of these standards has been violated, you can either file an 
 
 Constructive criticism and suggestions are welcome, but high-traffic forums do not generally have the bandwidth for extensive discourse. Consider writing a blog post if you feel that you have enough to say on a particular subject.
 
+### Be open
+
+As a project and community build on the principles of open source, we ask everyone to be open. Where possible, we ask that you use public forums for conversation, discussion, and when asking for help. Open discussions allow: future users to benefit from the question/discussion, potentially incorrect or incomplete answers to be more swiftly corrected, and more people to see and engage in the initial conversation. More concretely, ask for help on forums with high visibility and have design discussions in places where future contributors or users can read about why/how a decision was made.
+
 ### For Researchers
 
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
I don't know, this seems like move of an aspirational value than something that we should require of people. What about people who are introverted and just want to ask some question and get an answer with as little fuss as possible? I would say that they should not ask specific people directly unless those people have advertised that this is ok to do, but I don't see why we should force people to "be open".

StefanKarpinski:
Note that the NumPy version is a statement about what the project's values are and what the project strives to be like — it does not tell other people that they need to be open.

logankilpatrick:
Is there a way you think this could be phrased to allow flexibility but still re-enforce the idea that we want things to be open? In my opinion, there are already discussions happening on places like Slack where the people choose to talk there not because they are introverted but because it is easy (this could be false, just my narrow perspective). I tried to leave things open ended by saying "Where possible, we ask that"

Kolaru:
In its current formulation the new paragraph feels rather intrusive to me. It sounds to me like it is saying I should not discuss Julia privately, ever. The Numpy version avoid that by scoping this advice to only discussion related to the project. For julia this could be replaced by "discussions pertaining to a community project".

Along a similar line, I think it would be worth scoping the paragraph around how the private channels of discussion of the julia's community are expected to be used. Especilly since two paragraphs before, there is a line about confidentiality when contacting a steward.

On a different, it feels like a tool like Slack does not really go in the direction described here: there is no persistence of the content, the content's discoverability is low, and finding the tool in the first place may not be obvious. I hardly consider it to be a pulic platform to begin with. So I have a hard time understanding how this CoC paragraph and e.g. Slack would interact.

logankilpatrick:
Part of the motivation for this is specifically making it part of our community structure to move away from mediums in which content disappears (or at least not using them for important discussions). Without specifically calling out Slack, it would indeed make the heavy use of it against this "Open" ideal.

KristofferC:
While that might be a good idea it feels a bit "harsh" to put that into the CoC for how people in the julia community should behave. Remember it is a "code" and it is under the "We expect the following standards to be observed and upheld by all participants in any community forum" heading. 

Also, I think openness and "permanentness" are kind of two different things. For example, I would consider the Julia slack to be open.

MasonProtter:
Yeah, perhaps a better avenue for something like this would be a separate "statement of values" instead of putting it in the code of conduct. If you put things into the CoC that we have no plan or mechanism to enforce, then I'd say that weakens the rest of the CoC because it makes everything seem more aspirational rather than enforce rules of engagement.

logankilpatrick:
Closing based on feedback

PR #1315, opened 2021-07-05
adds a 2021 publication

sadit:
Adds a publication to the 2021 list (in the page format)

Tellez, E.S., Ruiz, G., Chavez, E. et al. A scalable solution to the nearest neighbor search problem through local-search methods on neighbor graphs. Pattern Anal Applic 24, 763–777 (2021). https://doi.org/10.1007/s10044-020-00946-w


logankilpatrick:
Thank you!

PR #1319, opened 2021-07-11
new paper added

BERENZ:
Paweł Gola's paper from the Journal of Political Economy. Codes are available at author's webpage http://www.pawelgola.com/


logankilpatrick:
Can you add which Julia packages or the like were used heavily?

BERENZ:
He uses `Distributions.jl`, `StatPlots.jl` and `PyPlot.jl` packages but I cannot say which one was the most used. Codes are mainly custom functions and reporting results using plots.

logankilpatrick:
Thank you!

PR #1324, opened 2021-07-16
Schoene et al. 2019

brenhinkeller:
May be hard to otherwise find, since it doesn't cite the Julia paper, but the code used is now a registered package.


logankilpatrick:
Thanks!

PR #1325, opened 2021-07-17
add papers using julia

AshtonSBradley:
This PR just adds 3 papers to `research.md` that use julia for simulations.


logankilpatrick:
Looks great!

PR #1332, opened 2021-07-26
Create juliacon2021.md

logankilpatrick:
Hopefully this time I don't crash the site :/


logankilpatrick:
Based off of: https://github.com/JuliaLang/www.julialang.org/blob/main/blog/2021/04/gsod-2020-wrapup.md

tlienart:
add `+++` at the top as per comment on Slack, it should work

PR #1339, opened 2021-07-30
Remove JuliaObserver, promote JuliaPackages

iamed2:
This PR removes references to JuliaObserver, which is old, slow, and only references METADATA. I saw JuliaPackages was broken so I was going to remove that, but it looks like it was working as recently as June so I expect to see it come online at some point. Thus, I replaced the JuliaObserver link on the front page with one to JuliaPackages, with the hopes it comes back online.

### DIFF
--- a/index.html
+++ b/index.html
@@ -393,7 +393,6 @@ end
       <div class="row">
         <div class="col-12" style="text-align: center">
           <a class="btn btn-sm btn-outline-danger" href="https://juliahub.com/ui/Packages">JuliaHub: Package Search</a>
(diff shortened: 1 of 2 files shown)


DilumAluthge:
@djsegal

fredrikekre:
https://juliapackages.com/ just gives an error page ("Application error") ?

iamed2:
Last time I can verify it was working is June 27, which is fairly recent. We'll see what the maintainer says about its status.

logankilpatrick:
Merging since it seems Julia Packages is more up to date

PR #1340, opened 2021-08-01
Add blog post about survey of the General registry

giordano:


blog/2021/08/general-survey.md
```diff
@@ -0,0 +1,268 @@
...
+
+* most packages do follow good programming practices (documentation, testing,
+  continuous integration)! In particular:
+  * high fraction of packages with any tests (96%), and at least 10% of code as
+    tests (76%)
+  * almost all packages have setup for continuous integration (95%)
+  * documentation is not as popular; only 49% of packages have docs in the
+    repository.
+  * most packages either have a README or docs (88% have at least twenty lines
+	of one or the other)
```
pfitzseb, line 227:
```suggestion
  * most packages either have a README or docs (88% have at least twenty lines of one or the other)
```
This isn't rendered properly otherwise:
grafik

giordano: Thanks!

blog/2021/08/general-survey.md
```diff
@@ -0,0 +1,268 @@
...
+![Distribution of packages by number of lines of code](/assets/blog/2021-08-general-survey/lines-of-code.png)
+
+We also used [`WordCloud.jl`](https://github.com/guo-yong-zhi/WordCloud.jl) to
+create a word cloud of the packages in the General registry by number of lines
+and code.
+
+![Word cloud of packages in the General registry by number of lines of code](/assets/blog/2021-08-general-survey/wordcloud.png)
+
+The following table shows the 14 largest packages in the registry, by number of
+lines of code in their `src` directories.  We should note that most of these
```
ericphanson, line 43: As discussed on slack:
```suggestion
lines of code in their `src` directories.  We should note that many of these
```

vtjnash:
This was a fun read! But if you still have the data from these runs, it might be interesting to re-do the two histogram plots as KDE plots instead, as they can be better at showing this type of data: https://discourse.julialang.org/t/kernel-density-estimation-status/5928/3

PR #1344, opened 2021-08-07
alter platform support table to reflect current

vtjnash:
Win7 is EOL since almost 2 years ago

AArch64, PPC, and FreeBSD are all generally stable, but we don't have people able to actively support them


DilumAluthge:
Idk if I'd agree with moving FreeBSD to Tier 2.

DilumAluthge:
On an unrelated note, the colors need to be fixed so that e.g. "Tier 1" is always green, "Tier 2" is always orange, "Tier 3" is always red, and "Tier 4" is always black.

DilumAluthge:
For FreeBSD, I would suggest that we do #1345 instead.

downloads/index.md
```diff
@@ -348,7 +348,7 @@ The platforms currently supported by Julia are listed below. They are divided in
       <td rowspan="2"> FreeBSD </td>
       <td rowspan="2"> 11.0+ </td>
       <td> x86-64 (64-bit) </td>
-      <td> <font color="green">Tier 1</font> </td>
+      <td> <font color="green">Tier 2</font> </td>
```
ararslan, line 351:
```suggestion
      <td> <font color="green">Tier 1</font> </td>
```
This was addressed by https://github.com/JuliaLang/www.julialang.org/pull/1345.

ararslan (changes requested):

vtjnash:
updated to fix colors

downloads/index.md
ararslan: This isn't resolved

vtjnash: This is still updating FreeBSD to  be tier 2, as intended. Although I think (after this PR), it was mentioned we should also add mention of these changes to the current release notes?

vtjnash:
Bump. This has been ready to merge for awhile now.

downloads/index.md
```diff
@@ -304,12 +304,12 @@ The platforms currently supported by Julia are listed below. They are divided in
       <td> <font color="orange">Tier 2</font> </td>
     </tr>
     <tr>
-      <td> 7+ </td>
+      <td> 8+ </td>
       <td> i686 (32-bit) / x86-64 (64-bit) </td>
-      <td> <font color="crimson">Tier 3</font> </td>
+      <td> <font color="orange">Tier 2</font> </td>
```
ararslan, line 309: Is this change intended? I thought we do all of our Windows stuff with Windows 10 nowadays and don't check 8.

vtjnash: We don't actively check it, but it is close enough to 10 that we expect it to keep working (for now).  We will deprecate it from the list whenever it is EOL from Microsoft.

ararslan (approved):

downloads/index.md
giordano: Windows 8 was EOL in January 2016. Windows 8.1 reached end of mainstream support (whatever it is) in January 2018, and will reach end of extended support in January 2013

PR #1353, opened 2021-08-18
Update Arch Linux installation instructions

vnnovell:
Include the fact that the installation instructions are the same for Manjaro and Julia now lives in the community repository that is maintained by trusted users of Arch. See https://wiki.archlinux.org/title/official_repositories for details on different repositories in Arch.

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -181,14 +181,11 @@ Recent Debian/ubuntu distributions include their own build of Julia, which can b
 sudo apt install julia
 ```
 
-## Arch Linux
-The Arch User Repository has [a package for Julia](https://aur.archlinux.org/packages/julia-bin) that is built from the official binaries of Julia. To install it run:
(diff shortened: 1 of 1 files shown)


downloads/platform.md
```diff
@@ -181,14 +181,11 @@ Recent Debian/ubuntu distributions include their own build of Julia, which can b
 sudo apt install julia
 ```
 
-## Arch Linux
-The Arch User Repository has [a package for Julia](https://aur.archlinux.org/packages/julia-bin) that is built from the official binaries of Julia. To install it run:
+## Arch Linux and Manjaro
+The Arch Community Repository has [a package for Julia](https://archlinux.org/packages/community/x86_64/julia/) to allow for installation using pacman. To install it run:
```
giordano, line 185: This is a different package, which we _don't_ want to encourage users to install

giordano (changes requested):

Issue #1357, opened 2021-08-20
the blog is gone


StefanKarpinski:
https://www.julialang.org/blog/ is currently a 404

logankilpatrick:
Yep, just pushed a commit but not clear as to why

StefanKarpinski:
Locally, Franklin can't generate the site at all for me due to a syntax error in the JuliaCon highlights post.

logankilpatrick:
Good catch. It really is the little things (:head-bang)

rikhuijzer:
https://github.com/tlienart/Franklin.jl/issues/785

tlienart:
While I agree that Franklin can probably fail more clearly and block the deployment upon specific errors, there's currently a preview path for this website exactly to avoid such problems, so people should make sure that whatever they commit works on preview fully before they merge!

giordano:
Preview doesn't work for pull requests from forks, which is the case for #1356.  So at the moment we can't rely on always having the preview deployed

PR #1358, opened 2021-08-20
fix blog (#1357)

StefanKarpinski:


logankilpatrick:
Can't find anything in the Franklin docs about this, seems right though since it fixes thing locally. Merging for now

PR #1363, opened 2021-08-25
Eliminate recommendation to use atsign-assert

timholy:
Ugh, didn't realize this was still hanging around.


ViralBShah:
@timholy Is this ok to merge?

timholy:
It seems settled now, so yes. I updated the branch, when the tests finish this can be merged.

PR #1367, opened 2021-08-29
Add note that macOS ARM == M1 Mac

logankilpatrick:


downloads/index.md
```diff
@@ -179,7 +179,7 @@ Checksums for this release are available in both, [MD5](https://julialang-s3.jul
       <td colspan="3"> </td>
     </tr>
     <tr>
-      <th> macOS ARM <a href="/downloads/platform/#macos">[help]</a></th>
+      <th> macOS ARM (M1 Mac)<a href="/downloads/platform/#macos">[help]</a></th>
```
fredrikekre, line 182: Maybe
```suggestion
      <th> macOS ARM (M1 processor)<a href="/downloads/platform/#macos">[help]</a></th>
```
?

logankilpatrick: I think "M-series processor" might be good as well since one can imagine an M2 someday soon.

DilumAluthge: You could also consider "Apple Silicon".

PR #1369, opened 2021-08-30
Set upcoming release to 1.7.0-beta4

ararslan:
Also uncomment musl since the build actually went through

### DIFF
--- a/config.md
+++ b/config.md
@@ -24,6 +24,6 @@
 If the following lines are commented, the "upcoming release" section
 in `downloads/index.md` will not be shown.
 -->
-@def upcoming_release = "1.7.0-beta3"
+@def upcoming_release = "1.7.0-beta4"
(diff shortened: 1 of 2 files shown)


config.md
```diff
@@ -24,6 +24,6 @@
 If the following lines are commented, the "upcoming release" section
 in `downloads/index.md` will not be shown.
 -->
-@def upcoming_release = "1.7.0-beta3"
+@def upcoming_release = "1.7.0-beta4"
 @def upcoming_release_short = "1.7"
-@def upcoming_release_date = "July 7, 2021"
+@def upcoming_release_date = "August 24, 2021"
```
fredrikekre, line 29: Not August 30?

ararslan: The actual tag was made August 24, and it's what the REPL banner says, so I figured might as well be consistent with the REPL banner.

PR #1375, opened 2021-09-09
Slack: add the link to the Julia Slack for users that already have a Julia Slack account

DilumAluthge:
We currently have a link for new users to create a Slack account.

But we don't currently have a link for existing users to log back in.


DilumAluthge:
cc: @logankilpatrick

logankilpatrick:
Good idea, thanks!

PR #1376, opened 2021-09-13
update x86 mac description to mention Rosetta

ericphanson:
It sounds like this would've saved @mkborregaard some pain, and probably lots of other folks who don't know to ask on Slack :).


DilumAluthge:
Looks good to me.

logankilpatrick (approved):

PR #1378, opened 2021-09-18
Add #julia channel on Libera to community page

ninjin:
We never really left; just stayed with the network when the freenode
domain was “lost”.

### DIFF
--- a/community/index.md
+++ b/community/index.md
@@ -62,7 +62,7 @@ From: https://stackoverflow.com/questions/31821974/support-user-time-zone-in-emb
      <!-- 3 -->
      <div class="col-lg-4 col-md-6 feature">
        <h3>Chat</h3>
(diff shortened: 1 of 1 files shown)


logankilpatrick:
We could also put it into the https://julialang.org/community/otherchannels/ section

ninjin:
On Fri 01 Oct 2021, Logan Kilpatrick wrote:
> 
> @ninjin how large of a community is on this channel? I am happy to get this listed on the website but we usually reserve the main page for community with 1,000+ members 

Given the transient nature of IRC, it is pretty hard to tell. Based on very few samples, we tend to have less than 100 “idlers” (such as myself), but remember that these people are required to maintain an active connection to the network at all times, so it is not like say Slack, Zulip, Discord, or Matrix where essentially any user that has ever visited and has not deleted their account would count as a “member”. In addition to idlers, we have people dropping in now and then just asking questions, chatting, and then leaving as they may not be willing to have their computer (or a server) on 24/7. But to me these people are equally worthy to count as members of the channel, just that it is difficult to count them due to how IRC operates.

Lastly, I guess there is also the consideration that given how quickly and chaotically freenode went down in flames earlier this year [1] when we lost that channel, there was little time to migrate to Libera in an orderly fashion and I am sure plenty of our old freenode members might even think that there is no Julia IRC presence after the freenode debacle. At least we could see this in the discussion in #community a few weeks ago, where several senior members of the community thought this to be the case.

[1]: https://en.wikipedia.org/wiki/Freenode#Ownership_change_and_conflict

PR #1383, opened 2021-09-28
add juliaup in platform.md

heliosdrm:
According to @davidanthoff  in https://discourse.julialang.org/t/julia-v1-6-3-has-been-released/68737/7, "the Windows version is really not an experiment anymore".


DilumAluthge:
1. Juliaup lives in the JuliaLang GitHub organization
2. Juliaup uses the official release binaries

So I don't know if it belongs in the "Unofficial Binaries" section. I would instead make a separate section (immediately preceding the "Unofficial Binaries" section) specifically for Juliaup.

heliosdrm:
Or rather at the end of the "Windows" section?

DilumAluthge:
Eventually, Juliaup will support Windows, Linux, and macOS.

So, actually, I think it makes the most sense to put Juliaup in the "Cross-platform installer" section, the same section that talks about `jill.py`.

logankilpatrick:
Look good, thank you!

PR #1389, opened 2021-10-08
strongly encourage use of stable rather than LTS

StefanKarpinski:


logankilpatrick:
Is there any context on this? What promoted it?

StefanKarpinski:
I was looking at the downloads page (to download a 1.0 Julia to test something on it) and realized that other than the ordering, there was no real guidance of what version to use and that might explain why so many people look at the downloads and think "long term support—that sounds like something I want!" So I decided this was the place to add a note giving people the guidance that, no, unless you're at some organization with exceptionally low risk tolerance, you should not be using the LTS version.

PR #1391, opened 2021-10-09
jsoc: gsod: projects.md: fix typo

thazhemadam:


thazhemadam:
On a related note - I'm not an expert on this, but my understanding is that GSoD has changed since the 2020 edition, but to me, the page seems to be relatively the same. Should the docs be updated in any way to reflect that too?

logankilpatrick:
Yes, they likely should.

thazhemadam:
Do you have anything in mind for how they should be updated? Happy to make a PR if you want.

PR #1397, opened 2021-10-14
Remove the instructions for downloading unofficial binaries

DilumAluthge:
I don't really understand why we provide instructions for downloading unofficial binary distributions. Whenever someone reports a bug on GitHub or Discourse, the first thing we ask is "are you using the official binaries from julialang.org". And, as a general principle, we don't accept bug reports if they can't be reproduced with the official binaries.

So I don't really see the point in giving these instructions on the JuliaLang website.


ViralBShah:
Right they are not the official binaries.

DilumAluthge:
> I think it would make sense to provide a Flatpak package wrapping official binaries instead, and recommend these for distributions that support it.

I think that this could be useful.

DilumAluthge:
> Most Arch users would use the [package in the official repositories](https://archlinux.org/packages/community/x86_64/julia/) which is the usual build which doesn't use Julia's LLVM. The package mentioned here is from the unofficial community-maintained AUR

I see. Well, I still think it's better to recommend that people get their binaries from julialang.org, even if the AUR is repackaging the official binaries, since the AUR may lag behind.

PR #1401, opened 2021-10-18
ColPrac

willtebbutt:


logankilpatrick:
I should be more clear about what we are looking for here. I am interested in contributing guides where there is a list or section which spells out ways to contribute beyond the obvious. While the ColPrac guide is great for the specifics of contributing (and we should like it appropriately) I really want guides that say here are 5 things you can do today to start being a contributor

PR #1424, opened 2021-10-31
Update books.md

storopoli:
Added Julia Data Science from @storopoli, @rikhuijzer and @lazarusA.

I also took the liberty to add a final dot `.` in the last entry before this.


logankilpatrick:
Thanks you and congrats!

PR #1426, opened 2021-11-07
Update the downloads page

ViralBShah:
We had too many notices before people can get to the downloads. Reorganizing things a bit, on order to allow people to quickly get to the downloads, while having all the details available lower on the page.


downloads/index.md
```diff
@@ -1,30 +1,20 @@
...
-
 ~~~
 <a class="github-button" href="https://github.com/JuliaLang/julia" data-size="large" data-show-count="true" aria-label="Star JuliaLang/julia on GitHub">Star</a>
 ~~~
 
+Please star us [on GitHub](https://github.com/JuliaLang/julia). If you use Julia for your research, please do [cite us](https://julialang.org/research/).
+
 We provide several ways for you to run Julia:
 
 * In the terminal using the built-in Julia command line using the binaries provided below.
```
logankilpatrick, line 11:
```suggestion
* In the terminal using the built-in Julia command (download binaries provided below).
* In VS Code via the [Julia VS Code Extension](https://code.visualstudio.com/docs/languages/julia#_getting-started) (download binaries provided below)
```

logankilpatrick: This isn't totally complete since we only mention VS Code but seems right to add it to the list there

ViralBShah: The VS code extension does not provide a Julia download. This is not the right place for mentioning it.

logankilpatrick:
My preference would be to remove:
```
We provide several ways to download and run Julia:

In the terminal using the built-in Julia command line using the binaries provided below.
Using Docker images from Docker Hub maintained by the Docker Community.
```
If we want to point people to docker, we can have a specific section for that

logankilpatrick:
We can maybe also link this video in case someone needs a little bit more of a how to guide for installing Julia: https://www.youtube.com/watch?v=t67TGcf4SmM

ViralBShah:
Good one. I think we should link it (but not embed) on this page. We may want to embed it in other locations that are tailored to new users or even first time programmers.

ViralBShah:
I agree - the prime real estate is not the right place to talk about Docker. However, I no longer have a natural place to link the video.

logankilpatrick:
That's okay, one day soon we will revamp this page to make it a little fancier.

ViralBShah:
I like the video - so let's put it in the learning section of the website, where it can help first time users.

On this page, we should get out of the way as quickly as possible.

PR #1432, opened 2021-11-16
downloads: put the note about IP addresses back

StefanKarpinski:


fredrikekre:
It was moved to the bottom of this page, so you have to remove it there then.

StefanKarpinski:
Thanks, I missed that it was moved to the bottom, just went to the downloads page and saw that the note was gone (and the PR that moved it didn't have any meaningful comments about anything). It should still probably be at the top so that there's a decent chance that people have seen it before downloading. In theory those who distribute Julia on other channels should display the same note.

PR #1435, opened 2021-11-20
Set current stable version to 1.6.4

ararslan:
Also comment out the stable PowerPC table entry because the buildbots seem to have been offline for months and thus no binaries.


DilumAluthge:
> Also comment out the stable PowerPC table entry because the buildbots seem to have been offline for months and thus no binaries.

@vchuravy

PR #1436, opened 2021-11-20
Add Github and twitter handles

ViralBShah:


DilumAluthge (approved):

DilumAluthge:
For those Julia creators that have personal and/or academic websites, that would also be something that we could link (in addition to the GitHub and Twitter handles).

PR #1438, opened 2021-11-21
Fix why_contribute link

monsij:
The text "benefits of becoming a Julia contributor" had a typo in linking to the heading ID. 

contribute_page


ViralBShah:
Thanks @monsij

PR #1440, opened 2021-11-25
Windows 8 is EOLed

PallHaraldsson:
It's best to not promise it working, as it's not popular anyway (never was). Windows 8.1 IS supported still, and 8 in a sense by Microsoft, only by you upgrading to 8.1.

Powershell 4.0 can be relied on for Windows 8.1+, so that could be a reason if you want to support it. For Windows Server, versions still supported, you however could not rely on that recent (the shell mode in the REPL might however be less needed on Server).

Window 8 has 1.07% usage share, and 8.1. 3.17% (additionally).


DilumAluthge:
@staticfloat @vtjnash Are you fine with dropping support for Windows 8 and requiring Windows 8.1+

vtjnash:
Largely the same to me. It was tier 2 anyways

ararslan (approved):

PR #1354, opened 2021-08-18
Update the website to reflect that 1.6 is the new LTS

ViralBShah:
Most likely this can be merged when 1.7 is released.

PR #1419, opened 2021-10-23
blog post: Julia 1.7 Highlights

KristofferC:
Since the release is getting closer I think it makes sense to put this up for further feedback.

PR #1441, opened 2021-11-29
Set the stable version to 1.7.0, and set the LTS to 1.6

DilumAluthge:
TODO:
- [x] https://github.com/JuliaLang/julia/pull/43243
- [x] Update the date in this PR to reflect the correct date of the release

### DIFF
--- a/config.md
+++ b/config.md
@@ -12,7 +12,7 @@
 <!-- Templating of the Downloads -->
 @def stable_release = "1.7.0"
 @def stable_release_short = "1.7"
-@def stable_release_date = "Dec ??, 2021"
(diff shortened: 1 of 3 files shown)


DilumAluthge:
@ararslan Closes https://github.com/JuliaLang/www.julialang.org/pull/1354, based on your latest commit?

ararslan:
Yep. This should be good to merge any time but it would be nice to have it and #1419 merged close together.

DilumAluthge:
I think we probably want this PR, #1419, and the Discourse announcement to all happen at the same time. It would be best if the same person (ideally you) could execute all three at the same time.

I guess you should first check if #1419 is finalized.

PR #1444, opened 2021-11-30
In "Platform Specific Instructions" -> "macOS", don't recommend that the user use `sudo` when creating the symlink

DilumAluthge:
I don't think this `sudo` is actually necessary, although it would be good for someone to confirm this.


DilumAluthge:
Hmmm. I don't think this is always correct.

fredrikekre:
I think it is needed for all the commands, not just `ln`.

DilumAluthge:
Yeah, I think you are right. I think by default, `/usr/local` is owned by `root` and is read-only to non-`root` users.

DilumAluthge:
https://github.com/JuliaLang/www.julialang.org/pull/1445

PR #1452, opened 2021-12-01
Remove "update every 24h" from download info

SaschaMann:
This is no longer true as running the build on a schedule caused issues each release. The file rebuild is triggered manually during the release process.

cc @ararslan


ararslan:
Thanks!

PR #1454, opened 2021-12-01
Update link to apple silicon label

giordano:
Someone renamed the label `apple silicon` to `apple silicon (m1)`, so the [old link](https://github.com/JuliaLang/julia/labels/apple%20silicon) now points to an empty list


fredrikekre:
I have updated the label instead.

giordano:
Cool, thanks :smile:

PR #1456, opened 2021-12-03
Get to the downloads quicker

ViralBShah:
Risk personas are moved to just before LTS
Package logs are moved to the bottom of the page.


logankilpatrick:
I thought @StefanKarpinski  said we had to have IP address logging at the top?

ViralBShah:
I might have missed that. @StefanKarpinski is it required at the top, or can it at least be after the LTS (if not at the bottom)?

ViralBShah:
Netlify doesn't seem to like this PR.

PR #1458, opened 2021-12-06
Preview of DTable blogpost

fredrikekre:


ViralBShah:
No preview here (the link doesn't show anything). @logankilpatrick Are you the right person to look into this?

giordano:
Also in #1457 the preview doesn't work

logankilpatrick:
Perhaps @DilumAluthge ? I don't have access to the netlify account that this is being deployed from

logankilpatrick:
The files are being committed to the gh-preview branch in the previews folder, so it must be on the deploy side. We might need to clean that branch as well.

fredrikekre:
@tlienart  maybe^?

tlienart:
I'm not entirely sure whether this is my fault or not but my netlify account was suspended because I didn't provide a credit card (I wasn't using it but I'm now unsure whether the JuliaLang preview was somehow connected to my account and that this caused an interruption here?)

@logankilpatrick or @ViralBShah do you have access to the netlify account linked to this website?

logankilpatrick:
I don't have access to the netlify account but we could indeed do the other repo for previews.

tlienart:
Let's maybe open this in a separate issue; I think for the present blogpost as long as reviewer(s) can run it successfully locally (`serve()` + check the page looks fine) it shouldn't be a blocker that the preview is un-available.

ViralBShah:
Another issue seems better in the long run.

PR #1460, opened 2021-12-08
DTable blogpost fixes

krynju:
1. DataFrames filter legacy syntax removed - replaced with proper syntax and rerun its benchmarks
2. DataFrames note added on reductions - recommends that users should prefer to use Statistics.jl when using DataFrames.jl
3. The summary table with percentages is now fixed (configs slower than dtable had an error in the calculation, which led to weird percentage values)
4. The summary table has been adjusted to "x times faster", "x times slower" format due to being hard to understand
5. Recalculated the summary table and updated the values in the text to match it


krynju:
Thanks for quick merge!

blog/2021/12/dtable-performance-assessment.md
```diff
@@ -73,18 +73,18 @@ That means the benchmark code was accordingly adjusted to make sure the packages
...
-|       Reduce (single column)     |                   $3012.6%$ |                          $194.2%$ |
-|         Reduce (all columns)     |                   $2612.3%$ |                          $274.5%$ |
-|            Groupby (shuffle)     |                   $1700.1%$ |                          $-99.8%$ |
-| Reduce per group (single column) |                   $1921.4%$ |                          $-64.9%$ |
-| Reduce per group (all columns)   |                   $2130.0%$ |                          $-14.0%$ |
 
+|                        Operation | times faster than Dask     | times faster than DataFrames.jl  |
+| --------------------------------:| --------------------------:| --------------------------------:|
+|                          Map     |                      $4.9$ |                           $-2.5$ |
+|                       Filter     |                     $-1.6$ |                          $-14.5$ |
```
StefanKarpinski, line 82: I'm not sure what definition is being used here, but unless something takes negative time to run, it shouldn't be possible for these to be negative. Speedup multiple is typically computed as `reference_time/comparison_time` which must be positive. Come to think of it, I'm not entirely sure what definition would be used for percentage speedup in the first place? I guess you'd convert times into speeds (by taking the inverse) and then compute a percentage improvement of speeds? I.e. something like this:
```jl
100*(comparison_speed-reference_speed)/reference_speed =
100*(comparison_speed/reference_speed-1) =
100*(reference_time/comparison_time-1)
```
If that's how those were computed, then the relative speedups should be:
```
julia> [x/100+1 for x in [405.7, -11.3, 3012.6, 2612.3, 1700.1, 1921.4, 2130.0]]
7-element Vector{Float64}:
  5.0569999999999995
  0.887
 31.125999999999998
 27.123
 18.000999999999998
 20.214000000000002
 22.3
```

krynju: You're right on this, it's probably the better way to express this properly.

I know negative doesn't mathematically make sense, but I couldn't think of a better way to include both `x times faster` and `x times slower` in one table.

The idea is that you take the negative sign and turn the comparison around, so instead "dtable is x times faster than Y" it's "dtable is x times slower than Y"

In short that's the formula to get these results.
It's correct, but the negative 
```
julia> f = (x, y) -> begin
           if x < y
               (y-x)/x + 1
           else
               -(x-y)/y - 1
           end
       end
#29 (generic function with 1 method)

julia> f(3, 9)
3.0

julia> f(9, 3)
-3.0
```

StefanKarpinski: So the typical way to express the fact that something is in fact slower in an "times faster" comparison is to have a multiple that is less than one. For example, "0.8x faster" means "1.25x slower". Anything else is fishy and confusing and using negative values is just straight up wrong, imo. There's also the issue that the other non-negative values are not what they seem like they should be. Either you're computing speedup factor incorrectly now or you were computing percentage speedup incorrectly before (or I did the math wrong—also a possibility).

StefanKarpinski: The first table looks good, assuming the computations are correct. The second table still has negative values, so I'm assuming you didn't update that one? To avoid any ambiguity here, could you post the actual run times and I can compute them as I would compute them for comparison?

StefanKarpinski: Hey, @krynju, can we get this fixed? I'm kind of uncomfortable with possibly wrong, definitely confusing performance claims being in a blog post on JuliaLang.org — it would be good to get it fixed sooner rather than later.

krynju: Sure, I'll get this done in ~30min. Sorry, been a little busy!

krynju: data + code https://gist.github.com/krynju/c32c4a5cf0f8c6e9252778e9d14b71be
Sorry again for the delay!

PR #1471, opened 2021-12-26
Update REPL for try Julia section

logankilpatrick:


DilumAluthge:
Looks like it's 1.6.3 right now. Any plans on upgrading it to 1.7.1!

logankilpatrick:
I think I have the permission on Repl.it to globally update the Julia version that is used for newly created sessions but I need to confirm if that is true and how to bump to 1.7. It uses NIX so should be simple

PR #1486, opened 2022-01-04
classes.md added UC courses

gvdr:
Added three courses I teach in Julia at the University of Canterbury, NZ.


logankilpatrick:
Thank you!

PR #1487, opened 2022-01-04
Add 3 classes at Université de Montréal

tpoisot:


logankilpatrick:
Thanks for adding these!

PR #1489, opened 2022-01-06
Update classes.md with Oregon State Uni courses

SimonEnsemble:
adding the three classes I teach, where the students use Julia and Pluto notebooks.


logankilpatrick:
Thank you!

PR #1491, opened 2022-01-10
GSoC: wasm: remove iodide

fonsp:
Because Iodide is no longer maintained, we have Pluto integration instead: https://pluto-wasm-backend.netlify.app/editor.html


logankilpatrick:
Thank you!

PR #1493, opened 2022-01-11
Update RL section under Flux.md

findmyway:
cc @logankilpatrick


logankilpatrick:
Can you add if this is a 175 or 350 hour project?

findmyway:
Updated

logankilpatrick:
Thank you!

PR #1495, opened 2022-01-12
Create governance.md

logankilpatrick:
First pass at an updated governance structure.


logankilpatrick:
One part that we might want to omit or put in practice is:
> Signatories provide oversight to make sure that funding is being used for legitimate purposes.

As far as I know, we don't have a mechanism for this and the oversight comes directly from Avik.

StefanKarpinski:
We should update the set of signatories so that Avik is one of them, but I do think that ultimately the buck stops with the signatories.

Since this is explicitly a governance page, unlike the blog post it's a copy of, which was just about Julia-related entities, I would separate it into organizations that have some role in governance, which would include the "Julia", "Julia Stewards" and "NumFOCUS" sections and separate that from the non-governance organizations, which would be "The Julia Lab" and "Julia Computing". The way this is currently organized seems to imply that those entities are part of Julia's governance; it does say otherwise if you actually read it, but it seems better not to imply that at all.

logankilpatrick:
Maybe this should be an entities page instead of governance? @StefanKarpinski I think having part about the Lab and JC makes sense to me.

Avik is also one of the signatories already

ChrisRackauckas:
Avik's affiliation to Julia Computing is orthogonal to any involvement with the JuliaLang organization. That kind of conflation is exactly what Stefan is talking about. Not only is it just wrong, it gives people a very misleading idea of what the JuliaLang organization actually is. Julia Lab and Julia Computing are completely separate entities and I do not see why they would be mentioned on the page any more than say Invenia or Steven Johnson's group.

StefanKarpinski:
> Avik is also one of the signatories already

Good, then that part is accurate. One of the signatories is just more active in the role than others.

Calling this a governance page seems fine to me, just move the two sections about entities without any governance role further down into a "Related Organizations with no Governance Role" and add a little paragraph introducing that by saying something like this:

> There are organizations that people sometimes mistakenly believe have some kind of official control over the Julia project. This section details two of them, but please note that neither has any official relationship to the Julia project, nor any governance capacity with respect to the project. Individuals who have some control over the project and happen to belong to these organizations, have that control as individuals and not as members of those organizations and takes the separation of those roles quite seriously.

logankilpatrick:
Just to clarify, are you saying I am conflating the two things? 

To be clear, I was just pointing out that technically Avik is already a julia project signatory and is listed on this page.

ChrisRackauckas:
He is a Julia project signatory with a Julia Computing affiliation. That does not mean that Julia Computing is part of the JuliaLang governance, but that this person who is a signatory has an affiliation with both groups.

StefanKarpinski (commented):
This looks good to me now. I have a couple of tweaks, but I'll merge this first and will make those in a separate PR.

governance.md
```diff
@@ -0,0 +1,45 @@
...
+
+The Julia Project also employs a Community Manager. The role began with [Alex Arslan in 2017](https://discourse.julialang.org/t/welcome-alex-as-our-julia-community-manager/3756) and it is currently filled by [Logan Kilpatrick](https://twitter.com/OfficialLoganK) who started in early 2020.
```
giordano, line 11: > The Julia Project also employs a Community Manager.

@logankilpatrick is this still accurate?

PR #1498, opened 2022-01-14
Added Prifling (2021) reference

roding:


logankilpatrick:
Hey! Can you add the main Julia Package(s) used in this paper?

roding:
Hi, I actually didn't use any packages at all. Is it ok as is then?

logankilpatrick:
Ah, so it just uses base Julia?

roding:
Yep, only base Julia.

logankilpatrick:
Thank you!

PR #1500, opened 2022-01-16
Update README.md to add note on Netlify previews

logankilpatrick:
Assuming the preview build works, just adding a note on this so I don't forget in the future.


logankilpatrick:
It looks like the build previews are timing out after 31 minutes, not sure if I have something misconfigured or not

PR #1502, opened 2022-01-16
Remove tryjulia which does not work.

ViralBShah:
@logankilpatrick Note that this does not work and I am removing it.

### DIFF
--- a/index.html
+++ b/index.html
@@ -24,7 +24,7 @@
     </div>
   </div>
 
-	
+
   {{insert navbar.html}}
 
   <!--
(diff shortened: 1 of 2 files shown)


logankilpatrick:
@ViralBShah Why does this not work? It was working last time I checked it and updated this fill?

ViralBShah:
Doesn't load at all for me - just get a big blank rectangle. Maybe try incognito mode or something where you are not logged in?

ViralBShah:
I tried the URL by itself and it works fine. Maybe it has trouble in the iframe?

PR #1504, opened 2022-01-16
Revert "Remove tryjulia which does not work."

logankilpatrick:
Reverts JuliaLang/www.julialang.org#1502 so I can test the page in an incognito tab

### DIFF
--- a/index.html
+++ b/index.html
@@ -24,7 +24,7 @@
     </div>
   </div>
 
-
+	
   {{insert navbar.html}}
 
   <!--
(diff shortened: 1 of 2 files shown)


logankilpatrick:
@ViralBShah it looks like Chrome's incognito window breaks this. I verified that without incognito, even when you are signed out, it still works. Do you think it's worth adding a note to turn off incognito if it's on?

logankilpatrick:
Safari private browsing seems to not have an issue with this

PR #1506, opened 2022-01-16
Add a new course to Julia in the classroom page

jdossgollin:
Specifically my Environmental Data Science class at Rice University, https://jdossgollin.github.io/environmental-data-science/


logankilpatrick:
Thank you!

PR #1511, opened 2022-01-24
move macOS M1 build into footnote

### LINKED ISSUE #1510: Presentation of the M1 Julia download misleads users
From Slack

image

I think the M1 download should either be completely removed (until 1.8 irons out the show stopper bugs) or put into an extra section "Technical Preview", which could btw also contain the musl build...

laborg:
Fix #1510

This PR moves the M1 build into a footnote, so that users don't think that the M1 build is what they should download.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -94,7 +94,8 @@ You should *only* be using the long-term support (LTS) version of Julia if you w
 ~~~
 <sup>[1]</sup>
 ~~~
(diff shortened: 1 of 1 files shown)


giordano:
Superseded by #1513?

laborg:
Honestly, I don't think that the suffix `(experimental)` is well placed in the "Stable" section, but at least its better than the current state.

fonsp:
Sorry, I didnt see this PR! I think it might better reflect the current state... Can we still consider this PR?

DilumAluthge:
If people prefer this PR, we can do that instead. It will need to be re-opened and rebased, and the merge conflicts will need to be fixed.

PR #1517, opened 2022-01-26
[GSOC] add GraphNeuralNetworks.jl

CarloLucibello:
I'm not sure if I should put this as a separate entry (as I did) or on Flux's page.
cc @Wimmerer

Flux's page is a bit rotten and bloated, we should polish it. Maybe @logankilpatrick and @darsnack are already onto it?


logankilpatrick:
I have no preference for which page gets updated. It's up to @DhairyaLGandhi

DhairyaLGandhi:
It can go on the Flux page.

logankilpatrick:
@DhairyaLGandhi Can you clean up the page to make note of this and send people over to the Flux page?

ViralBShah:
I'm merging this for now. Better to have it and refactor later, than not.

CarloLucibello:
Thanks!

PR #1519, opened 2022-01-28
Fix broken github links

nodivbyzero:
Links to `Help Wanted Issues` and `Good First Issue` on Home page were broken.
This PR fixes those links.

### DIFF
--- a/_layout/foot_general.html
+++ b/_layout/foot_general.html
@@ -37,8 +37,8 @@
         <li><a href="/contribute">Contributor's Guide</a></li>
         <li><a href="https://github.com/JuliaLang/julia/issues">Issue Tracker</a></li>
         <li><a href="https://github.com/JuliaLang/julia/security/policy">Report a Security Issue</a></li>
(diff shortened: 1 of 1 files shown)


fredrikekre:
The existing links work for me.

nodivbyzero:
@fredrikekre Can you try to open them in Incognito mode?

logankilpatrick:
I believe you have to be signed into GitHub for them to work if I recall correctly

fredrikekre:
Ah, maybe <https://github.com/search?q=is%3Aopen+is%3Aissue+language%3AJulia+label%3A%22help+wanted%22> works regardless of whether you are logged in?

logankilpatrick:
Closing for now since those links exist specifically to be broader than the issues just in core julia. If you think there's something we can do to remind folks they need to sign in, let me know!

logankilpatrick:
Updated to the version @fredrikekre sent above, thank you both!

Issue #1272, opened 2021-04-16
Add JuliaMono webfonts to /assets/ ?


cormullion:
We could add the JuliaMono webfonts to the julialang.org website. They could live in, say,  `www.julialang.org/assets/fonts/JuliaMono/webfonts/` as the files:

```
JuliaMono-Black.woff2
JuliaMono-Bold.woff2
JuliaMono-BoldLatin.woff2
JuliaMono-ExtraBold.woff2
JuliaMono-Light.woff2
JuliaMono-Medium.woff2
JuliaMono-Regular.woff2
JuliaMono-RegularLatin.woff2
JuliaMono-SemiBold.woff2
```

The total size of this new directory would be about 5MB. People would then be able to use the fonts for package documentation, blog posts, [etc](https://github.com/search?o=desc&q=juliamono+language%3Acss&s=indexed&type=Code).. To do this, they would add something like this to a CSS file:

```css
@font-face {
	font-family: JuliaMono;
	src:
		local('JuliaMono-Regular'),
		url("https://www.julialang.org/assets/fonts/JuliaMono/webfonts/JuliaMono-Regular.woff2");
}
```

I've no idea what the bandwidth would be, or whether that would be a cost/pricing issue? At present, the actual OpenType fonts are [stored on github](https://github.com/cormullion/juliamono), but the webfonts (WOFF) are served from there via the NPM/JavaScript-focussed https://www.jsdelivr.com CDN. It works fine (but I have no more information than that).

There would be no visible changes to this website. (And I'm not saying that people should start using the fonts anywhere, unless they want to. 😄) 

I don't know whether you would want a disclaimer in the directory somewhere to the effect that the fonts aren't "official"? Although it does promote Julia, in a way...

fredrikekre:
> There would be no visible changes to this website.

Obviously code on the website should use it!

staticfloat:
So is the benefit here that users don't have to use `jsdelivr` but can instead use `julialang.org`?  It's not that I am afraid of the extra traffic, (Fastly is quite generous with us so it should be fine) but more that I'm not sure if it's much benefit so I'd prefer to keep it being served on other people's bills than ours as long as makes sense.  :P

cormullion:
Well I don't know anything about these CDNs - which ones are good, which aren't, whether two are better than one, how much downloads cost, etc - so I can't put any numbers on the benefits of using "julialang.org" or another. Might as well leave everything as it is for now... 😀

[...]

PR #1520, opened 2022-01-30
Switch to JuliaMono for code blocks

rikhuijzer:
This PR is a shot at #1272. Benefits of using this font that I can think of are more consistent appearance because this ensures that the font is available and coolness because it is JuliaMono. Drawback is slower page loading because the font has to be downloaded.

I also propose to use the Artifacts system instead of storing the files in the repo or using a CDN. CDNs are pretty useless because browsers do not cache the same thing if it comes from different domains for security reasons (https://httptoolkit.tech/blog/public-cdn-risks/). Even worse, CDNs are bad for privacy and may slow the page down due to an extra DNS lookup.

## Preview

The preview is at zoom level 100% because that's probably how most people read the page. The screenshot is from a code block at https://julialang.org/blog/2021/10/DEQ/.

### Consolas (current monospace font)

image

### JuliaMono

image

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -46,21 +46,8 @@ jobs:
 
       - uses: julia-actions/cache@v1
       - uses: julia-actions/julia-buildpkg@v1
-
-      # This ensures that NodeJS and Franklin are loaded then it installs
-      # highlight.js which is needed for the prerendering step.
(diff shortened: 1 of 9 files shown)


rikhuijzer:
I give up. I did basically the same at https://github.com/rikhuijzer/JuliaTutorialsTemplate/pull/2 and that worked without any problems.

PR #1522, opened 2022-01-31
Fix link

enricoschumann:
The site looks somewhat different from here https://web.archive.org/web/20190221214631/https://lectures.quantecon.org/jl/ 
But I think the new link points to the correct site.


logankilpatrick:
Thanks for this!

PR #1523, opened 2022-01-31
[jsoc] Add GUI and interop pages

barche:
This adds projects for CxxWrap and QML. I also removed the stale commented out topic in tooling.

CC @SimonDanisch


logankilpatrick:
Thank you!!

PR #1529, opened 2022-02-03
add makie and jsserve gsoc proposal

piever:
This adds a project on UI creation with Makie + JSServe.

CC: @SimonDanisch


logankilpatrick:
Thank you!

PR #1530, opened 2022-02-03
add macOS tarball to the download table

Moelf:


logankilpatrick:
Can you clarify why this is necessary?

Moelf:
https://gitlab.cern.ch/sft/lcgcmake/-/merge_requests/1087#note_5232536

because sometimes a tarball is more suitable and since we already have those, we should put them in the table.

PR #1546, opened 2022-02-14
New JSS paper

BERENZ:
This paper is about an R package the JuliaConnectoR that allows connecting Julia and R. I am not sure whether the formatting should be the same as for Julia packages.


logankilpatrick:
Thank you!

PR #1553, opened 2022-02-14
Update 10years.md: Add George Datseris subsection

Datseris:
I am sorry I am late to the party, I just saw this today. Neverthelss, I am very grateful and would like to contribute my retrospect as well. To the next 10 years then :)


logankilpatrick:
Thanks!

PR #1557, opened 2022-02-14
Update 10years.md: Add Sundar R's subsection

digital-carver:
Thoughts that have been bouncing around in my mind anyway, glad I got an opportunity to write them down here.


logankilpatrick:
Well put! Thanks

PR #1569, opened 2022-02-15
Update 10years.md: Tamas Gal @tamasgal

tamasgal:
Coming late to the party but maybe not too late `;)`


fredrikekre:
https://github.com/JuliaLang/www.julialang.org/commit/6aaa5536155c39a2b44eee0169859ee637c69bf3

PR #1576, opened 2022-02-15
Move around FluxML and other ML projects for NumFocus

darsnack:
After talking with @ViralBShah, it seems that FluxML should apply under the NumFocus umbrella this year. I've separated Flux-specific projects from other ML-related projects. I also added some projects based on current discussions with students.

cc Flux mentors: @ToucheSir @lorenzoh @DhairyaLGandhi
cc mentors that are affected: @maleadt @findmyway @jonathan-laurent @aviks


jsoc/projects.md
```diff
@@ -42,4 +42,5 @@ We also have Julia project's available under other organizations. If you are app
 * [SciML - NumFocus](https://github.com/numfocus/gsoc#sciml)
   - [SciML](/jsoc/gsoc/sciml/) - Scientific Machine Learning
   - [Differential Equations](/jsoc/gsoc/diffeq/) - Numerical methods for high-performance solving of differential equation models.
+* FluxML - NumFocus
```
darsnack, line 45: It seems like we will need to also make a PR to https://github.com/numfocus (is this right @logankilpatrick ?). Can we point to `/jsoc/gsoc/flux` as our "ideas page"?

PR #1585, opened 2022-02-15
Add new package to publication list - EntropyHub.jl

MattWillFlood:
Proposing a new reference in the list to a publication about a new Julia package called EntropyHub.jl.


logankilpatrick:
Thanks!

PR #1587, opened 2022-02-16
Add a "fallback" link for users in case the embedded repl.it doesn't load

DilumAluthge:


masad-frost:
Hey! Are embeds breaking for you guys?

PR #1591, opened 2022-02-16
JSoC: Remove obsolete MLJ projects

ablaom:
cc@logankilpatrick


logankilpatrick:
Thanks! I will merge this, but please also add if each project is 175 or 350 hours

PR #1627, opened 2022-02-25
Remove SciML and diffeq projects

logankilpatrick:


logankilpatrick:
@ChrisRackauckas it looks like you have a different projects page which is more up to date on the SciML site, can I remove these from here? They are not being shown right now just sitting in the background.

PR #1629, opened 2022-02-25
make expected results explicit in web app project

piever:
Just a small change to make the "expected results" section explicit (before the expected results were there, but the section had no name).

CC: @SimonDanisch


logankilpatrick:
Thank you!!

PR #1632, opened 2022-02-25
JSOC: Add detail to one MLJ project, and add list of projects (doc)

ablaom:
@logankilpatrick 

I've removed the "/toc" someone added because it doesn't seem to be rendering correctly, and added manually a list of projects instead. However, these are just text, not hyperlinks.


logankilpatrick:
I fixed the TOC

ablaom:
Thank you @logankilpatrick ❤️

PR #1634, opened 2022-02-25
Update with outcomes, skills, and difficulty for some compiler projects

ianatol:
CC: @vtjnash @logankilpatrick


logankilpatrick:
Thank you!

PR #1635, opened 2022-02-26
Update loop optimization jsoc

chriselrod:
I'm not particularly qualified to help with the first project, but perhaps @vchuravy or @MasonProtter (given your recent experience with StaticCompiler) are willing to help mentor that one, or know someone who would be?

It's not directly related to loops, but I wanted to add something that actually involves writing Julia code, and it seemed like it'd be nice if it produces something that is generally useful.
The idea would be that it lets someone write macros to apply a custom pipeline to a function.
```julia
@mycustompipeline function foo(args...)
    # do something
   3 / 8
end
```

Perhaps it could also help people experiment with reordering the default passes to see their impact on runtime performance.
If it seems too difficult, I could remove it and leave just the one project. Or maybe it'd make more sense to move this to something not loop specific, and give it real mentors (again, assuming some exist).


logankilpatrick:
Merging in for now but feel free to continue the discussion here

PR #1641, opened 2022-03-10
Minor doc, corrected link to support tiers

sylvaticus:


logankilpatrick:
Nice catch, thank you!

PR #1648, opened 2022-03-21
Update benchmark plotting scripts

acxz:
I tried to run the benchmark plotting script to recreate the plot however ran into some difficulties with older syntax.
This PR updates the benchmark plotting syntax to be up to date with the ipynb used for generating the plot and runs successfully for me under Julia v1.7.

I also came across a duplicate benchmarks svg file that is not being used. The svg file being used in the benchmarks.md page is located under _assets/benchmarks/ and not under _assets/images/


MilesCranmer:
Awesome!

Since you are updating the plot anyways, do you think you could slightly offset the points along the x-direction as a function of the specific benchmark (e.g., mandelbrot gets x+=0.01, pi_sum gets x+=0.02, etc.)? It would be great to see the overlapped benchmarks.

acxz:
I see, that's smart, yeah let me try to see how that can be done.

ViralBShah:
Is this ready to merge?

acxz:
@ViralBShah yep!

MilesCranmer:
@acxz Any idea why the plot looks like it hasn't changed yet? 

https://web.archive.org/web/diff/20220118113840/20220508160242/https://julialang.org/benchmarks/

acxz:
This PR doesn't update the plot.

MilesCranmer:
Ah, I see. What is left to actually upload the plot image itself?

MilesCranmer:
Maybe we want to first manually update the csv/svg? (since there are some large performance changes since the last update)

Then, an idea would be to set up a GitHub action on the Microbenchmarks repo to automatically submit a PR to this repo if the benchmarks are significantly different than the currently saved ones?

acxz:
We should move this convo to https://github.com/JuliaLang/Microbenchmarks/issues/48

but basically we need to fix: https://github.com/JuliaLang/Microbenchmarks/issues/29 first

PR #1650, opened 2022-03-24
Downloads: M1 preview release experimental annotation

fonsp:
Part 2 of #1513


DilumAluthge (approved):

giordano:
For what is worth, julia v1.8 for the M1 is in a _much_ better shape than in v1.7

PR #1656, opened 2022-04-05
Update research.md

ranocha:
Add another research article.

Closes https://github.com/trixi-framework/Trixi.jl/issues/724


ViralBShah:
We are actually no longer accepting more publications since the list is no longer manageable. I'll merge this one for now, but we'll probably remove the entire list.

PR #1661, opened 2022-04-10
New Blog Post: SimpleChains.jl Specializing Scientific Machine Learning with Julia

ChrisRackauckas:
Still finalizing the code in the post. SimpleChains.jl will be open sourced upon the release of this.

### DIFF
--- a/blog/2022/04/simplechains.md
+++ b/blog/2022/04/simplechains.md
@@ -1,26 +1,209 @@
-@def rss_pubdate = Date(2022, 2, 14)
-@def rss = """10 years ago today, we published 'Why we Created Julia' ..."""
-@def published = "14 February 2022"
-@def title = "Why We Use Julia, 10 Years Later"
-@def authors = """The Julia Community"""
+@def rss_pubdate = Date(2022, 2, 12)
+@def rss = """10x-ing PyTorch: Specializing Scientific Machine Learning with Julia ..."""
+@def published = "12 February 2022"
+@def title = "10x-ing PyTorch: Specializing Scientific Machine Learning with Julia"
+@def authors = """Chris Elrod, Chris Rackauckas"""
 @def mintoclevel=2
 @def maxtoclevel=3
 
+# 10x-ing PyTorch: Specializing Scientific Machine Learning with Julia
 
-#### Announcing: SimpleChains.jl
+Machine learning is a huge discipline, with applications ranging from natural language processing to solving partial differential equations. It is from this landscape that major frameworks, such as PyTorch, TensorFlow, and Flux.jl, arise to be packages for "all of machine learning". While some of these frameworks have the backings of megacorporations, specifically Facebook and Google, driving their development, the Julia community has relied on the speed and productivity of the Julia programming language itself in order for its open source community to keep up with the pace of development. It is from this aspect which Flux.jl derives its "slimness": while PyTorch and TensorFlow include entire separate languages and compilers (torchscript, XLA, etc.), Flux.jl is just Julia. It is from this that the moniker "you could have built it yourself" is commonly used to describe Flux.jl.
 
-We're happy to announce the open sourcing of SimpleChains.jl, a library developed by Julia Computing and PumasAI in collaboration with Roche and the University of Maryland, Baltimore.
+However, in this post we'll take a different look at how the programmability of Julia can help in the machine learning space. Specifically, by targetting the grand space of "all machine learning", the frameworks inevitably make trade-offs that accelerate some aspects of the code at the detriment to others. This comes from the inevitable trade-off between simplicity, generality, and performance. However, the ability to easily construct machine learning libraries thus presents an interesting question: can this development feature be used to easily create alternative frameworks which focus its performance on more non-traditional applications or aspects?
 
-SimpleChains.jl origin is as a part of the DeepPumas product, where it was optimized for quickly fitting small models on the CPU. Early development favored a design that would:
+The answer is yes, you can quickly build machine learning frameworks which greatly outperform the giants in specialized cases using the Julia programming language, and we demonstrate this with our new release SimpleChains.jl.
+
+## Scientific Machine Learning (SciML) and "Small" Neural Networks
+
+SimpleChains.jl is a library developed by Julia Computing and PumasAI in collaboration with Roche and the University of Maryland, Baltimore. The purpose of SimpleChains.jl is to be as fast as possible for small neural networks. This is because SimpleChains.jl originated as a solution for the DeepPumas product for scientific machine learning (SciML) in clinical pharmacology. In this domain, small neural networks are combined with known nonlinear mixed effects models (statistical models with differential equations) to discover previously unknown mechanisms and prognostic factors. For a short introduction to how this is done, check out the following video by Niklas Korsbo:
+
+<iframe width="560" height="315" src="https://www.youtube.com/embed/TFB_lt1KMto" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+
+This SciML methodology has been shown across many disciplines, from black hole dynamics to the development of earthquake safe buildings, to be a flexible method capable of discovering guiding (bio)physical equations. Here's a recent talk with walks through the various use cases of SciML throughout the sciences:
+
+<iframe width="560" height="315" src="https://www.youtube.com/embed/eSeY4K4bITI?start=668" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+
+For more details on the software and methods, check out this manuscript. 
+
+However, the unique aspects of how neural networks are used in these contexts make it rife for specializing of performance. Specifically, in the context of machine learning one normally relies on the following assumption: the neural networks are large enough that the O(n^3) cost of matrix-matrix multiplication (or other kernels like convolutions) domainates the the runtime. This is essentially the guiding principle behind most of the mechanics of a machine learning library:
+
+1. Matrix-matrix multiplication scales cubicly while memory allocations scale linearly, so attempting to mutate vectors with non-allocating operations is not a high priority. Just use `A*x`.
+2. Focus on accelearting GPU kernels as fast as possible! Since these large matrix-matrix operations will be fastest on GPUs and are the bottleneck, performance benchmarks will essentially just be a measurement of how fast these specific kernels are.
+3. When doing reverse-mode automatic differentiation (backpropogation), feel free to copy values to memory. Memory allocations will be hidden by the larger kernel calls. 
+4. Also, feel free to write a "tape" for generation backpropogation. The tape does add the cost of essentially building a dictionary during the forward pass, but that will be hidden by the larger kernel calls.
+
+Do these assumptions actually hold in our case? And if they don't, can we focus on these aspects to draw more performance out for our use cases?
+
+## Digging In: Small Neural Network Performance Overheads
+
+It's easy to show that these assumptions breakdown when we start focusing on this smaller neural network use case. For starters, lets look at assumption (1). It's not hard to show where these two will unravel:
+
+```julia
+using LinearAlgebra, BenchmarkTools
+function alloc_timer(n)
+    A = rand(n,n)
+    B = rand(n,n)
+    C = rand(n,n)
+    t1 = @belapsed $A*$B
+    t2 = @belapsed mul!($C,$A,$B)
+    t1,t2
+end
+ns = 2 .^ (2:9)
+res = [alloc_timer(n) for n in ns]
+alloc   = [x[1] for x in res]
+noalloc = [x[2] for x in res]
+ 
+using Plots
+plot(ns,alloc,label="*",xscale=:log10,yscale=:log10,legend=:bottomright,
+     title="Micro-optimizations only matter for small matmuls")
+plot!(ns,noalloc,label="mul!")
+savefig("microopts.png")
+```
+
+![](https://i0.wp.com/www.stochasticlifestyle.com/wp-content/uploads/2019/09/microopts.png?w=600&ssl=1)
+
+When we get to larger matrix-matrix operations, such as 100x100 * 100x100, we can effectively write off any overheads due to memory allocations. But we definitely see that there is a potential for some fairly significant performance gains in the lower end! Notice too that these gains are realized by using the pure-Julia LoopVectorization.jl as the standard BLAS tools tend to have extra threading overhead in this region (again, not optimizing as much in this region). 
+
+Also, if you've been a riding the GPU gospel without looking into the details this may be a shocker but it's not hard to show the seems on this one either. GPUs are designed as dumb slow chips with many cores, and thus they are only effective on very parallel operations, such as large matrix-matrix multiplications. It is from this point that assumption (2) is derived for large newtork operations. However, it is again in the case of the small where such GPU kernels will be outperformed by well-designed CPU kernels due to the lack of parallel opportunities:
+
+```julia
+using LinearAlgebra, BenchmarkTools
+function alloc_timer(n)
+    A = rand(n,n)
+    B = rand(n,n)
+    C = rand(n,n)
+    t1 = @belapsed $A*$B
+    t2 = @belapsed mul!($C,$A,$B)
+    t1,t2
+end
+ns = 2 .^ (2:9)
+res = [alloc_timer(n) for n in ns]
+alloc   = [x[1] for x in res]
+noalloc = [x[2] for x in res]
+ 
+using Plots
+plot(ns,alloc,label="*",xscale=:log10,yscale=:log10,legend=:bottomright,
+     title="Micro-optimizations only matter for small matmuls")
+plot!(ns,noalloc,label="mul!")
+savefig("microopts.png")
+```
+
+![](https://i0.wp.com/www.stochasticlifestyle.com/wp-content/uploads/2019/09/microopts.png?w=600&ssl=1)
+
+However, matrix-matrix operations only occur when batching is able to be used (where each column of the B matrix in A*B is a separate batch). In many cases in scientific machine learning, such as [the calculation of vector-Jacobian products in ODE adjoints](https://youtu.be/6hhF6Llv4sI?t=342), this operation is a matrix-vector multiplication. These operations are smaller and only O(n^2), and as you would guess these effects are amplified in this scenario:
+
+```julia
+using LinearAlgebra, BenchmarkTools
+function alloc_timer(n)
+    A = rand(n,n)
+    B = rand(n,n)
+    C = rand(n,n)
+    t1 = @belapsed $A*$B
+    t2 = @belapsed mul!($C,$A,$B)
+    t1,t2
+end
+ns = 2 .^ (2:9)
+res = [alloc_timer(n) for n in ns]
+alloc   = [x[1] for x in res]
+noalloc = [x[2] for x in res]
+ 
+using Plots
+plot(ns,alloc,label="*",xscale=:log10,yscale=:log10,legend=:bottomright,
+     title="Micro-optimizations only matter for small matmuls")
+plot!(ns,noalloc,label="mul!")
+savefig("microopts.png")
+```
+
+![](https://i0.wp.com/www.stochasticlifestyle.com/wp-content/uploads/2019/09/microopts.png?w=600&ssl=1)
+
+And remember, the basic operations of a neural network are `sigma.(W*x .+ b)`, and thus there's also an O(n) element-wise operation. As you would guess, this operation becomes more significant as n gets smaller and is requires even more consideration for memory operations. 
+
+```julia
+using LinearAlgebra, BenchmarkTools
+function alloc_timer(n)
+    A = rand(n,n)
+    B = rand(n,n)
+    C = rand(n,n)
+    t1 = @belapsed $A .* $B
+    t2 = @belapsed ($C .= $A .* $B)
+    t1,t2
+end
+ns = 2 .^ (2:11)
+res = [alloc_timer(n) for n in ns]
+alloc   = [x[1] for x in res]
+noalloc = [x[2] for x in res]
+ 
+using Plots
+plot(ns,alloc,label="=",xscale=:log10,yscale=:log10,legend=:bottomright,
+     title="Micro-optimizations matter for BLAS1")
+plot!(ns,noalloc,label=".=")
+savefig("microopts_blas1.png")
+```
+
+![](https://i0.wp.com/www.stochasticlifestyle.com/wp-content/uploads/2019/09/microopts_blas1.png?w=600&ssl=1)
+
+This already highly motivates a project focused on the performance for this case, but assumptions (3) and (4) point us to additionally look at the implementation of the backpropogation. The [trade-off between different machine learning library approaches to automatic differentiation has already been discussed at length](https://www.stochasticlifestyle.com/engineering-trade-offs-in-automatic-differentiation-from-tensorflow-and-pytorch-to-jax-and-julia/), but what the general discussions can miss is the extra opportunities afforded when really specializing on a domain. Take for example the use-case inside of neural ordinary differential equations (neural ODEs) and ODE adjoints. As mentioned above, in this use case the backwards pass is applied immediately after the forward pass. Thus while [a handwritten adjoint to a neural network layer](https://github.com/SciML/DiffEqFlux.jl/blob/v1.8.1/src/fast_layers.jl#L38-L56) can look like:
+
+```julia
+ZygoteRules.@adjoint function (f::FastDense)(x,p)
+  W = ...
+  b = ...
+  r = W*x .+ b
+  y = f.σ.(r)
+  function FastDense_adjoint(v)
+    σbar = ForwardDiff.derivative.(f.σ,r)
+    zbar = v .* σbar
+    Wbar = zbar * x'
+    bbar = zbar
+    xbar = W' * zbar
+    nothing,xbar,vcat(vec(Wbar),bbar)
+  end
+  y,FastDense_adjoint
+end
+```
+
+for `sigma.(W*x .+ b)` to calculate `J'v`, you can greatly optimize this if you know that the backwards pass will immediately preceed the forward pass. Specifically, there's no need to generate closures to store values since there no indeterminate future where the gradient might be needed, instead you can immediately proceed to calculate it. And, if you're only applying it to a vector of known size `v`, then this operation can be done non-allocating by mutating a cache vector. Lastly, if we know we're only going to use the derivative w.r.t. `x` (`xbar`), then we can eliminate many calculations. Look at the simplified version:
+
+```julia
+r = W*x .+ b
+y = σ.(r)
+σbar = derivative.(σ,r)
+zbar = v .* σbar
+Wbar = zbar * x'
+bbar = zbar
+xbar = W' * zbar   
+```
+
+and now cached:
+
+```julia
+mul!(cache,W,x)
+cache .= σ.(cache .+ b)
+cache .= derivative.(σ,cache)
+cache .= v .* cache
+mul!(xbar,W',cache)
+```
+
+or in other words, we can write this as a mutating operation with a single cache vector: `vjp!(xbar,W,b,σ,v,cache)`. All of the overheads of any automatic differentiation or mutations? Gone.
+
+Of course, building this up for anything other than the simplest case takes a much larger effort. In comes SimpleChains.jl.
+
+## SimpleChains.jl: An Optimized Machine Learning Library for SciML Use Cases
+
+SimpleChains.jl is the solution to this problem. SimpleChains.jl is a small machine learning framework optimized for quickly fitting small models on the CPU. Early development favored a design that would:
+
 1. Allow us to achieve good performance, ideally approaching the CPU's potential peak FLOPs.
-2. Focus on small size meant we could largely forgo cache tiling in early development.
-3. Have an API where vectors of parameters (and their gradients) are first class, rather than having parameters live with the layers, to make it easier to work with various optimizers or solvers that expect contiguous vectors.
+2. Focus on small size meant we could largely forgo large kernel optimizations (such as cache tiling) in the early stages of development.
+3. Have an API where vectors of parameters (and their gradients) are first class, rather than having parameters live with the layers, to make it easier to work with various optimizers or solvers that expect contiguous vectors (such as BFGS).
 4. Be written in "pure Julia" for ease of development and optimization; while making heavy use of [LoopVectorization.jl](https://github.com/JuliaSIMD/LoopVectorization.jl/), SimpleChains.jl does not rely on any BLAS or NN libraries. It is a long term aim to extend this loop-compiler approach to optimization to also producing pullbacks automatically, without requiring them to be handwritten. However, the compiler-focused approach is already levered for ease of implementation: while we still have to hand-write gradients, we do not need to hand-optimize them.
 
+## SimpleChains.jl in Action: 10x-ing PyTorch in Small Examples
 
-#### Example and Benchmark
+Let's test MNIST with LeNet5. Note that this example will be a very conservative estimate of speed because, as a more traditional machine learning use case, batching can be used to make use of matrix-matrix multiplications instead of the even smaller matrix-vector kernels. That said, even in this case we'll be able to see a substantial performance benefit because of the semi-small network sizes.
 
-Using MNIST and LeNet5:
+The following is the Julia code using SimpleChains.jl for the training:
+
 ```julia
 using SimpleChains, MLDatasets
 
@@ -74,7 +257,10 @@ SimpleChains.error_mean_and_loss(lenetloss, xtrain4, p)
 SimpleChains.error_mean_and_loss(lenetloss, xtest4, ytest1, p)
 ```
 
-Two runs of 10 epochs using PyTorch following [this script](https://github.com/chriselrod/LeNetTorch) on a V100 GPU using a batch size of 2048 produces:
+#### PyTorch
+
+Before we show the results, let's look at the competition. Here's two runs of 10 epochs using PyTorch following [this script](https://github.com/chriselrod/LeNetTorch) on a V100 GPU using a batch size of 2048 produces:
+
 ```
 Took: 17.33
 Accuracy: 0.9492
@@ -96,8 +282,10 @@ Took: 10.78
 Accuracy: 0.9841
 ```
 
-<!-- FIXME: update link once open  -->
-Flux on an A100 GPU, using [this script]():
+#### Flux.jl
+
+The standard machine learning library in Julia, [Flux.jl](https://github.com/FluxML/Flux.jl) was benchmarked using [this script]() with an A100 GPU. How fast was that?
+
 ```julia
 julia> @time train!(model, train_loader)
  74.678251 seconds (195.36 M allocations: 12.035 GiB, 4.28% gc time, 77.57% compilation time)
@@ -114,6 +302,7 @@ julia> eval_loss_accuracy(train_loader, model, device),
 ((loss = 0.0819f0, acc = 97.4967), (loss = 0.076f0, acc = 97.6))
 ```
 Flux on a V100 GPU:
+
 ```julia
 julia> @time train!(model, train_loader)
  75.266441 seconds (195.52 M allocations: 12.046 GiB, 4.02% gc time, 74.83% compilation time)
@@ -131,6 +320,7 @@ julia> eval_loss_accuracy(train_loader, model, device),
 ```
 
 Flux on an AMD EPYC 7513 32-Core Processor:
+
 ```julia
 julia> @time train!(model, train_loader)
 106.340589 seconds (67.05 M allocations: 67.053 GiB, 4.18% gc time, 29.39% compilation time)
@@ -148,6 +338,7 @@ julia> eval_loss_accuracy(train_loader, model, device),
 ```
 
 Flux on an Intel i9 10980XE 18-Core Processor:
+
 ```julia
 julia> @time train!(model, train_loader)
  70.517619 seconds (98.27 M allocations: 67.833 GiB, 3.94% gc time, 39.30% compilation time)
@@ -164,7 +355,10 @@ julia> eval_loss_accuracy(train_loader, model, device),
 ((loss = 0.072f0, acc = 97.7883), (loss = 0.0642f0, acc = 97.99))
 ```
 
+#### How long did SimpleChains.jl take?
+
 SimpleChains on an AMD EPYC 7513 32-Core Processor:
+
 ```julia
 julia> @time SimpleChains.train_batched!(G, p, lenetloss, xtrain4, SimpleChains.ADAM(3e-4), 10);
  34.410432 seconds (55.84 M allocations: 5.920 GiB, 3.79% gc time, 85.95% compilation time)
@@ -184,7 +378,9 @@ julia> SimpleChains.error_mean_and_loss(lenetloss, xtrain4, p)
 julia> SimpleChains.error_mean_and_loss(lenetloss, xtest4, ytest1, p)
 (0.9831, 0.053463124f0)
 ```
+
 SimpleChains on an Intel i9 10980XE 18-Core Processor:
+
 ```julia
 julia> @time SimpleChains.train_batched!(G, p, lenetloss, xtrain4, SimpleChains.ADAM(3e-4), 10);
  34.915917 seconds (83.99 M allocations: 5.431 GiB, 3.91% gc time, 95.22% compilation time)
@@ -204,7 +400,9 @@ julia> SimpleChains.error_mean_and_loss(lenetloss, xtrain4, p)
 julia> SimpleChains.error_mean_and_loss(lenetloss, xtest4, ytest1, p)
 (0.9829, 0.053510584f0)
 ```
+
 SimpleChains on an Intel i7 1165G7 4-Core Processor (thin and light laptop CPU):
+
 ```julia
 julia> @time SimpleChains.train_batched!(G, p, lenetloss, xtrain4, SimpleChains.ADAM(3e-4), 10);
  42.664702 seconds (103.71 M allocations: 5.236 GiB, 2.85% gc time, 77.60% compilation time)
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
microopts_blas1

ChrisRackauckas:
microopts_blas2
microopts_blas3

ChrisRackauckas:
microopts_blas1

ChrisRackauckas:
microopts_blas2

ChrisRackauckas:
microopts_blas3

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+
+<iframe width="560" height="315" src="https://www.youtube.com/embed/TFB_lt1KMto" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+
+This SciML methodology has been shown across many disciplines, from black hole dynamics to the development of earthquake safe buildings, to be a flexible method capable of discovering guiding (bio)physical equations. Here's a recent talk with walks through the various use cases of SciML throughout the sciences:
```
chriselrod, line 27:
```suggestion
This SciML methodology has been shown across many disciplines, from black hole dynamics to the development of earthquake safe buildings, to be a flexible method capable of discovering/guiding (bio)physical equations. Here's a recent talk which walks through the various use cases of SciML throughout the sciences:
```

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+
+<iframe width="560" height="315" src="https://www.youtube.com/embed/eSeY4K4bITI?start=668" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+
+For more details on the software and methods, check out this manuscript.
```
chriselrod, line 31: Was `this` supposed to be a link?

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+
+For more details on the software and methods, check out this manuscript. 
+
+However, the unique aspects of how neural networks are used in these contexts make it rife for specializing of performance. Specifically, in the context of machine learning one normally relies on the following assumption: the neural networks are large enough that the O(n^3) cost of matrix-matrix multiplication (or other kernels like convolutions) domainates the the runtime. This is essentially the guiding principle behind most of the mechanics of a machine learning library:
+
+1. Matrix-matrix multiplication scales cubicly while memory allocations scale linearly, so attempting to mutate vectors with non-allocating operations is not a high priority. Just use `A*x`.
+2. Focus on accelearting GPU kernels as fast as possible! Since these large matrix-matrix operations will be fastest on GPUs and are the bottleneck, performance benchmarks will essentially just be a measurement of how fast these specific kernels are.
```
chriselrod, line 36:
```suggestion
2. Focus on accelearting GPU kernels to be as fast as possible! Since these large matrix-matrix operations will be fastest on GPUs and are the bottleneck, performance benchmarks will essentially just be a measurement of how fast these specific kernels are.
```

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+3. When doing reverse-mode automatic differentiation (backpropogation), feel free to copy values to memory. Memory allocations will be hidden by the larger kernel calls. 
+4. Also, feel free to write a "tape" for generation backpropogation. The tape does add the cost of essentially building a dictionary during the forward pass, but that will be hidden by the larger kernel calls.
```
chriselrod, line 38:
```suggestion
4. Also, feel free to write a "tape" for generating backpropogation. The tape does add the cost of essentially building a dictionary during the forward pass, but that will be hidden by the larger kernel calls.
```
Also, optimizing activation functions isn't a high priority, because (like copying) that scales linearly.
SimpleChains will fuse certain cheap activation functions with matmul and conv kernels, e.g. `relu`.
Unlike Flux, it'll SIMD all of them.

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+Do these assumptions actually hold in our case? And if they don't, can we focus on these aspects to draw more performance out for our use cases?
+
+## Digging In: Small Neural Network Performance Overheads
+
+It's easy to show that these assumptions breakdown when we start focusing on this smaller neural network use case. For starters, lets look at assumptions (1)(2). It's not hard to show where these two will unravel:
+
+```julia
+using LinearAlgebra, BenchmarkTools, CUDA, LoopVectorization
+
+function mygemmavx!(C, A, B)
```
chriselrod, line 49:
```suggestion
function mygemmturbo!(C, A, B)
```
LoopVectorization's docs should probably be updated.

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+    end
+end
+
+function alloc_timer(n)
+    A = rand(Float32,n,n)
+    B = rand(Float32,n,n)
+    C = rand(Float32,n,n)
+    t1 = @belapsed $A * $B
+    t2 = @belapsed (mul!($C,$A,$B))
+    t3 = @belapsed (mygemmavx!($C,$A,$B))
```
chriselrod, line 65:
```suggestion
    t3 = @belapsed (mygemmturbo!($C,$A,$B))
```

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+allocgpu   = [x[4] for x in res]
+noallocgpu = [x[5] for x in res]
+
+using Plots
+plot(ns, alloc, label="*", xscale=:log10, yscale=:log10, legend=:bottomright,
+    title="Which Micro-optimizations matter for BLAS3?",
+    yticks=10.0 .^ (-8:0.5:2),
+    ylabel="Time (s)", xlabel="N",)
+plot!(ns,noalloc,label="mul! (OpenBLAS)")
+plot!(ns,noalloclv,label="mygemmavx!")
```
chriselrod, line 85:
```suggestion
plot!(ns,noalloclv,label="mygemmturbo!")
```

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+plot!(ns,allocgpu,label="* gpu")
+plot!(ns,noallocgpu,label="mul! gpu")
+savefig("microopts_blas3.png")
+```
+
+![](https://user-images.githubusercontent.com/1814174/162625573-0c195a60-dc06-44fb-af04-6c0c389ded02.png)
+
+When we get to larger matrix-matrix operations, such as 100x100 * 100x100, we can effectively write off any overheads due to memory allocations. But we definitely see that there is a potential for some fairly significant performance gains in the lower end! Notice too that these gains are realized by using the pure-Julia LoopVectorization.jl as the standard BLAS tools tend to have extra threading overhead in this region (again, not optimizing as much in this region).
```
chriselrod, line 93: Can we?
What if we ran matmuls in multiple threads, and took the mean time instead of the minimum?
Mean is much more representative of actual performance whenever the GC is involved.

But if we don't want to make an overly complicated benchmark just for the sake of demoing how our GC doesn't scale with threads, maybe at least using `mean` instead of `minimum` times in our benchmarks will be more accurate, while also helping get this point across?

The easiest way to get mean times out of BenchmarkTools I'm aware of:
```julia
mean(@benchmark rand(40)).time
```

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+plot!(ns,noallocgpu,label="mul! gpu")
+savefig("microopts_blas3.png")
+```
+
+![](https://user-images.githubusercontent.com/1814174/162625573-0c195a60-dc06-44fb-af04-6c0c389ded02.png)
+
+When we get to larger matrix-matrix operations, such as 100x100 * 100x100, we can effectively write off any overheads due to memory allocations. But we definitely see that there is a potential for some fairly significant performance gains in the lower end! Notice too that these gains are realized by using the pure-Julia LoopVectorization.jl as the standard BLAS tools tend to have extra threading overhead in this region (again, not optimizing as much in this region). 
+
+But, if you've been a riding the GPU gospel without looking into the details then this plot may be a shocker! However, GPUs are designed as dumb slow chips with many cores, and thus they are only effective on very parallel operations, such as large matrix-matrix multiplications. It is from this point that assumption (2) is derived for large newtork operations. But again, in the case of the small where such GPU kernels will be outperformed by well-designed CPU kernels due to the lack of parallel opportunities
```
chriselrod, line 95:
```suggestion
But, if you've been riding the GPU gospel without looking into the details then this plot may be a shocker! However, GPUs are designed as dumb slow chips with many cores, and thus they are only effective on very parallel operations, such as large matrix-matrix multiplications. It is from this point that assumption (2) is derived for large newtork operations. But again, in the case of small networks such GPU kernels will be outperformed by well-designed CPU kernels due to the lack of parallel opportunities.
```

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+plot!(ns,noalloc,label="mul! (OpenBLAS)")
+plot!(ns,noalloclv,label="mygemvavx!")
+plot!(ns,allocgpu,label="* gpu")
+plot!(ns,noallocgpu,label="mul! gpu")
+savefig("microopts_blas2.png")
+```
+
+![](https://user-images.githubusercontent.com/1814174/162625320-310d633a-34bf-407e-8cc9-ec55ca895d83.png)
+
+And remember, the basic operations of a neural network are `sigma.(W*x .+ b)`, and thus there's also an O(n) element-wise operation. As you would guess, this operation becomes more significant as n gets smaller and is requires even more consideration for memory operations.
```
chriselrod, line 146: Last sentence needs some rewriting.

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+```
+
+![](https://user-images.githubusercontent.com/1814174/162625320-310d633a-34bf-407e-8cc9-ec55ca895d83.png)
+
+And remember, the basic operations of a neural network are `sigma.(W*x .+ b)`, and thus there's also an O(n) element-wise operation. As you would guess, this operation becomes more significant as n gets smaller and is requires even more consideration for memory operations. 
+
+```julia
+using LinearAlgebra, BenchmarkTools, CUDA, LoopVectorization
+
+function mybroadcastavx!(C, A, B)
```
chriselrod, line 151: You could change all the `avx` to `turbo`.

I didn't want to keep answering "Is this x86 only?" questions.
(Although, the SimpleChains.jl example below is about 4x faster on my 4 core laptop than on my M1 mac mini.)

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+plot!(ns,allocgpu,label="= gpu")
+plot!(ns,noallocgpu,label=".= gpu")
+savefig("microopts_blas1.png")
+```
+
+![](https://user-images.githubusercontent.com/1814174/162625049-b26fd0fd-271a-4c73-a44d-f40b36b18136.png)
+
+This already highly motivates a project focused on the performance for this case, but assumptions (3) and (4) point us to additionally look at the implementation of the backpropogation. The [trade-off between different machine learning library approaches to automatic differentiation has already been discussed at length](https://www.stochasticlifestyle.com/engineering-trade-offs-in-automatic-differentiation-from-tensorflow-and-pytorch-to-jax-and-julia/), but what the general discussions can miss is the extra opportunities afforded when really specializing on a domain. Take for example the use-case inside of neural ordinary differential equations (neural ODEs) and ODE adjoints. As mentioned above, in this use case the backwards pass is applied immediately after the forward pass. Thus while [a handwritten adjoint to a neural network layer](https://github.com/SciML/DiffEqFlux.jl/blob/v1.8.1/src/fast_layers.jl#L38-L56) can look like:
```
chriselrod, line 191:
```suggestion
This already highly motivates a project focused on the performance for this case, but assumptions (3) and (4) point us to additionally look at the implementation of the backpropogation. The [trade-off between different machine learning libraries' approaches to automatic differentiation has already been discussed at length](https://www.stochasticlifestyle.com/engineering-trade-offs-in-automatic-differentiation-from-tensorflow-and-pytorch-to-jax-and-julia/), but what the general discussions can miss is the extra opportunities afforded when really specializing on a domain. Take for example the use-case inside of neural ordinary differential equations (neural ODEs) and ODE adjoints. As mentioned above, in this use case the backwards pass is applied immediately after the forward pass. Thus while [a handwritten adjoint to a neural network layer](https://github.com/SciML/DiffEqFlux.jl/blob/v1.8.1/src/fast_layers.jl#L38-L56) can look like:
```
Not sure about this one.

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+    Wbar = zbar * x'
+    bbar = zbar
+    xbar = W' * zbar
+    nothing,xbar,vcat(vec(Wbar),bbar)
+  end
+  y,FastDense_adjoint
+end
+```
+
+for `sigma.(W*x .+ b)` to calculate `J'v`, you can greatly optimize this if you know that the backwards pass will immediately preceed the forward pass. Specifically, there's no need to generate closures to store values since there no indeterminate future where the gradient might be needed, instead you can immediately proceed to calculate it. And, if you're only applying it to a vector of known size `v`, then this operation can be done non-allocating by mutating a cache vector. Lastly, if we know we're only going to use the derivative w.r.t. `x` (`xbar`), then we can eliminate many calculations. Look at the simplified version:
```
chriselrod, line 211:
```suggestion
for `sigma.(W*x .+ b)` to calculate `J'v`, you can greatly optimize this if you know that the backwards pass will immediately preceed the forward pass. Specifically, there's no need to generate closures to store values since there no indeterminate future where the gradient might be needed, instead you can immediately proceed to calculate it. And, if you're only applying it to a vector of known size `v`, then this operation can be done without allocating by mutating a cache vector. Lastly, if we know we're only going to use the derivative w.r.t. `x` (`xbar`), then we can eliminate many calculations. Look at the simplified version:
```

SimpleChains doesn't care whether it's known size or not, it'll avoid allocating either way.
(Unless the thing of unknown size is larger than anything it's been called with previously, in which case it'll allocate [`resize!`] to increase the size of its memory buffer to be big enough.)

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,489 @@
...
+mul!(cache,W,x)
+cache .= σ.(cache .+ b)
+cache .= derivative.(σ,cache)
+cache .= v .* cache
+mul!(xbar,W',cache)
+```
+
+or in other words, we can write this as a mutating operation with a single cache vector: `vjp!(xbar,W,b,σ,v,cache)`. All of the overheads of any automatic differentiation or mutations? Gone.
+
+Of course, building this up for anything other than the simplest case takes a much larger effort. In comes SimpleChains.jl.
```
chriselrod, line 235: Should I implement `vjp!` before the SimpleChains?

The example that follows of course is just MNIST, an extremely "small" model, in that the training data takes only around 0.175 GiB of RAM, and the number of parameters is small compared to the absurdly huge numbers we've been hearing thrown around recently for networks like GPT3, but at 44_216 parameters, it's still absurdly massive compared to things like statistical models.

ChrisRackauckas: It would be nice to specialize it in DiffEqSensitivity.jl. I guess we can wait and see what is and isn't optimized by the compiler, though to really know the effect it would take an implementation of a `vjp!` function

blog/2022/04/simplechains.md
ChrisRackauckas: Yeah that's probably worth trying.

blog/2022/04/simplechains.md
ChrisRackauckas: yeah add it?

blog/2022/04/simplechains.md
ChrisRackauckas: That's probably worth mentioning as a side note or footnote. It's hard to put into the simple explainer though.

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,493 @@
...
+```
+Took: 11.24
+Accuracy: 0.9759
+Took: 10.78
+Accuracy: 0.9841
+```
+
+#### Flux.jl
+
+The standard machine learning library in Julia, [Flux.jl](https://github.com/FluxML/Flux.jl) was benchmarked using [this script]() with an A100 GPU. How fast was that?
```
ChrisRackauckas, line 336: @chriselrod I don't have this script.

chriselrod: https://github.com/PumasAI/SimpleChains.jl/blob/35d09c4ed3f72e43d73a2c391a6bbdceeb779e21/examples/mnist_lenet.jl

chriselrod: https://github.com/PumasAI/SimpleChains.jl/blob/conv/examples/mnist_lenet.jl

chriselrod: Eventually we'd need to `s/conv/main/g`, or aim it at a specific commit.
But for now the script is only on the conv branch, which is still in the unmerged PR.

ChrisRackauckas: oof, still worth putting for now.

blog/2022/04/simplechains.md
```diff
@@ -0,0 +1,493 @@
...
+
+## Scientific Machine Learning (SciML) and "Small" Neural Networks
+
+SimpleChains.jl is a library developed by [Julia Computing](https://juliacomputing.com/) and [Pumas-AI](https://pumas.ai/) in collaboration with [Roche](https://www.roche.com/) and the [University of Maryland, Baltimore](https://www.pharmacy.umaryland.edu/centers/ctm/). The purpose of SimpleChains.jl is to be as fast as possible for small neural networks. This is because SimpleChains.jl originated as a solution for the DeepPumas product for [scientific machine learning (SciML)](https://www.stochasticlifestyle.com/the-essential-tools-of-scientific-machine-learning-scientific-ml/) in clinical pharmacology. In this domain, small neural networks are combined with known nonlinear mixed effects models (statistical models with differential equations) to discover previously unknown mechanisms and prognostic factors. For a short introduction to how this is done, check out the following video by Niklas Korsbo:
```
jgobburu, line 23: This is because SimpleChains.jl originated as a solution for the DeepPumas product for [scientific machine learning (SciML)](https://www.stochasticlifestyle.com/the-essential-tools-of-scientific-machine-learning-scientific-ml/) in healthcare data analytics.

ViralBShah:
SimpleChains.jl should first be open sourced, then this PR updated with a link, and then merged...

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,493 @@
+@def rss_pubdate = Date(2022, 4, 11)
+@def rss = """10x-ing PyTorch: Specializing Scientific Machine Learning with Julia ..."""
+@def published = "11 April 2022"
+@def title = "10x-ing PyTorch: Specializing Scientific Machine Learning with Julia"
+@def authors = """Chris Elrod, Chris Rackauckas"""
```
vjd, line 5:
```suggestion
@def authors = """Chris Elrod, Niklas Korsbo, Chris Rackauckas"""
```

chriselrod:
> SimpleChains.jl should first be open sourced, then this PR updated with a link, and then merged...

Out of curiosity, who all has the GitHub permissions to open source it?

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,493 @@
...
+allocgpu   = [x[4] for x in res]
+noallocgpu = [x[5] for x in res]
+
+using Plots
+plot(ns,alloc,label="=",xscale=:log10,yscale=:log10,legend=:bottomright,
+     title="Which Micro-optimizations matter for BLAS1?",
+     ylabel = "Time (s)", xlabel = "N",
+     yticks = 10.0 .^ (-8:0.5:2),)
+plot!(ns,noalloc,label=".=")
+plot!(ns, noalloc, label="mybroadcastturbo!")
```
korsbo, line 183: The label here does not match that of the plot.

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,493 @@
...
+
+# assess training and test loss
+SimpleChains.error_mean_and_loss(lenetloss, xtrain4, p)
+SimpleChains.error_mean_and_loss(lenetloss, xtest4, ytest1, p)
+
+
+# reinitialize to randomize parameters
+# SimpleChains.init_params!(lenet, p);
+
+# train without additional memor allocations
```
korsbo, line 297:
```suggestion
# train without additional memory allocations
```

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,493 @@
...
+
+1. Allow us to achieve good performance, ideally approaching the CPU's potential peak FLOPs.
+2. Focus on small size meant we could largely forgo large kernel optimizations (such as cache tiling) in the early stages of development.
+3. Have an API where vectors of parameters (and their gradients) are first class, rather than having parameters live with the layers, to make it easier to work with various optimizers or solvers that expect contiguous vectors (such as BFGS).
+4. Be written in "pure Julia" for ease of development and optimization; while making heavy use of [LoopVectorization.jl](https://github.com/JuliaSIMD/LoopVectorization.jl/), SimpleChains.jl does not rely on any BLAS or NN libraries. It is a long term aim to extend this loop-compiler approach to optimization to also producing pullbacks automatically, without requiring them to be handwritten. However, the compiler-focused approach is already levered for ease of implementation: while we still have to hand-write gradients, we do not need to hand-optimize them.
+
+## SimpleChains.jl in Action: 10x-ing PyTorch in Small Examples
```
korsbo, line 246: Well, the benchmarks here are not quite 10x. Some of them a fair bit less than 10x.

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,493 @@
+@def rss_pubdate = Date(2022, 4, 11)
+@def rss = """10x-ing PyTorch: Specializing Scientific Machine Learning with Julia ..."""
+@def published = "11 April 2022"
+@def title = "10x-ing PyTorch: Specializing Scientific Machine Learning with Julia"
```
korsbo, line 4: I'm not sure about the 10x claim when the benchmarks in the post does not quite support that.

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,493 @@
...
+# train
+@time SimpleChains.train_batched!(G, p, lenetloss, xtrain4, SimpleChains.ADAM(3e-4), 10);
+
+# assess training and test loss
+SimpleChains.error_mean_and_loss(lenetloss, xtrain4, p)
+SimpleChains.error_mean_and_loss(lenetloss, xtest4, ytest1, p)
+
+
+# reinitialize to randomize parameters
+# SimpleChains.init_params!(lenet, p);
```
korsbo, line 295:
```suggestion
```

korsbo (commented):
Nice writing

I'm not sure about the 10x claim since the benchmarks in here does not quite support that claim. Would we have done well enough in a enchmark on a small `Dense(Chain(5,5,tanh), Chain(5,5,tanh), Chain(5,1))` to warrant its inclusion?

blog/2022/04/simple-chains.md
ChrisRackauckas: 11.24 seconds to 1.324148 seconds?

ChrisRackauckas:
microopts_blas1
microopts_blas2
microopts_blas3

blog/2022/04/simple-chains.md
korsbo: That could be close enough for some "artistic licence" but it also looks a bit cherry-picked as the AMD benchmark is only 5x faster. 

As it stands now, I think the Julia community would not bat an eye but I'd expect a bit of eye-rolling from those feeling a little defensive of PyTorch. 10x-ing is a better punchphrase than something softer or with qualifiers though. @chriselrod would we support our case by adding a more SciML-representative benchmark of a small MLP?

ChrisRackauckas: We can go with 5x, though we know the benchmark is ultra-conservative given the showcase of how matvecs will have a much larger difference, and the benchmark is matmuls.

chriselrod: > 11.24 seconds to 1.324148 seconds?

I'll try a couple things to see if I can squeeze out another 20%.

ViralBShah:
Note that if this doesn't get merged today, the date must be changed in the blog post.

blog/2022/04/simple-chains.md
chriselrod: Okay, first thing I tried was about a 2% improvement.

My other idea will require a more radical change, which is transposing the internal data so we can SIMD across batch sizes.

chriselrod:
FWIW, on my 7980XE, this is what I get for `mygemmturbo!` vs `mul! (OpenBLAS)`:
```julia
julia> @benchmark A_mul_B!($C,$A,$B) # LoopVectorization
BenchmarkTools.Trial: 2933 samples with 1 evaluation.
 Range (min … max):  1.666 ms …  2.540 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.689 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.697 ms ± 51.168 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

      ▁▅██▆▄
  ▂▂▄▆███████▇▆▅▄▄▄▄▃▃▃▃▃▂▃▂▂▂▂▂▂▂▂▁▁▁▂▁▂▁▂▂▁▂▂▂▂▂▁▁▂▂▁▂▂▂▂▂ ▃
  1.67 ms        Histogram: frequency by time        1.82 ms <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> BLAS.set_num_threads(18);

julia> @benchmark mul!($C,$A,$B) # OpenBLAS
BenchmarkTools.Trial: 1733 samples with 1 evaluation.
 Range (min … max):  2.727 ms …  3.572 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     2.894 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   2.876 ms ± 65.728 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                         ▂▄▆██▅▇▃▃▄
  ▁▁▂▂▃▂▃▂▃▂▃▃▃▄▄▆▆▆▅▆▆▆▇▇▅▅▅▅▄▃▄▄▅▄▃▄▆▆▇███████████▆▆▄▃▃▂▂▁ ▄
  2.73 ms        Histogram: frequency by time        2.97 ms <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> log10(1.666e-3)
-2.7783250029292312
```
This would make it competitive with the GPU time from your plot.
MKL does better:
```julia
julia> using MKL

julia> @benchmark mul!($C,$A,$B)
BenchmarkTools.Trial: 3063 samples with 1 evaluation.
 Range (min … max):  1.433 ms …  2.582 ms  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     1.593 ms              ┊ GC (median):    0.00%
 Time  (mean ± σ):   1.621 ms ± 70.218 μs  ┊ GC (mean ± σ):  0.00% ± 0.00%

                            ▁▃▄▁               █▅▁
  ▂▂▂▂▂▂▁▂▁▂▂▃▂▂▂▁▁▂▂▃▇▇▇▅▅▅████▇▄▃▃▂▂▂▂▂▂▂▂▁▂▆███▆▄▃▂▂▂▂▂▁▂ ▃
  1.43 ms        Histogram: frequency by time        1.75 ms <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> log10(1.433e-3)
-2.8437538096026556
```
But I don't actually have a usable GPU on this system (or the 10980XE system)...

blog/2022/04/simple-chains.md
chriselrod: Do we have any sort of UDE type example using PyTorch, where we can show we >= 10x them?

It'd be great to have two examples:
1. The intended use case, where hopefully we vastly outperform them.
2. Then we show scaling by moving more towards their domain with the PyTorch example, where we can demonstrate that we were so far ahead in the first, we still have plenty of advantage left over with >= 4x.

chriselrod: Unfortunately, I think it'd be difficult to extract much more performance out of this example without larger changes.

SIMD-ing across batch may still be worth it.

ViralBShah:
We should not come across as bashing anything but constructive.

blog/2022/04/simple-chains.md
ChrisRackauckas: The problem is that to full specialize the UDE training on SimpleChains, SimpleChains needs to be open sourced so we can integrate it into DiffEqSensitivity.jl and such, which is a chunk of what will be done next 😅.  So I think we just have to choose a stopping point somewhere. If it's easy, try doing the spiral neural ODE (https://gist.github.com/ChrisRackauckas/4a4d526c15cc4170ce37da837bfc32c4) which is from the neural ODE paper.

korsbo: Going full-out SciML and using UDEs in the benchmark might be problematic anyways since much of the time difference could be in the DE solvers. I was more thinking of something like training an MLP to map from a vector of 10 inputs to a single output. That seems like it could be pretty "clean" in the sense that it really only benchmarks the ML tool and representative in that it's the kind of networks we often repeatedly solve in UDEs.

ChrisRackauckas:
We made a decision to fix up some docs first and go live on Thursday. Let's be ready for the HN primetime on that day.

ViralBShah:
So merge Thursday 10am ET?

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,489 @@
...
+       eval_loss_accuracy(test_loader, model, device)
+((loss = 0.1579f0, acc = 95.3583), (loss = 0.1495f0, acc = 95.54))
+
+julia> @time train!(model, train_loader)
+  1.676934 seconds (1.04 M allocations: 1.840 GiB, 5.64% gc time, 0.63% compilation time)
+
+julia> eval_loss_accuracy(train_loader, model, device),
+       eval_loss_accuracy(test_loader, model, device)
+((loss = 0.0819f0, acc = 97.4967), (loss = 0.076f0, acc = 97.6))
+```
```
GlenHertz, line 344: Would help for understanding/readability if made clear that the first run includes compile time.  It is kind of mentioned in the summary but readers may get confused if not mentioned earlier.

chriselrod: I'll update these examples tonight.
1. Timings will change to reflect the latest changes.
2. I'll add a note about compile time, and also try to improve SimpleChains'.
3. I'll switch the acc, loss order for the Flux benchmarks to reflect the order for SimpleChains and PyTorch.

chriselrod: I'll also add a small mlp example.

chriselrod:
Here is a fun MLP example (@korsbo), learning 2x2 matrix exponentials:
```julia
using SimpleChains

mlpd = SimpleChain(
  static(4),
  TurboDense(tanh, 32),
  TurboDense(tanh, 16),
  TurboDense(identity, 4)
)

function f(x)
  N = Base.isqrt(length(x))
  A = reshape(view(x, 1:N*N), (N,N))
  expA = exp(A)
  vec(expA)
end

T = Float32;
X = randn(T, 2*2, 10_000);
Y = reduce(hcat, map(f, eachcol(X)));
Xtest = randn(T, 2*2, 10_000);
Ytest = reduce(hcat, map(f, eachcol(Xtest)));

@time p = SimpleChains.init_params(mlpd);
G = SimpleChains.alloc_threaded_grad(mlpd);

mlpdloss = SimpleChains.add_loss(mlpd, SquaredLoss(Y));
mlpdtest = SimpleChains.add_loss(mlpd, SquaredLoss(Ytest));

report = let mtrain = mlpdloss, X=X, Xtest=Xtest, mtest = mlpdtest
  p -> begin
    let train = mlpdloss(X, p), test = mlpdtest(Xtest, p)
      @info "Loss:" train test
    end
  end
end

report(p)
for _ in 1:2
  @time SimpleChains.train_batched!(
    G, p, mlpdloss, X, SimpleChains.ADAM(1e-6), 1_000_000, batchsize = size(X)[end]
  );
  report(p)
end
```
vs PyTorch:
```python
import torch
import time

class MatrixExponentEstimator(torch.nn.Module):
    def __init__(self):
        super(MatrixExponentEstimator, self).__init__()
        self.d0 = torch.nn.Linear(4, 32)
        torch.nn.init.xavier_normal_(self.d0.weight)
        self.t0 = torch.nn.Tanh()
        self.d1 = torch.nn.Linear(32, 16)
        torch.nn.init.xavier_normal_(self.d1.weight)
        self.t1 = torch.nn.Tanh()
        self.d2 = torch.nn.Linear(16, 4)
        torch.nn.init.xavier_normal_(self.d2.weight)

    def forward(self, x):
        x = self.t0(self.d0(x))
        x = self.t1(self.d1(x))
        return self.d2(x)

def f(x):
    return torch.matrix_exp(x.reshape((2,2))).reshape((4,))

def apply_matrix_exponential(x):
    return torch.stack([f(x_i) for x_i in torch.unbind(x)])

def train():
    epochs = 10000
    # epochs = 1000000
    trainx = torch.randn(10000, 2*2)
    trainy = apply_matrix_exponential(trainx)
    testx = torch.randn(10000, 2*2)
    testy = apply_matrix_exponential(testx)

    model = MatrixExponentEstimator()
    adam = torch.optim.Adam(model.parameters(), lr = 1e-3)
    loss_fn = torch.nn.MSELoss()
    print('Initial Train Loss: {:.4f}'.format(loss_fn(model(trainx), trainy)))
    print('Initial Test Loss: {:.4f}'.format(loss_fn(model(testx), testy)))
    for _ in range(2):
        t_start = time.time()
        for _ in range(epochs):
            adam.zero_grad()
            loss_fn(model(trainx), trainy).backward()
            adam.step()
        print('Took: {:.2f} seconds'.format(time.time() - t_start))
        print('Train Loss: {:.4f}'.format(loss_fn(model(trainx), trainy)))
        print('Test Loss: {:.4f}'.format(loss_fn(model(testx), testy)))
                
if __name__ == '__main__':
    train()
```
I get
```julia
julia> report(p)
┌ Info: Loss:
│   train = 0.0033560519f0
└   test = 0.034497753f0

julia> for _ in 1:2
         @time SimpleChains.train_batched!(
           G, p, mlpdloss, X, SimpleChains.ADAM(), 10_000, batchsize = size(X)[end]
         );
         report(p)
       end
  0.819090 seconds (4 allocations: 80 bytes)
┌ Info: Loss:
│   train = 0.0029986524f0
└   test = 0.0280362f0
  0.781178 seconds (4 allocations: 80 bytes)
┌ Info: Loss:
│   train = 0.0015479826f0
└   test = 0.02393272f0
```
(note this already compiled the first train), vs
```julia
Initial Train Loss: 6.2453
Initial Test Loss: 6.6992
Took: 19.05 seconds
Train Loss: 0.0058
Test Loss: 0.1240
Took: 18.99 seconds
Train Loss: 0.0014
Test Loss: 0.0960
```
so that puts us at 25x faster for this example.

Shrinking the data set increases our advantage. E.g., 10x smaller dataset but 10x more epochs, and SimpleChains takes about 1.1 seconds while PyTorch takes around 60s.
However, less data means we overfit the test data here. I could try combating it with regularization, but I figured increasing the training data set size is reasonable.

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,489 @@
...
+(0.9841, 0.048268355f0)
+
+julia> @time SimpleChains.train_batched!(G, p, lenetloss, xtrain4, SimpleChains.ADAM(3e-4), 10);
+  6.320631 seconds
+
+julia> SimpleChains.error_mean_and_loss(lenetloss, xtrain4, p)
+(0.99485, 0.017716348f0)
+
+julia> SimpleChains.error_mean_and_loss(lenetloss, xtest4, ytest1, p)
+(0.9888, 0.035394628f0)
```
chriselrod, line 464:
```suggestion
julia> @time SimpleChains.train_batched!(G, p, lenetloss, xtrain, SimpleChains.ADAM(3e-4), 10);
 41.053800 seconds (104.10 M allocations: 5.263 GiB, 2.83% gc time, 77.62% compilation time)

julia> SimpleChains.accuracy_and_loss(lenetloss, xtrain, p),
       SimpleChains.accuracy_and_loss(lenetloss, xtest, ytest, p)
((0.9491333333333334, 0.16993132f0), (0.9508, 0.15890576f0))

julia> @time SimpleChains.train_batched!(G, p, lenetloss, xtrain, SimpleChains.ADAM(3e-4), 10);
  5.320512 seconds

julia> SimpleChains.accuracy_and_loss(lenetloss, xtrain, p),
       SimpleChains.accuracy_and_loss(lenetloss, xtest, ytest, p)
((0.9700833333333333, 0.10100537f0), (0.9689, 0.09761506f0))
```

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,489 @@
...
+```
+
+Note that smaller batch sizes improve accuracy per epoch, and batch sizes were set to be proportional to the number of threads.
+
+## Benchmark Summary
+
+Latency before the first epoch begins training is problematic, but SimpleChains.jl is fast once compiled.
+Post-compilation, the 10980XE was around 70% faster than Flux with the A100 GPU.
+The 1165G7, a laptop CPU featuring AVX512, was competive, handily trouncing any of the competing machine learning libraries when 
+they were run on CPU, and even beat PyTorch on a V100. Again, we stress that this test case followed the more typical machine learning
```
chriselrod, line 474:
```suggestion
they were run on far beefier CPUs, and even beat PyTorch on both the V100 and A100. Again, we stress that this test case followed the more typical machine learning
```

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,489 @@
...
+julia> SimpleChains.error_mean_and_loss(lenetloss, xtest4, ytest1, p)
+(0.9888, 0.035394628f0)
+```
+
+Note that smaller batch sizes improve accuracy per epoch, and batch sizes were set to be proportional to the number of threads.
+
+## Benchmark Summary
+
+Latency before the first epoch begins training is problematic, but SimpleChains.jl is fast once compiled.
+Post-compilation, the 10980XE was around 70% faster than Flux with the A100 GPU.
```
chriselrod, line 472:
```suggestion
Post-compilation, the 10980XE was competitive with Flux using an A100 GPU, and about 35% faster than the V100.
```
I realized the loaders weren't putting the data onto the GPU up front. Doing this improved the Flux GPU times (in the suggestions above), so my advantage is smaller.

But that makes Flux look quite good compared to PyTorch on the GPU, which was oddly slow.
This also suggests that PyTorch's problem isn't that the data wasn't already on the GPU, as Flux was many times faster when experiencing that same problem.

chriselrod:
I've updated benchmarks and added comments reflecting the latest revisions to SimpleChains.

I also added a section with a smaller matrix exponential example, where the 10980XE is 35x faster than PyTorch, and the Epyc 7513 is around 100x faster.
@ChrisRackauckas, you may want to merge then edit that (as well as other) comments to more smoothly fit the narrative.
But I think it supports the narrative, as we show that our advantage gets increasingly extreme for these small problems -- and, with a data set of 10_000, this example still isn't all that small compared to some of our use cases in SciML!
The MNIST example then shows it continues to work well as problems grow to the size where we switch to the GPU, probably pushing that size up quite a bit. Of course, the V100 and A100 both cost several times more money than the 10980XE, so SimpleChains is still well ahead in compute/$ for MNIST.
Maybe some folk consider MNIST a toy problem -- it certainly is compared to some of the massive transformers running around -- but I think this shows we're good for a large range of problem sizes.

Someone will also need to [approve this PR](https://github.com/PumasAI/SimpleChains.jl/pull/48) so that it can be merged and we can tag a new release for this blog post.
I used that branch for all updated benchmarks.

blog/2022/04/simple-chains.md
chriselrod: My poor laptop has a hard time taking the heat, runs out of power budget, or something that causes it to begin throttling.
So I didn't copy/paste the script as one giant block, but waited a few seconds before running the second `@time SimpleChains.train_batched!` to let it cool off. Might be worth mentioning this.

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,672 @@
...
+
+## Conclusion
+
+There are many things that can make a library achieve high-performance, and nothing is as essential as knowing how it will be used. While the big machine learning frameworks have done extremely well focusing on the top-notch performance for 99.9% of their users, one can still completely outclass them when focusing on some of the 0.1% of applications which fall outside of what they have been targeting. This is the advantage of composability and flexibility: a language that allows you to easily build a machine learning framework is also a language which allows you to build alternative frameworks which are optimized for alternative people. SimpleChains.jl will not be useful to everybody, but it will be extremely useful to those who need it.
+
+## Opportunities
+
+If you are interested in this topic and want to work on Julia and machine learning, note that the [DeepPumas team is hiring](https://pumas.ai/company/machine-learning-scientist/). Additionally, [SciML internships are also available at Julia Computing](https://jobs.juliacomputing.com/jobs/ndvlJz9fHYcr/machine-learning-intern-remote).
```
korsbo, line 672:
```suggestion
If you are interested in this topic and want to work on Julia and machine learning, note that the DeepPumas team is hiring ([here](https://pumas.ai/company/machine-learning-scientist/) and [here](https://pumas.ai/company/scientist-deeppumas/)). Additionally, [SciML internships are also available at Julia Computing](https://jobs.juliacomputing.com/jobs/ndvlJz9fHYcr/machine-learning-intern-remote).
```

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,672 @@
...
+    t5 = @belapsed CUDA.@sync(mul!($C,$A,$B))
+    t1,t2,t3,t4,t5
+end
+ns = 2 .^ (2:11)
+res = [alloc_timer(n) for n in ns]
+alloc      = [x[1] for x in res]
+noalloc    = [x[2] for x in res]
+noalloclv  = [x[3] for x in res]
+allocgpu   = [x[4] for x in res]
+noallocgpu = [x[5] for x in res]
```
GlenHertz, line 77:
```suggestion
alloc      = [t[1] for t in res]
noalloc    = [t[2] for t in res]
noalloclv  = [t[3] for t in res]
allocgpu   = [t[4] for t in res]
noallocgpu = [t[5] for t in res]
```

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,672 @@
...
```
GlenHertz, line 130:
```suggestion
alloc      = [t[1] for t in res]
noalloc    = [t[2] for t in res]
noalloclv  = [t[3] for t in res]
allocgpu   = [t[4] for t in res]
noallocgpu = [t[5] for t in res]
```

blog/2022/04/simple-chains.md
```diff
@@ -0,0 +1,672 @@
...
+    t5 = @belapsed CUDA.@sync($C .= $A .* $B)
+    t1,t2,t3,t4,t5
+end
+ns = 2 .^ (2:11)
+res = [alloc_timer(n) for n in ns]
+alloc      = [x[1] for x in res]
+noalloc    = [x[2] for x in res]
+noalloclv  = [x[3] for x in res]
+allocgpu   = [x[4] for x in res]
+noallocgpu = [x[5] for x in res]
```
GlenHertz, line 175:
```suggestion
alloc      = [t[1] for t in res]
noalloc    = [t[2] for t in res]
noalloclv  = [t[3] for t in res]
allocgpu   = [t[4] for t in res]
noallocgpu = [t[5] for t in res]
```

PR #1666, opened 2022-04-14
Add JuliaStats projects

ParadaCarleton:
@ajaynshah @ViralBShah First draft for GSOC proposal. Thoughts?

@ajaynshah, could you complete the rest of the description so we can get this up sometime tomorrow? I've left the following blank, since I'm lacking info:
1. The CRRao.jl description is incomplete since I still don't know much about it or what contributing to it would involve.
2. Some mentorships are blank, since I'm not sure which mentors would be interested in what.


ViralBShah:
CRRao.jl should be linked into this PR before we merge.

ParadaCarleton:
@ViralBShah I've made some more edits and I believe it's ready to merge, since we can always make more changes later if we need to.

PR #1671, opened 2022-04-21
Update ETHZ course infos

luraess:
Add new custom URL and new Fall 2022 edition.


logankilpatrick:
Using the same structure as the MIT course, I love it!

PR #1672, opened 2022-04-22
Add Mailchimp sign up form to footer

logankilpatrick:
This was a lot of work but I got it working and looking in an acceptable way. Not sure we will have an actual newsletter but it makes sense to capture emails for things like JuliaCon blasts, sponsorship, etc.


logankilpatrick:
Only part I would change if I could would be that the success message shows up on the left side, which seems odd, but that that point we already have the email saved so it shouldn't matter much

PR #1673, opened 2022-04-22
Better link

PallHaraldsson:
[skip ci]


PallHaraldsson:
I noticed the outdated number (came straight from Google), but I can't locate if this page is actually linked from the site itself anymore. So I went looking and found the other problem I fixed. It's at least harmless to merged the latter commit, and I guess very preferred for the former...

logankilpatrick:
thanks!

PR #1678, opened 2022-05-10
move benchmarks plotting to JuliaLang/Microbenchmarks

acxz:
move benchmarks SVG file to images folder

update links to benchmarks webpage

This is a follow up to https://github.com/JuliaLang/Microbenchmarks/pull/62

The benchmark scripts and data has been moved over to https://github.com/JuliaLang/MicroBenchmarks and these are now redundant here.
Only the image is kept of the final graph which is moved over to where the rest of the images are.

The benchmarks.md page is updated with permalinks of where the older data and scripts reside.


logankilpatrick:
This looks good to me, can you also confirm @ViralBShah

acxz:
@logankilpatrick there seems to be an error. The actual website https://julialang.org/benchmarks/ is not displaying the svg file but the markdown is https://github.com/JuliaLang/www.julialang.org/blob/main/benchmarks.md.

Maybe the link needs to be `/assets/images/benchmarks.svg` instead of `/_assets/images/benchmarks.svg`. I assume this will not show the svg file in markdown but will show it properly on website. Can you confirm/roll out a quick fix to the website?

logankilpatrick:
Fixed this just now, the issue is that the _assets gets converted to assets when the site is built so the URL path was wrong yet still renders on GitHub.

acxz:
thx for the quick response!

PR #1680, opened 2022-05-25
Set stable release to 1.7.3, comment out macOS M1

ararslan:
As usual the date in the config file is set to match the REPL banner rather than today's date.

We don't have macOS M1 binaries for this release.


ararslan:
Build segfaulted and since 1.7 has known serious issues on M1 we decided to just omit it for 1.7.3. It's much more stable on 1.8.

PR #1687, opened 2022-06-23
Clean README and a few other webpages

pitmonticone:


logankilpatrick:
These are great fixes, thank you very much!!

PR #1689, opened 2022-06-24
Add new Julia Job Board

logankilpatrick:


ChrisRackauckas (approved):

logankilpatrick:
Hey @tlienart is the build error a blocker? It runs fine locally for me.

giordano:
There is no link to https://julialang.org/jobs/ from anywhere else on the website, how people should get there?

giordano:
Quick feedback: looks nice, but the process of submitting new ads consists of opening an issue here?  That sounds like quite some noise for people following this repository (and also a bit unusual).  I believe a dedicated repository would be better (so people interested in job ads can also follow that one).  Related, there is also this job ad posted in Slack this week: https://elxw.fa.em3.oraclecloud.com/hcmUI/CandidateExperience/en/sites/CX_1001/job/4521

logankilpatrick:
Yeah, that's fair, I don't suspect it will be much more noise given the volume of Julia jobs is not that high but @ViralBShah is not in favor of us hosting a job board

ViralBShah:
My main worry is that there is already a jobs listing in discourse. Having a separate board here means people posting jobs have to not only open but also close filled jobs and such. And not having enough jobs being posted makes it look a bit like a ghost town. Not to mention that when it gets popular, it will get spammed with all kinds of other jobs (and other stuff). 

My suggestion is we link to linkedin and indeed and such sites that have julia in the job description, and people hiring for specific julia jobs post them to discourse. On discourse, we should certainly make the jobs section first class and visible.

StefanKarpinski:
I agree that having a random smattering of jobs on the website seems not ideal. We have jobs on Discourse and that seems to be working well enough.

ViralBShah:
I can't revert this PR through the UI. I'll get around to deleting this later unless someone beats me to it.

ChrisRackauckas:
The discourse is definitely not working well enough because it seems a good chunk of companies prefer to send me an email instead. The issue is that with Discourse there is no running tab on which opportunities are still open. It needs some form of an editable wiki

ViralBShah:
No companies will actually manage our job board and close listings. If they email you, ask them to post their job posting link to discourse. Ideally the position they link to will be closed by them if it fills and is reflected on their own job portal.

Are we expecting a company's HR to edit some wiki or board and maintain the pages? I don't see that working. I can't even imagine who maintains all this stuff, makes accounts, deals with spam, and fixes things manually etc.

ViralBShah:
Discourse links to the opportunities. People have to follow the links to see if they are open. I don't see the need to do more than that.

ChrisRackauckas:
Of course the company's HR won't edit it, that's not how job wikis work. This is something that is very common across different fields that have very specialized requirements. The mathjobs wiki for example is extremely successful: http://notable.math.ucdavis.edu/wiki/Mathematics_Jobs_Wiki , but there are many others https://www.astrobetter.com/wiki/tiki-index.php?page=Rumor+Mill . . Job postings get up there in a day, and if you mention one is filled the details get changed in a day. It's a system where everyone knows that it "should" be up to date, and contribute to help it stay maintained. I would think that taking a page from the book of how the good posting sites are managed could be helpful, especially since the ones which tend to have the worst reputation (the econ jobs one 😅) tend to be done through forum sites.

PR #1713, opened 2022-08-05
Put the 2022 playlist on the landing page.

ViralBShah:
The video window looks really weird. @logankilpatrick @shashi Do either of you know how to fix it? BTW, it looks equally ugly on julialang.org now too.

Also, it no longer shows a random JuliaCon video and just the one. Which is quite terrible. Can we make the old thing work somehow?

### DIFF
--- a/index.html
+++ b/index.html
@@ -11,6 +11,7 @@
 
 <body>
 
+  <!--
   <div class="container-fluid" id="top-alert">
     <div class="alert alert-dark alert-dismissible mb-0" role="alert">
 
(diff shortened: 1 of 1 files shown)


ViralBShah:
How do we make this pick a random video, or does it already do that?

ViralBShah:
The video window looks really weird. @logankilpatrick @shashi Do either of you know how to fix it? BTW, it looks equally ugly on julialang.org now too.

ViralBShah:
I'm merging this one for now, but we should do a follow-up PR to fix.

PR #1715, opened 2022-08-07
Remove the link to `status.julialang.org` from the footer

DilumAluthge:
`status.julialang.org` has frequent downtime. It's not up reliably enough for us to advertise it on the Julia website.


DilumAluthge:
Once `status.julialang.org` has more reliable uptime, we can add it back to the footer.

DilumAluthge:
cc: @staticfloat

PR #1730, opened 2022-08-16
add the julia 1.8 release highlight blog post

KristofferC:
Free for all for people to add stuff.

PR #1732, opened 2022-08-18
Set stable release to 1.8.0

ararslan:
- Move 1.7.3 to old releases
- Uncomment macOS ARM in stable download table
- Remove the "(experimental)" note from macOS ARM
- Comment Linux ARMv7 in stable download table

Should probably be merged in tandem with #1730?

### DIFF
--- a/config.md
+++ b/config.md
@@ -10,9 +10,9 @@
 @def author = ""
 
 <!-- Templating of the Downloads -->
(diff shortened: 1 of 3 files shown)


PallHaraldsson:
Should ROCM (and maybe also oneAPI) get higher tier? I understand it's stable, and far better than just one year ago.

PR #1734, opened 2022-08-18
Missing text at the end of a paragraph

yakir12:
Hope I got the gist right...?


logankilpatrick (approved):
LGTM, @KristofferC can you confirm this as well?

KristofferC:
Thanks, I also fixed this in https://github.com/JuliaLang/www.julialang.org/pull/1736.

PR #1739, opened 2022-08-18
Over 8100 registered packages going with "8000"

PallHaraldsson:
Feel free to use 8100 number (or 8000, as at Juliacomputing.com, where I got the other number), if you want up update 10x more often. Otherwise will be "correct" but misleading, and I don't plan to make PRs that often.

### DIFF
--- a/index.html
+++ b/index.html
@@ -386,7 +386,7 @@ end
 
       <br>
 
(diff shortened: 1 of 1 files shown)


logankilpatrick:
Thanks!

PR #1749, opened 2022-09-10
Set stable release to 1.8.1

ararslan:


DilumAluthge:
To reduce confusion, I would recommend that in the future, we merge the website PR at the exact same time that we post the announcement on Discourse. See also my comment here: https://discourse.julialang.org/t/unable-to-download-julia-1-8-1/87027/16?u=dilumaluthge

cc: @KristofferC @ararslan

PR #1750, opened 2022-09-11
On the Downloads page, distinguish between the Git tag date and the official release announcement date

DilumAluthge:
Users often confuse the date of the Git tag with the date of the official release announcement. For a recent example, see [this Discourse thread](https://discourse.julialang.org/t/unable-to-download-julia-1-8-1/87027). However, I have seen this same confusion multiple times in the past.

Therefore, on the downloads page, we should explicitly distinguish between the Git tag date and the official release announcement date.

### DIFF
--- a/config.md
+++ b/config.md
@@ -12,10 +12,12 @@
 <!-- Templating of the Downloads -->
 @def stable_release = "1.8.1"
 @def stable_release_short = "1.8"
-@def stable_release_date = "September 6, 2022"
+@def stable_release_tag_date = "September 6, 2022" # Date that the Git tag was created on GitHub
+@def stable_release_announce_date = "September TBA, 2022" # Date that the announcement was posted in the "Announcements" category on Discourse
 @def lts_release = "1.6.7"
 @def lts_release_short = "1.6"
-@def lts_release_date = "July 19, 2022"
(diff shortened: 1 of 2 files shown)


ararslan:
Rather than manually tracking two separate dates, could we just have a prominent note near the top of the downloads page that explains what the dates listed with the versions are? I'm not sure that the announced date is all that useful in practice.

ViralBShah:
Two separate dates feels unnecessarily complex. I agree with @ararslan here - just articulate it clearly on the page.

logankilpatrick:
@DilumAluthge what is the status of this PR? Do we still want to do it?

LilithHafner:
The best solution is to have all the timestamps say the same date.

For 1.9.4, the github release "created_at" and "published_at" dates are
```
  "created_at": "2023-11-14T08:46:03Z",
  "published_at": "2023-11-14T18:33:20Z",
```
([source](https://api.github.com/repos/JuliaLang/julia/releases/129409016))

The [Discourse post](https://discourse.julialang.org/t/julia-v1-9-4-has-been-released/106230) is from 14th, and the website lists the 14th.

So, looks like we're all good here.

PR #1766, opened 2022-10-17
Promote `x86_64-linux-musl` from tier 3 to tier 2

DilumAluthge:


DilumAluthge:
This can be merged now. The `build` portion of x86_64-linux-musl has been stably green for a while now.

PR #1767, opened 2022-10-17
Demote i686-linux-gnu from tier 1 to tier 2

DilumAluthge:


ViralBShah:
It would be nice to keep this at tier 1 as long as possible.

PR #1770, opened 2022-10-30
Cleaner downloads table

IanButterworth:
Just an idea. Treats it more like a table, removes repetition.

The other tables need updating if approved of

This PR

Screen Shot 2022-10-30 at 12 05 08 PM

vs main

Screen Shot 2022-10-30 at 12 05 59 PM


IanButterworth:
Not sure what's going on with CI but this is a local render

Screen Shot 2022-10-30 at 12 05 08 PM

vs main

Screen Shot 2022-10-30 at 12 05 59 PM

DilumAluthge (approved):
I think we should preserve the text `aarch64` (for Linux ARM 64-bit), but otherwise LGTM.

logankilpatrick:
@IanButterworth this looks good to me, I am happy to proceed with a merge if you think it's good to go.

IanButterworth:
The other tables need updating before merge.

logankilpatrick:
One piece of feedback, maybe we can add a sentence somewhere that says what "archive" means in this context, I genuinely don't know what it is.

logankilpatrick:
Are you willing/able to do that as well? If not, I can take a stab this week sometime.

logankilpatrick:
@IanButterworth Are we good to merge this?

IanButterworth:
> The other tables need updating before merge.

Please feel free to do this and push here. I think they should all match in format.

ViralBShah:
I agree we should not use the word `archive`. Most people don't know what it means.

logankilpatrick:
I can take this PR over the finish line, but what do we want "Archive" to say? It is unclear to me what that means here.

ViralBShah:
I suggest just updating this and we can wordsmith later.

PR #1771, opened 2022-11-01
promote Win32 back to tier 1

vtjnash:
We have not had CI issues in a fairly long time[^1], and we do expect it to work and be tested at all times still.

[^1]: https://github.com/JuliaLang/www.julialang.org/commit/55ff318eec7e09e269a3229508e6621f14f9d004

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -316,7 +316,7 @@ Different OSes and architectures have varying [tiers of support](/downloads/#sup
     </tr>
     <tr>
       <td> i686 (32-bit) </td>
-      <td> <font color="orange">Tier 2</font> </td>
+      <td> <font color="green">Tier 1</font> </td>
     </tr>
     <tr>
       <td> 8.1+ </td>


KristofferC:
CI for windows on 1.8 is borken: https://github.com/JuliaLang/julia/pull/46984

vtjnash:
That is https://github.com/JuliaLang/julia/issues/46940

KristofferC:
Is that really on the 1.8 branch?

vtjnash:
No, that is on master (and i686-linux-gnu), and indeed, also needs to reapply a PR to reproduce the issue. But it appears to be the same, although we were not running win32 CI on buildkite yet at the time of that failure for comparison.

DilumAluthge:
This shouldn't be merged until https://github.com/JuliaCI/julia-buildkite/pull/226 has been merged.

KristofferC:
Just pointing out, that the one who bumps a platform to tier 1 is also signing up to fix any issue that occurs in the future that is unique to that platform :).

ViralBShah:
I feel it is best to leave it at Tier 2...

vtjnash:
It passes as much as other tier 1 platforms (Win64 + linux32), since it shares pretty much all code and issues with one of those 2. We can demote it again just as easily if that ceases to be true.

KristofferC:
> We can demote it again just as easily if that ceases to be true.

What? A tier is not a reflection of the current state of CI. It is a commitment saying that if CI drifts away from what the tier list states then we will not do a release until CI has been fixed.

ViralBShah:
I don't think it is a good idea to promote win32 up again. See-sawing it up and down the tiers is not a good look.

vtjnash:
We can revert. I just didn't want this to conflict with the Win8 PR. And I do think it is fairly stable on CI and still useful to make releases for it

Issue #1772, opened 2022-11-01
screenhot (old startup message) 


spaette:
_assets/images/julia-gnome.png

**Version 0.3.1**

The above file is not linked to in any `JuliaLang/www.julialang.org` files.

However `JuliaLang/julia` contains the file contrib/julia.appdata.xml which links to a same named .png.

**Version 1.8.2**

![julia-gnome](https://user-images.githubusercontent.com/111918424/199318502-b4fe87f7-e9c4-45c2-a12e-a0953590a0e2.png)

There's the Draft PR `#47390` over on `JuliaLang/julia` which among other matters fixes a bad url to julia-gnome.png.

PR #1773, opened 2022-11-02
image file (update)

spaette:
https://github.com/JuliaLang/www.julialang.org/issues/1772

replace Version 0.3.1 gnome-terminal screenshot with Version 1.8.2 gnome-terminal screenshot

a file of the name julia-gnome.png is listed in the julia repository's contrib/julia.appdata.xml file


spaette:
`Type "?" for help, "]?" for Pkg help.`

What's the purpose of the second `?` in the startup message?

ViralBShah:
The second `?` is for help on the package manager.

Issue #1675, opened 2022-05-01
The "Documentation" button on the home page should link to the "Learn" tab


CameronBieganek:
New users are getting sent straight to the manual instead of being directed towards tutorials. I propose that the big blue "Documentation" button link to the "Learn" tab of the page. Or alternatively, we could add a second button, labeled "Learn" or "Tutorials", between the "Download" and "Documentation" buttons to more prominently highlight the alternative learning resources.

xref:
https://discourse.julialang.org/t/blog-post-about-my-experiences-with-julia/79976/168?u=cameronbieganek

ViralBShah:
Learn can be after immediately after Documentation, but shouldn't be before. Please submit a PR for that if you like that.

ViralBShah:
> Learn can be after immediately after Documentation, but shouldn't be before. Please submit a PR for that if you like that.

Done in #1705

CameronBieganek:
@ViralBShah  Please reopen. I think there was a misunderstanding about the intent of this issue. Re-reading my original post, I guess it's not super clear, but what I would like to do is add a third button that says "Learn", which would be located to the right of the blue "Documentation" button in the following screenshot:

<img width="809" alt="image" src="https://user-images.githubusercontent.com/8310743/179429113-abafd248-c36b-4d8d-b36d-5f9bf47384c0.png">

I don't know anything about web programming, so I can't really submit a PR.

ViralBShah:
@CameronBieganek I disagree that we should have a Learn button there. I made it closer in the navbar to make it slightly more prominent.

However, happy to leave this open for people to chime in.

sylvaticus:
Yep, I obviously agree with the op... 😁
I believe "documentation" is too much confusing for casual, first time explorer users... not to consider it sends to a totally different website...
I have made a pr to _replace_ documentation with learn.
In my opinion experienced users would still know how to reach the documentstion from the link bar at the top.

logankilpatrick:
Closing this for now, I think having the documentation link as is should be the right thing to do. I will open a PR in the main Julia docs to add a link back to the learning section on the site.

PR #1775, opened 2022-11-08
Changed second main link from "Documentation" to "Learn"

sylvaticus:
https://discourse.julialang.org/t/suggestion-change-documentation-to-learn-as-the-second-main-button-on-julialang-org/89929

### DIFF
--- a/index.html
+++ b/index.html
@@ -39,7 +39,7 @@
       <p>
       <a class="btn btn-success btn-lg" href="/downloads/" role="button">Download</a>
       &nbsp;
-      <a class="btn btn-primary btn-lg" href="https://docs.julialang.org" role="button">Documentation</a>
+      <a class="btn btn-primary btn-lg" href="/learning/" role="button">Learn</a>
       <span class="btn float-md-right">
         <a class="github-button" href="https://github.com/JuliaLang/julia" data-icon="octicon-star" data-size="large" data-show-count="true" aria-label="Star JuliaLang/julia on GitHub">Star</a>
       </span>


jd-foster:
This is a variant/duplicate of #1675.
FWIW, I agree with your suggestion, but to have Documentation as the third button.

logankilpatrick:
I have a hard time seeing how moving the order on these links will actually help users. See some other sites for examples of this similar approach:
- https://numpy.org
- https://jupyter.org

I don't think this will move the needle at all IMO

logankilpatrick:
Learn is directly next to the docs tab, so users are as close as they can get!

sylvaticus:
Up to you, but in both the links you posted, when one click on the "main button" ("get started" in Numpy) or ("Install Jupyter hub") in Jupyter will have:
- a page with the same navigation bar, footer and style as the home page and rest of the site
- a page with toward actionable practical instruction to start using the  "product"

I just think that currently the (new) user is lost when she click on the second link.

image

(secondary problems: (a) no link back to the home page, (b) column should be centered)

Perhaps my problem is only aesthetic, but .. it just looks odd :-|

logankilpatrick:
I added something to this effect here: https://github.com/JuliaLang/julia/pull/47646

logankilpatrick:
And note that I also have an open issue to make use of the extra area on the right side.

logankilpatrick:
I light of the other PR's we have open, I am going to close this. Open to more discussion around how we can get folks to the learning resources faster but it just doesn't seem like this would specifically be the right move. Thanks for the thought on this!

PR #1780, opened 2022-11-17
Set stable version to 1.8.3

Nyoho:
I would like to download 1.8.3 from the download page https://julialang.org/downloads/

### DIFF
--- a/config.md
+++ b/config.md
@@ -14,9 +14,9 @@
 NOTE: When updating for a new release, make sure to also rerun
 `downloads/oldreleases.jl`
 -->
-@def stable_release = "1.8.2"
+@def stable_release = "1.8.3"
 @def stable_release_short = "1.8"
-@def stable_release_date = "September 29, 2022"
+@def stable_release_date = "November 15, 2022"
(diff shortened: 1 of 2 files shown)


Nyoho:
In spite of using `oldreleases.jl`, many lines in oldrelease.md were somehow updated.

DilumAluthge (changes requested):
Just FYI, Julia 1.8.3 has not actually been released yet. We'll post an announcement on Discourse when the release has officially been made.

Nyoho:
I see. Does it mean that this release https://github.com/JuliaLang/julia/releases/tag/v1.8.3 is not the official release? Anyway, I'll wait the post of Discourse. Thank you.

KristofferC:
>  Does it mean that this release https://github.com/JuliaLang/julia/releases/tag/v1.8.3 is not the official release?

It is the tag for the official release but there is some time between that and binaries getting created and put in their correct place etc.

Nyoho:
I completely understood! Thank you for explanation.

PR #1788, opened 2022-12-05
add "Julia for Data Analysis" book

bkamins:


logankilpatrick:
Thanks and congrats again!

PR #1794, opened 2022-12-23
Remove the RL part

findmyway:
Removed the RL part due to that I may not have enough time to mentor students this summer.


logankilpatrick:
Thanks for the quick follow up!

PR #1801, opened 2023-01-03
Update Documenter JSOC projects

mortenpi:
Removing two of the projects, since (1) for LaTeX a lot of the improvements have already been done, and (2) the docsystem API is not a good project for someone who is not already familiar with the docsystem etc.


logankilpatrick:
Thank you!

PR #1804, opened 2023-01-06
Update juliastats.md

ParadaCarleton:
I've updated the JuliaStats GSOC to add extra topics. (I've contacted the mentors listed here to see if they're interested in helping again, but haven't heard back yet.)


ViralBShah:
I believe there is broad interest from XKDR folks, who I met with recently. Let's merge and update again later as necessary.

PR #1806, opened 2023-01-12
Create GSOC description of Trixi.jl

ranocha:
Here is our proposal from the [trixi-framework](https://github.com/trixi-framework). I hope it's not too late (but the GSOC deadline itself seems to be fine).

It would be great if you could tell us whether you need anything else, @logankilpatrick 

CC @sloede


ranocha:
> I believe trixi may need a mention here as well: https://julialang.org/jsoc/projects/

Thanks! I did that in [1dc2d38](https://github.com/JuliaLang/www.julialang.org/pull/1806/commits/1dc2d38073e45938c9920fe0ac6dddf1390b5399)

PR #1765, opened 2022-10-17
Once Julia 1.9 is released, promote `aarch64-apple-darwin` from tier 2 to tier 1

DilumAluthge:
Please do not merge this PR until Julia 1.9 has been released.

PR #1814, opened 2023-01-22
M-series no longer experimental in 1.9

ViralBShah:


DilumAluthge:
See also: https://github.com/JuliaLang/www.julialang.org/pull/1765

ViralBShah:
Right, I wanted to mark that the beta is not experimental to make sure it doesn't discourage folks from trying.

DilumAluthge (approved):

PR #1820, opened 2023-02-01
GSoC: update JuliaImages project list

johnnychen94:
Based on the announcement in slack:

image

cc: @timholy


timholy:
I'm still mulling over participating. The benchmarks thing is something we need and which I'd be likely to work on myself, so if we get a good proposal then perhaps it makes sense to mentor.

johnnychen94:
Okay. I'll update the pages accordingly tomorrow. We'll get only two potential projects "benchmarking" and "GPU", will it be okay to have you @timholy as the mentor?

timholy:
OK, I've officially decided I would be willing to mentor the benchmarking project. I think that really is the top priority for me, so the GPU project would at most be an add-on once we have robust benchmarking.

ViralBShah:
We probably need to update the project descriptions - can reope

johnnychen94:
I've chosen to only keep the benchmark project idea here in the list.

The projects ImageIO, EXIF and QRCoders are almost done in GSoC'22/OSPP'22.
GPU project is probably too vague to figure out a doable path so I removed it. We may want to restrict it into a narrower landscape, for instance "Improving ImageFiltering for performance"

ViralBShah:
Is it good to merge?

timholy:
Thanks @johnnychen94 !

PR #1827, opened 2023-02-06
information_on_Julia_courses

pszufe:
Update the list by:
1. recent course at MIT
2. another course being taught at SGH Warsaw School of Economics


logankilpatrick:
Thank you!

PR #1849, opened 2023-03-04
Clean up old ML projects and add forward pointer to Flux

darsnack:
Remove:
- old CUDA hacking project (as discussed on Slack, it is stale and overlaps significantly with an existing FluxML project)
- fluxml.md idea list file (no longer necessary, this is on our own website)

Add forward pointer to FluxML GSoC page (in case people don't read carefully and still end up on the generic ML idea page)


darsnack:
cc @logankilpatrick

logankilpatrick (approved):

logankilpatrick:
Thank you!

PR #1852, opened 2023-03-11
GSoC: no pluto this year 💔

fonsp:
Unfortunately we won’t have time to mentor this summer! 💔

We’ll join again next year!


fonsp:
@logankilpatrick

PR #1855, opened 2023-03-16
Over 9100 packages

PallHaraldsson:


PallHaraldsson:
40 million downloads may also be outdated. I don't find a source for a higher number. Elsewhere for 35 million. I want to quote the highest number elsewhere, and at least want to know.

PR #1873, opened 2023-04-15
Add in code to fetch and display release date for a specific version

logankilpatrick:


logankilpatrick:
closes #1850

downloads/oldreleases.jl
```diff
@@ -71,9 +82,16 @@ open("./oldreleases.md", "w") do io
             """)
     for (version, files) in releases
         n = length(files)
+        release_date = ""
```
KristofferC, line 85: I think this block is just `get(release_dict, version, "Unknown")`

logankilpatrick: Awesome, switched it over.

PR #1875, opened 2023-04-17
Add the Julia 1.9 highlight blogpost

KristofferC:
Julia 1.9 is drawing close and there should of course be a corresponding highlight blog post. This is a start of it but contributions for new exciting things that are missed here is also welcome.


blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,376 @@
...
+
+For other types, the default sorting algorithm has been changed to the internal `ScratchQuickSort`, which is stable and generally faster than `QuickSort`, although it does allocate memory. For situations where memory efficiency is crucial, you can override these new defaults by specifying `alg=QuickSort`.
+
+To learn more about the latest advancements in high-performance sorting in Julia, watch the JuliaCon 2022 talk, [Julia's latest in high performance sorting](https://www.youtube.com/watch?v=9SGuHFgJvSE).
+
+## Tasks and the interactive thread pool
+
+Before version 1.9, Julia treated all tasks equally, running them on all available threads without any distinction in priority. However, there are situations where you may want certain tasks to be prioritized, such as when running a [heartbeat](https://en.wikipedia.org/wiki/Heartbeat_(computing)) , providing an interactive interface, or displaying progress updates.
+
+To address this need, you can now designate a task as interactive when you [Threads.@spawn](https://chat.openai.com/c/@ref)  it:
```
simonbyrne, line 164:
```suggestion
To address this need, you can now designate a task as interactive when you [`Threads.@spawn`](https://docs.julialang.org/en/v1/base/multi-threading/#Base.Threads.@spawn)  it:
```

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,376 @@
...
+
+function plot(contour::Countours.Contour)
+    ...
+end
+```
+
+However, adding package dependencies can have costs, such as increased load times or require installation of large artifacts (e.g., CUDA.jl). This can be burdensome for package authors who must constantly balance dependency costs against the benefits of new method extensions for an "average" package user.
+
+Julia 1.9 introduces "package extensions", a feature that (in a loose sense) automatically loads a file when a set of packages are loaded. This file contains the code to load the (weak) dependency and extend the method. The goal is that one shouldn't have to pay for features that one does not use. Package extensions provides functionality simiar to what Requires.jl already offers but with key advantages, such as allowing precompilation of conditional code and adding compatibility constraints on weak dependencies.
+Since these are now "first class", package authors should be less reluctant to start using it compared to Requires.jl could feel a little bit like a hack.
```
timholy, line 79:
```suggestion
Since these are now "first class", package authors should be less reluctant to start using it compared to Requires.jl.
```

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,376 @@
...
+# 1.8 (StaticArrays unconditionally loaded)
+julia> @time using ForwardDiff
+  0.590685 seconds (2.76 M allocations: 201.567 MiB)
+
+# 1.9 (StaticArrays not loaded)
+julia>  @time using ForwardDiff
+  0.247568 seconds (220.93 k allocations: 13.793 MiB)
+```
+    
+For a comprehensive guide on using package extensions, please refer to the [documentation](https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)) .
```
timholy, line 93:
```suggestion
For a comprehensive guide on using package extensions, please refer to the [documentation](https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)).
```

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,376 @@
...
+  acosh       0 bytes acosh (generic function with 12 methods)     
+  acot        0 bytes acot (generic function with 4 methods) 
+...
+```
+
+### Numbered prompt
+
+*Kristoffer Carlsson*
+
+Drawing heavily on inspiration from the IPython shell (and other notebook based systems like Mathematica), the Julia REPL can enable a "numbered prompt" that stores evaluated objects in the REPL for later use and keeps track of the number of expressions that have been evaluated.
```
timholy, line 217:
```suggestion
Drawing heavily on inspiration from the IPython shell (and other notebook-based systems like Mathematica), the Julia REPL can enable a "numbered prompt" that stores evaluated objects in the REPL for later use and keeps track of the number of expressions that have been evaluated.
```

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,376 @@
...
+
+Being able to refer to an earlier evaluated object can be useful if one for exampe forgets to store the result of a long computation to a variable and then executes something else (so that `ans` gets overwritten).
```
timholy, line 219:
```suggestion
Being able to refer to an earlier evaluated object can be useful if, for example, one forgets to store the result of a long computation to a variable and then executes something else (so that `ans` gets overwritten).
```

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,376 @@
...
+| Julia REPL with "numbered prompt" | IPython REPL |
+
+For instructions how to enable this, see the documentation [ADD LINK]
+
+
+## DelimitedFiles -- first stdlib to be upgradable
+
+*Kristoffer Carlsson*
+
+Julia comes with a set of standard libraries ("stdlibs") which are similar to normal packages except that they can be loaded without having to explicitly install them. Most of these stdlibs also comes "prebaked" in the sysimage that Julia ships which means that they technically gets loaded every time Julia is started.
```
timholy, line 233:
```suggestion
Julia comes with a set of standard libraries ("stdlibs") which are similar to normal packages except that they can be loaded without having to explicitly install them. Most of these stdlibs also comes "prebaked" in the sysimage that Julia ships which means that they technically get loaded every time Julia is started.
```

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,376 @@
...
+*Kristoffer Carlsson*
+
+Previously `pkg> up Foo` would freely update any dependency in the environment. Now `up` has the same preserve strategies that `add` observes, meaning that first `up Foo` will only allow `Foo` to update. It is possible to loosen this restriction a bit with the various `--preserve` command options to also allow dependencies to `Foo` to update. See the documentaion for `Pkg.update` for more information.
+
+
+### `pkg> add` will only auto update the registry once per day
+
+*Kristoffer Carlsson*
+
+Pkg will now remember the last time the registry was updated across julia sessions and only auto-update once per day when using an `add` command. Previously the registry auto-updated once per session. Note that the `update` command
```
timholy, line 296: Unfinished?

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,376 @@
...
+
+*Ian Butterworth*
+
+Previously, coverage testing could only be enabled for either `all` where all code visited is checked, `user` (the prior default for `Pkg.test`) where everything except for `Base` is checked including stdlibs, or `none` where tracking is disabled.
+
+The github action `julia-runtest` defaults to coverage testing on, meaning a lot of tracking outside of the package under test previously took place, slowing tests down, especially tight loops. 
+
+v1.8 introduced the ability to specify a path to either a file or directory for coverage to be tracked in via `--code-coverage=@path`, and v1.9 makes that the default for `Pkg.test(coverage=true)` (and thus used by `julia-runtest` by default).
+
+This change means often a lot less code needs to be tracked, and in cases where code from dependencies fall in tight loops this can heavily speed up the test suite. In one example Octivarian.jl tests with coverage enabled went from >2hrs to ~6 minutes.
```
timholy, line 335: Octivarian -> Octavian?

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,376 @@
...
+
+## LLVM Update to v14
+
+*Valentin Churavy, Mosè Giordano*
+
+LLVM is the underlying compiler infrastructure that Julia compiler builds ontop of. With Julia 1.9 we update the version used to [v14.0.6](https://releases.llvm.org/14.0.0/docs/ReleaseNotes.html).
```
timholy, line 341:
```suggestion
LLVM is the underlying compiler infrastructure that Julia's compiler builds upon. With Julia 1.9 we update the version used to [v14.0.6](https://releases.llvm.org/14.0.0/docs/ReleaseNotes.html). 
```

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,376 @@
...
+
+## Native half-precision floating-point arithmetic
+
+*Gabriel Baraldi, Mosè Giordano*
+
+To execute arithmetic operations on Float16 values julia used to promote them to Float32, and then convert them back to Float16 to return the result. 1.9 added support for native Float16 operations on AArch64 CPUs that have hardware support for half-precision floating-point arithmetic, like Apple's M series or Fujitsu's A64FX. In memory-bound applications this allows for up to 2x speedup compared to Float32 operations and 4x compared to Float64 operations. For example, on an M1 MacBook you can get
```
timholy, line 349:
```suggestion
To execute arithmetic operations on `Float16` values, julia used to promote them to `Float32`, and then convert them back to `Float16` to return the result. 1.9 added support for native `Float16` operations on AArch64 CPUs that have hardware support for half-precision floating-point arithmetic, like Apple's M series or Fujitsu's A64FX. In memory-bound applications this allows for up to 2× speedup compared to `Float32` operations and 4× compared to `Float64` operations. For example, on an M1 MacBook you can get
```

blog/2023/04/julia-1.9-highlights.md
IanButterworth:
```suggestion
This change means often a lot less code needs to be tracked, and in cases where code from dependencies fall in tight loops this can heavily speed up the test suite. In one example Octavian.jl tests with coverage enabled went from >2hrs to ~6 minutes.
```

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,376 @@
...
+
+With the introduction of Julia 1.9, native code caching is now available, resulting in a significant improvement in TTFX latency and paving the way for future enhancements across the ecosystem. Package authors can now utilize `precompile` statements or workloads with [SnoopPrecompile](https://timholy.github.io/SnoopCompile.jl/stable/snoop_pc/) to cache important routines in advance. Users can also create custom local "Startup" packages that load dependencies and precompile workloads tailored to their daily work.
+
+This feature comes with some tradeoffs, such as an increase in precompilation time by 10%-50%. However, since this is a one-time cost, we believe the tradeoff is well worth it. Cache files have also become larger due to the storage of more data and the use of a different serialization format.
+
+The graph below illustrates the changes in time-to-load (TTL), TTFX, and cache file size starting with Julia 1.7 (prior to the precompilation improvements):
```
johnnychen94, line 36: Many people around are wondering why native code caching(pkgimage) doesn't reduce TTL as PackageCompiler does, maybe we can explain it a bit here?

KristofferC: I guess the answer is that you are not executing any code in the package when you load it, so why would having native code available for it improve the load time? Not sure how to write that bst.

timholy: See if you like the updates, @johnnychen94

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,392 @@
...
+
+Previously, coverage testing could only be enabled for either `all` where all code visited is checked, `user` (the prior default for `Pkg.test`) where everything except for `Base` is checked including stdlibs, or `none` where tracking is disabled.
+
+The github action `julia-runtest` defaults to coverage testing on, meaning a lot of tracking outside of the package under test previously took place, slowing tests down, especially tight loops.
+
+v1.8 introduced the ability to specify a path to either a file or directory for coverage to be tracked in via `--code-coverage=@path`, and v1.9 makes that the default for `Pkg.test(coverage=true)` (and thus used by `julia-runtest` by default).
+
+This change means often a lot less code needs to be tracked, and in cases where code from dependencies fall in tight loops this can heavily speed up the test suite. In one example Octavian.jl tests with coverage enabled went from >2hrs to ~6 minutes.
+
+## LLVM Update to v14
```
giordano, line 353: @vchuravy @gbaraldi are there more highlights from llvm 14 that could be added here?

vchuravy: @pchintalapudi is this the first version with IRCE?

pchintalapudi: No, IRCE was introduced in 1.8.

blog/2023/04/julia-1.9-highlights.md
johnnychen94: The table looks great! -- one immediately knows that pkgimage _almost_ sunsets sysimage 😆 

Do we have ideas to reduce the TTL if not using PackageCompiler? From my first impression, it looks very hard if the bottleneck is at the validation-checking. I didn't check very actively these days so I might miss the discussion; if we got a clear plan, adding an advance notice on whether/how/when we get sysimage deprecated for TTL/TTFX would be much better for users to get the landscape!

KristofferC: > Do we have ideas to reduce the TTL if not using PackageCompiler? 

Much work has been done for this on master already.

timholy: Yeah, master is tons better than 1.9, although to set expectations properly we're not talking about orders of magnitude improvement (e.g., currently Makie goes from 7s->5s, which is a huge and very welcome change). @johnnychen94 I think you're basically right in your analysis: it's less clear that there are orders-of-magnitude improvements that can be made in TTL. In addition to the amazing work on optimization that's been done so far, there are some ideas for deferring validation that seem to me likely to get another significant boost. However, I am very reluctant to make projections or promises, so I'd rather not add any "advance notice" about how TTL will be better in the future.

blog/2023/04/julia-1.9-highlights.md
gbaraldi: Didn't SVE work better on LLVM14 than it did on 13 @giordano ?

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,393 @@
...
+
+With the introduction of Julia 1.9, native code caching is now available, resulting in a significant improvement in TTFX latency and paving the way for future enhancements across the ecosystem. Package authors can now utilize `precompile` statements or workloads with [PrecompileTools](https://github.com/JuliaLang/PrecompileTools.jl) to cache important routines in advance. Users can also create custom local "Startup" packages that load dependencies and precompile workloads tailored to their daily work.
+
+This feature comes with some tradeoffs, such as an increase in precompilation time by 10%-50%. However, since this is a one-time cost, we believe the tradeoff is well worth it. Cache files have also become larger due to the storage of more data and the use of a different serialization format.
+
+The graph below illustrates the changes in time-to-load (TTL), TTFX, and cache file size starting with Julia 1.7 (prior to the precompilation improvements):
+
+
+![](https://i.imgur.com/kydvmQL.png)
```
timholy, line 40:
```suggestion
![](https://i.imgur.com/kydvmQL.png)
Methodology: a demonstration workload was designed for each package. That workload was placed in the [Startup package](https://julialang.github.io/PrecompileTools.jl/stable/#Tutorial:-local-%22Startup%22-packages) and precompiled; for benchmarking, we load the Startup package and run the same workload.
```

It seems to make some sense to document the methodology used to make these measurements, while also calling extra attention to the benefits of custom startup packages.

mkitti (changes requested):
I made a few suggestions to fix a spelling mistake in the package extensions section as well. I made a few other changes to improve readability of the surrounding paragraph. I also suggest expanding contractions since these technical release notes are quite formal in nature.

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,399 @@
...
+import Contours
+
+function plot(contour::Countours.Contour)
+    ...
+end
+```
+
+However, adding package dependencies can have costs, such as increased load times or require installation of large artifacts (e.g., CUDA.jl). This can be burdensome for package authors who must constantly balance dependency costs against the benefits of new method extensions for an "average" package user.
+
+Julia 1.9 introduces "package extensions", a feature that (in a loose sense) automatically loads a file when a set of packages are loaded. This file contains the code to load the (weak) dependency and extend the method. The goal is that one shouldn't have to pay for features that one does not use. Package extensions provides functionality simiar to what Requires.jl already offers but with key advantages, such as allowing precompilation of conditional code and adding compatibility constraints on weak dependencies.
```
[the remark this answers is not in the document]
IanButterworth: Agree with some, not with others:
```suggestion
Julia 1.9 introduces "package extensions", a feature that, in a loose sense, automatically loads a module when a set of packages are loaded. The module, contained within a file in the `ext` directory of the parent package, loads the weak dependency and extend methods. The goal is that one should not have to pay for features that one does not use. Package extensions provide functionality similar to that which Requires.jl already offers but with key advantages, such as allowing precompilation of conditional code and adding compatibility constraints on weak dependencies.
```

blog/2023/04/julia-1.9-highlights.md
```diff
@@ -0,0 +1,399 @@
...
+authors = "..."
+published = "17 April 2023"
+rss_pubdate = Date(2023, 04, 17)
+rss = """Highlights of the Julia 1.9 release."""
++++
+
+
+After 3 betas and 3 release candidates, Julia version 1.9 has finally(!) been released. We would like to thank all the contributors to this release and all the testers that helped with finding regressions and issues in the pre-releases. Without you, this release would not have been possible.
+
+The full list of changes can be found in the [NEWS file](https://github.com/JuliaLang/julia/blob/release-1.9/NEWS.md), but here we'll give a more in-depth overview of some of the release highlights.
```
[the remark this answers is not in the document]
IanButterworth: I disagree about the "but". The former is the full but brief list, the latter is in-depth. There is a difference, so it seems appropriate.

```suggestion
The full list of changes can be found in the [NEWS file](https://github.com/JuliaLang/julia/blob/release-1.9/NEWS.md), but here we will give a more in-depth overview of some of the release highlights.
```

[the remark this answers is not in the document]
mkitti: Yes, there is a difference here. However, I think we want to emphasize the complementary nature of the blog not a contradiction or caveat that differentiates between the two.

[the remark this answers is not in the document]
mkitti: > This is not formal writing though.

The key differentiator is the intended audience. If the intended audience is just our buddies that we are on a first name basis with, then perhaps this is informal. If the intended audience is for people we do not know personally, such as a broad international audience of potential developers who may considering Julia, then I would argue this is in fact formal writing.

I can see either case, but I think we should err on the side of being formal here.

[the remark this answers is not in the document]
KristofferC: > The key differentiator is the intended audience. If the intended audience is just our buddies that we are on a first name basis with, then perhaps this is informal

I would disagree with that this is the only setting for "informal" writing. But this is a collection of different people's writing so I think it is fine if there are some inconsistencies. 

This is a highlight so it is supposed to be a bit more laid back and light hearted than standard technical writing.

FWIW, looked at the rust blogposts for releases and they seem to use similar language as the one here.

mkitti (approved):
Thanks for correcting the spelling and grammar.

DilumAluthge (approved):

PR #1889, opened 2023-05-10
add units to ttfx table

KristofferC:


KristofferC:
Hm, this causes the first column to occupy two rows and doesn't look very pretty.

ViralBShah:
While uglier, it is better in the table. Many people will only look at the table and not the text.

KristofferC:
I'll just put this in for now then

PR #1894, opened 2023-05-10
Add a methodology footnote

timholy:
I originally submitted this in https://github.com/JuliaLang/www.julialang.org/pull/1875#discussion_r1175303290 and then must have clicked resolve before it got merged.


timholy:
Did you check that the footnote formatting worked? I don't see it in the preview.

ViralBShah:
It did not work. I was just looking to get it to pass the CI. Should we just put it inline?

timholy:
Seems like a good idea. Where? It's kinda long and feels like an interruption no matter where I place it.

ViralBShah:
I think having it as the last paragraph in the Caching of Native code section seems like the right place to me.

PR #1897, opened 2023-05-10
Fix formatting on Methodology

timholy:
Attempted fix at formatting. Check the preview before merging.


timholy:
https://julialang.netlify.app/previews/pr1897/blog/2023/04/julia-1.9-highlights/

Seems mostly OK to me. If anyone is daring we could make "See Methodology below for details" into a link, but given that fancy formatting didn't work before I thought I'd avoid tempting fate.

PR #1903, opened 2023-06-01
rm notepad++

xgdgsc:
This doesn' t seem very useful listed here. Maybe replace with sublime make more sense?


ViralBShah:
I thought sublime was not maintained anymore. If sublime is worth a mention, please submit a separate PR.

PR #1904, opened 2023-06-05
New blogpost: PSA: Thread-local state is no longer recommended; Common misconceptions about `threadid()` and `nthreads()`

MasonProtter:
This is a blogpost some of us have been working on initially on hackmd, spurred by the somewhat surprising discovery of just how many packages out in the wild are using this dangerous pattern.

I feel myself being strongly pulled towards expanding this into a rather long form discussion of multithreading, and also wanting to wait for PRs to replace `@threads` with a `pmapreduce` or whatever, but I think the most important short term thing is to keep this blog post somewhat short, and focused on the specific problem of people indexing into storage using `threadid()` (or other common invalid usages of `threadid()`), and offering fixes without lowering the signal to noise ratio too much -- we need people to actually read and understand this!

Feedback, fixes, and bikeshedding all appreciated and welcome. 
______________

## TODO's before publication
- [ ] Update master & 1.9 docs to guide against misuse of `threadid()` e.g. 
    - [x] `add docs on task-specific buffering using @threads` [#48542](https://github.com/JuliaLang/julia/pull/48542)
    - [x] `add docs on task migration` [#50047](https://github.com/JuliaLang/julia/pull/50047)
    - [ ] Add `!!1compat` notes to threadid/nthreads/maxthreadid
    - [x] Add note to recommend against `@async`
- [x] Make sure that ^ 1.9 version is released so docs are discoverable
- [x] Get signoff by core multithreading developers
- [x] More detail in the Packages section, perhaps with examples (or remove it)
- [x] Bikeshead the title.

### DIFF
--- a/blog/2023/06/PSA-dont-use-threadid.md
+++ b/blog/2023/06/PSA-dont-use-threadid.md
@@ -112,7 +112,7 @@ The above `pmampreduce` can hopefully be added to base Julia at some point in th
 
 ## Another option: Use a package which handles this correctly
 
-We encourage people to check out various multithreading libraries like [Transducers.jl](https://github.com/JuliaFolds2/Transducers.jl) (or various frontends like [ThreadsX.jl](https://github.com/tkf/ThreadsX.jl), [FLoops.jl](https://github.com/JuliaFolds/FLoops.jl), and [Folds.jl](https://github.com/JuliaFolds/Folds.jl)), <!--- [Polyester.jl](https://github.com/JuliaSIMD/Polyester.jl), --> and [MultiThreadedCaches.jl](https://github.com/JuliaConcurrent/MultiThreadedCaches.jl).
+We encourage people to check out various multithreading libraries like [Transducers.jl](https://github.com/JuliaFolds2/Transducers.jl) (or various frontends like [ThreadsX.jl](https://github.com/tkf/ThreadsX.jl), [FLoops.jl](https://github.com/JuliaFolds/FLoops.jl), and [Folds.jl](https://github.com/JuliaFolds/Folds.jl)),  and [MultiThreadedCaches.jl](https://github.com/JuliaConcurrent/MultiThreadedCaches.jl).
 
 ### Transducers.jl ecosystem
 Transducers.jl is fundamentally about expressing the traversing of data in a structured and principled way, often with the benefit that it makes parallel computing easier to reason about. 
--- a/blog/2023/06/PSA-dont-use-threadid.md
+++ b/blog/2023/06/PSA-dont-use-threadid.md
@@ -214,7 +214,7 @@ MultiThreadedCaches.jl on the other hand attempts to make the `states[threadid()
 
 [^yielding]: ~~~<h4>~~~Don't try to reason about yielding~~~</h4>~~~
 @@long-footnote
-   Many existing uses of thread local state happen to be relatively robust and give correct answers only because the functions they are calling during execution do not yield. One may then think "well, I can just avoid this problem by making sure my code doesn't yeield", but we think this is a bad and unsustainable idea, because whether or not a function call will yield is not stable, obvious, or easily inspectable.
+   Many existing uses of thread local state happen to be relatively robust and give correct answers only because the functions they are calling during execution do not yield. One may then think "well, I can just avoid this problem by making sure my code doesn't yield", but we think this is a bad and unsustainable idea, because whether or not a function call will yield is not stable, obvious, or easily inspectable.
 
    For instance, if a function `f` is updated to include a background `@debug` statement or other forms of non-user-visible IO, it may change from being non-yielding to yielding. If during a call to `f`, the compiler encounters a dynamic dispatch where new code must be JIT compiled, a yield-point may be encountered, and any number of other internal changes could happen to code which can cause it to yielding.
 
--- a/blog/2023/06/PSA-dont-use-threadid.md
+++ b/blog/2023/06/PSA-dont-use-threadid.md
@@ -1,10 +1,4 @@
-@def rss_pubdate = Date(2023, 6, XX)
-@def rss = """ TBD """
-@def published = "xx June 2023"
-@def title = "PSA: Stop using `states[threadid()]`"
-@def authors = """Mason Protter, Valentin Churavy, ..."""
-@def mintoclevel=2
-@def maxtoclevel=3
++++ mintoclevel = 2 maxtoclevel = 3 title = "PSA: Stop using `states[threadid()]`" authors = "Mason Protter, Valentin Churavy, Ian Butterworth, ..." published = "XX June 2023" rss_pubdate = Date(2023, 06, XX) rss = """PSA: Stop using `states[threadid()]`""" +++
 
 # PSA: Stop using `states[threadid()]`
 Alt titles:
--- a/blog/2023/06/PSA-dont-use-threadid.md
+++ b/blog/2023/06/PSA-dont-use-threadid.md
@@ -89,11 +89,13 @@ Code which replaces the incorrect code pattern shown above can look like this:
 using Base.Threads: nthreads, @threads, @spawn
 using Base.Iterators: partition
 
-tasks_per_thread = 2 # customize this as needed. More tasks have more overhead, but better load balancing
+tasks_per_thread = 2 # customize this as needed. More tasks have more overhead, but better
+                     # load balancing
 
 chunk_size = max(1, length(some_data) ÷ (tasks_per_thread * nthreads()))
-data_chunks = partition(some_data, chunk_size) # partition your data into chunks that individual tasks will deal with
-# See also ChunkSplitters.jl and SplittablesBase.jl for partitioning data
+data_chunks = partition(some_data, chunk_size) # partition your data into chunks that
+                                               # individual tasks will deal with
+#See also ChunkSplitters.jl and SplittablesBase.jl for partitioning data
 
 tasks = map(data_chunks) do chunk
     # Each chunk of your data gets its own spawned task that does its own local, sequential work
@@ -106,8 +108,8 @@ tasks = map(data_chunks) do chunk
         return state
     end
 end
-states = fetch.(tasks) # get all the values returned by the individual tasks.
-# fetch is type unstable, so you may optionally want to assert a specific return type.
+states = fetch.(tasks) # get all the values returned by the individual tasks. fetch is type
+                       # unstable, so you may optionally want to assert a specific return type.
 
 do_something(states)
 ```
--- a/blog/2023/06/PSA-dont-use-threadid.md
+++ b/blog/2023/06/PSA-dont-use-threadid.md
@@ -100,7 +100,7 @@ do_something(states)
 This is a fully general replacement for the old, buggy pattern. However, for many applications this should be simplified down to a parallel version of `mapreduce`:
 ```julia
 using Threads: nthreads, @spawn
-function pmapreduce(f, op, itr; init=some_initial_value, chunks_per_thread::Int = 2)
+function tmapreduce(f, op, itr; init=some_initial_value, chunks_per_thread::Int = 2)
     chunk_size = max(1, length(itr) ÷ chunks_per_thread)
     tasks = map(Iterators.partition(itr, chunk_size)) do chunk
         @spawn mapreduce(f, op, chunk; init=init)
@@ -108,9 +108,9 @@ function pmapreduce(f, op, itr; init=some_initial_value, chunks_per_thread::Int 
     mapreduce(fetch, op, tasks; init=init)
 end
 ```
-In `pmapreduce(f, op, itr)`, the function `f` is applied to each element of `itr`, and then an *associative*[^assoc] two-argument function `op`.
+In `tmapreduce(f, op, itr)`, the function `f` is applied to each element of `itr`, and then an *associative*[^assoc] two-argument function `op`.
 
-The above `pmampreduce` can hopefully be added to base Julia at some point in the near future. In the meantime however it's somewhat simple to write your own as above.
+The above `tmampreduce` can hopefully be added to base Julia at some point in the near future. In the meantime however it's somewhat simple to write your own as above.
 
 ## Another option: Use a package which handles this correctly
 
@@ -119,7 +119,7 @@ We encourage people to check out various multithreading libraries like [Transduc
 ### Transducers.jl ecosystem
 Transducers.jl is fundamentally about expressing the traversing of data in a structured and principled way, often with the benefit that it makes parallel computing easier to reason about. 
 
-The above `pmapreduce(f, op, itr)` could be expressed as
+The above `tmapreduce(f, op, itr)` could be expressed as
 ```julia
 using Transducers
 itr |> Map(f) |> foldxt(op; init=some_initial_value)
--- a/blog/2023/06/PSA-dont-use-threadid.md
+++ b/blog/2023/06/PSA-dont-use-threadid.md
@@ -68,7 +68,7 @@ However, this is not a good long term solution because
 
 If you want a recipe that can replace the above buggy one with something that can be written using only the `Base.Threads` module, we recommend moving away from `@threads`, and instead working directly with `@spawn` to create and manage tasks. The reason is that `@threads` does not have any builtin mechanisms for managing and merging the results of work from different threads, whereas tasks can manage and return their own state in a safe way.
 
-Tasks creating and returning their own state is inherently safer than the spawner of paralell tasks setting up state for spawned tasks to read from and write to. 
+Tasks creating and returning their own state is inherently safer than the spawner of parallel tasks setting up state for spawned tasks to read from and write to. 
 
 Code which replaces the incorrect code pattern shown above can look like this:
 ```julia
--- a/blog/2023/06/PSA-dont-use-threadid.md
+++ b/blog/2023/06/PSA-dont-use-threadid.md
@@ -1,7 +1,21 @@
-+++ mintoclevel = 2 maxtoclevel = 3 title = "PSA: Thread-local state is no longer recommended; Common misconceptions about threadid() and nthreads()" authors = "Mason Protter, Valentin Churavy, Ian Butterworth, ..." published = "XX June 2023" rss_pubdate = Date(2023, 06, XX) rss = """PSA: Thread-local state is no longer recommended; Common misconceptions about threadid() and nthreads()""" +++
++++
+mintoclevel = 2
+maxtoclevel = 3
+title = "PSA: Thread-local state is no longer recommended"
+authors = "Mason Protter, Valentin Churavy, Ian Butterworth, ..."
+published = "30 June 2023"
+rss_pubdate = Date(2023, 06, 30)
+rss = """PSA: Thread-local state is no longer recommended; Common misconceptions about threadid() and nthreads()"""
++++
 
-# PSA: Thread-local state is no longer recommended; Common misconceptions about threadid() and nthreads()
 
+*Common misconceptions about threadid() and nthreads()*
+
+\toc
+
+
+
+## The buggy pattern
 
 Partially due to the evolving history of our parallel and concurrent interfaces[^historynote], some Julia programmers have been writing *incorrect* parallel code that contains the possibility of race conditions which can give wrong results. This pattern has even been erroneously recommended in [previous official julia blogposts](https://github.com/JuliaLang/www.julialang.org/blob/main/blog/2019/07/multithreading.md#thread-local-state). It is important for the stability and correctness of the ecosystem that these usages are identified and fixed.
 
@@ -22,15 +36,17 @@ do_something(states)
 
 The above code is **incorrect** because the tasks spawned by `@threads` are  allowed to yield to other tasks during their execution. This means that between reading `old_val` and storing `new_val` in the storage, the task could be paused and a new task running on the same thread with the same `threadid()` could concurrently write to `states[tid]`, causing a race condition and thus work being lost.
 
-This is not actually a problem with multithreading specifically, but really a concurrency problem, and it can be demonstrated even with a single thread. For example: 
+This is not actually a problem with multithreading specifically, but really a concurrency problem, and it can be demonstrated even with a single thread. For example:
+
 ```
 $ julia --threads=1
 ```
+
 ```julia
 julia> f(i) = (sleep(0.001); i);
 
 julia> let state = [0], N=100
-           @sync for i ∈ 1:N 
+           @sync for i ∈ 1:N
                Threads.@spawn begin
                    tid = Threads.threadid()
                    old_var = state[tid]
@@ -42,13 +58,14 @@ julia> let state = [0], N=100
        end
 (100, 5050)
 ```
+
 In the above snippet, we purposefully over-subscribed the CPU with `100` separate tasks in order to make the bug more likely to manifest, but the problem **can** arise even without spawning very many tasks.
 
 Any usage of `threadid()` in package or user code should be seen as a warning sign that the code is relying on implementation details, and is open to concurrency bugs.
 
-# Fixing buggy code which uses this pattern
+## Fixing buggy code which uses this pattern
 
-## Quickfix: Replace `@threads` with `@threads :static`
+### Quickfix: Replace `@threads` with `@threads :static`
 
 The simplest (though not recommended longterm) quickfix if you happen to use `@threads` is to replace usages of `@threads for ...` with `@threads :static for ...`. The reason for this is that the `:static` scheduler for `@threads` does not allow the asynchronous task migration and yielding that causes the bug in the first place.
 
@@ -56,13 +73,14 @@ However, this is not a good long term solution because
 1) It's relying on guard rails to prevent otherwise incorrect code to be correct
 2) `@threads :static` is not cooperative or composable, that means that if code inside your `@threads :static` loop also does multithreading, your code could be reduced to worse than single-threaded speeds, or even deadlock.
 
-## Better fix: Work directly with tasks
+### Better fix: Work directly with tasks
 
 If you want a recipe that can replace the above buggy one with something that can be written using only the `Base.Threads` module, we recommend moving away from `@threads`, and instead working directly with `@spawn` to create and manage tasks. The reason is that `@threads` does not have any builtin mechanisms for managing and merging the results of work from different threads, whereas tasks can manage and return their own state in a safe way.
 
-Tasks creating and returning their own state is inherently safer than the spawner of parallel tasks setting up state for spawned tasks to read from and write to. 
+Tasks creating and returning their own state is inherently safer than the spawner of parallel tasks setting up state for spawned tasks to read from and write to.
 
 Code which replaces the incorrect code pattern shown above can look like this:
+
 ```julia
 using Base.Threads: nthreads, @threads, @spawn
 using Base.Iterators: partition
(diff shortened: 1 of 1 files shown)


blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,174 @@
...
+    mapreduce(fetch, op, tasks; init=init)
+end
+```
+In `pmapreduce(f, op, itr)`, the function `f` is applied to each element of `itr`, and then an *associative*[^assoc] two-argument function `op`.
+
+The above `pmampreduce` can hopefully be added to base Julia at some point in the near future. In the meantime however it's somewhat simple to write your own as above.
+
+## Another option: Use a package which handles this correctly
+
+We encourage people to check out various multithreading libraries like [Transducers.jl](https://github.com/JuliaFolds2/Transducers.jl) (or various frontends like [ThreadsX.jl](https://github.com/tkf/ThreadsX.jl), [FLoops.jl](https://github.com/JuliaFolds/FLoops.jl), and [Folds.jl](https://github.com/JuliaFolds/Folds.jl)), <!--- [Polyester.jl](https://github.com/JuliaSIMD/Polyester.jl), --> and [MultiThreadedCaches.jl](https://github.com/JuliaConcurrent/MultiThreadedCaches.jl).
```
vchuravy, line 115:
```suggestion
We encourage people to check out various multithreading libraries like [Transducers.jl](https://github.com/JuliaFolds2/Transducers.jl) (or various frontends like [ThreadsX.jl](https://github.com/tkf/ThreadsX.jl), [FLoops.jl](https://github.com/JuliaFolds/FLoops.jl), and [Folds.jl](https://github.com/JuliaFolds/Folds.jl)),  and [MultiThreadedCaches.jl](https://github.com/JuliaConcurrent/MultiThreadedCaches.jl).
```

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,174 @@
...
+    @reduce out = op(some_initial_value, f(x))
+end
+out
+```
+
+### MultiThreadedCaches.jl
+
+MultiThreadedCaches.jl on the other hand attempts to make the `states[threadid()]`-like pattern safer by using lock mechanisms to stop data races. We think this is not an ideal way to proceed, but it may make transitioning to safer code easier and require fewer rewrites, such as in cases where a package must manage state which may be concurrently written to by a user, and the package cannot control how the user structures their code.
+
+[^historynote]: ### Concurrency & Parallelism
```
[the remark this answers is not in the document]
MasonProtter: It was originally put into the main body, but I moved it to a footnote because I wanted the main thrust of the article to be more focused, but I also couldn't bring myself to delete it.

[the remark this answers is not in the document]
vchuravy: Yeah, I am torn about it as well. I originally just wanted to define the terms, so that everyone understood them, but then it grew into this historical perspective.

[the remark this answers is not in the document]
giordano: Make it an Appendix section?

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,176 @@
...
+
+However, this is not a good long term solution because
+1) It's relying on guard rails to prevent otherwise incorrect code to be correct
+2) `@threads :static` is not cooperative or composable, that means that if code inside your `@threads :static` loop also does multithreading, your code could be reduced to worse than single-threaded speeds, or even deadlock.
+
+## Better fix: Work directly with tasks
+
+If you want a recipe that can replace the above buggy one with something that can be written using only the `Base.Threads` module, we recommend moving away from `@threads`, and instead working directly with `@spawn` to create and manage tasks. The reason is that `@threads` does not have any builtin mechanisms for managing and merging the results of work from different threads, whereas tasks can manage and return their own state in a safe way.
+
+Tasks creating and returning their own state is inherently safer than the spawner of paralell tasks setting up state for spawned tasks to read from and write to.
```
Seelengrab, line 71:
```suggestion
Tasks creating and returning their own state is inherently safer than the spawner of parallel tasks setting up state for spawned tasks to read from and write to. 
```

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,176 @@
...
+    chunk_size = max(1, length(itr) ÷ chunks_per_thread)
+    tasks = map(Iterators.partition(itr, chunk_size)) do chunk
+        @spawn mapreduce(f, op, chunk; init=init)
+    end
+    mapreduce(fetch, op, tasks; init=init)
+end
+```
+In `pmapreduce(f, op, itr)`, the function `f` is applied to each element of `itr`, and then an *associative*[^assoc] two-argument function `op`.
+
+The above `pmampreduce` can hopefully be added to base Julia at some point in the near future. In the meantime however it's somewhat simple to write your own as above.
```
Seelengrab, line 113: Should this be called `tmapreduce` instead, to not confuse it with `pmap` from Distributed (barring "these should be merged semantically anyway"), which spawns external processes?

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,176 @@
...
+
+    When the parallel scheduler was introduced, we decided that there was too much code in the wild using `@async` and expecting specific semantics, so `Threads.@spawn` was made available to access the new semantics. `@async` has various problems and we discourage its use for new code.
+
+    #### Uses of `threadid`/`nthreads`/`maxthreadid`
+
+    Over time, several features have been added that make relying on `threadid`, `nthreads` and even `maxthreadid` perilous:
+
+    1. Task migration: A task can observe multiple `threadid`s during its execution.
+    2. Interactive priority: `nthreads()` will report the number of non-interactive worker-threads, thus undercounting the number of active threads.
+    3. Thread adoption (v1.9): Foreign threads can now be adopted (and latter removed) at any time during the execution of the program.
```
Seelengrab, line 168:
```suggestion
    3. Thread adoption (v1.9): Foreign threads can now be adopted (and later be removed) at any time during the execution of the program.
```

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,176 @@
...
+    4. GC threads: The runtime can use additional threads to accelerate work like executing the Garbage Collector.
+
+    Any code that relies on a specific `threadid` staying constant, or on a constant number of threads during execution, is bound to be incorrect. As a rule of thumb, programmers should at most be querying the number of threads to motivate heuristics like how to distribute parallel work, but programs should generally **not** be written to depend on implementation details of threads for correctness. Rather, programmers should reason about *tasks*.
```
Seelengrab, line 171:
```suggestion
    Any code that relies on a specific `threadid` staying constant, or on a constant number of threads during execution, is bound to be incorrect. As a rule of thumb, programmers should at most be querying the number of threads to motivate heuristics like how to distribute parallel work, but programs should generally **not** be written to depend on implementation details of threads for correctness. Rather, programmers should reason about *tasks*, i.e. pieces of work that may execute concurrently with other code, independently of the number of *threads* that are used for executing them. 
```

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,176 @@
...
+    3. Thread adoption (v1.9): Foreign threads can now be adopted (and latter removed) at any time during the execution of the program.
+    4. GC threads: The runtime can use additional threads to accelerate work like executing the Garbage Collector.
+
+    Any code that relies on a specific `threadid` staying constant, or on a constant number of threads during execution, is bound to be incorrect. As a rule of thumb, programmers should at most be querying the number of threads to motivate heuristics like how to distribute parallel work, but programs should generally **not** be written to depend on implementation details of threads for correctness. Rather, programmers should reason about *tasks*. 
+    
+[^assoc]: ## Associativity
+    [Associativity](https://en.wikipedia.org/wiki/Associative_property) is an important property for parallel reducing functions because it means that `op(a, op(b, c)) == op(op(a, b), c)`, and hence the result does not depend on the order in which the reduction is performed. 
+    
+    Note that associativity is a not the same as commutivity which is the property that `op(a, b) = op(b, a)`. This is *not* required for parallel reducing functions.
```
Seelengrab, line 176:
```suggestion
[^assoc]: ## Associativity
    [Associativity](https://en.wikipedia.org/wiki/Associative_property) is an important property for parallel reducing functions, because it means that `op(a, op(b, c)) == op(op(a, b), c)`, and hence the result does not depend on the order in which the reduction is performed. 
    
    Note that associativity is not the same as commutivity, which is the property that `op(a, b) == op(b, a)`. This is *not* required for parallel reducing functions. 
```

blog/2023/06/PSA-dont-use-threadid.md
Seelengrab: Yeah I get that concern, I just think moving away from that terminology should be said explicitly as well, not just by introducing a confusing `pmapreduce` :shrug:

oscardssmith: the `t` in `tmapreduce` can stand for `task` :)

vchuravy: We also should start to reclaim parallelism from Distributed.jl xD It stole `@spawn` and now it tries to own the letter `p`.

MasonProtter: I've reverted it back to `tmapreduce`.

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,176 @@
++++ mintoclevel = 2 maxtoclevel = 3 title = "PSA: Stop using `states[threadid()]`" authors = "Mason Protter, Valentin Churavy, Ian Butterworth, ..." published = "XX June 2023" rss_pubdate = Date(2023, 06, XX) rss = """PSA: Stop using `states[threadid()]`""" +++
+
+# PSA: Stop using `states[threadid()]`
+Alt titles:
+- PSA: Multithreading with `states[threadid()]` is unsafe
+- PSA: Multithreading with `buffers[threadid()]` is unsafe
+- PSA: Don't assume `threadid()` is stable within a task
```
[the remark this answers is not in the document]
MasonProtter: I find this title rather misleadning, because the stability of `threadid()` is not the actual problem here. The problem is race conditions in updating the buffer at a specific `threadid()`. 

E.g. in the single threaded example in the blogpost, the `threadid()` is perfectly stable, it's always `1`, but the problem persists.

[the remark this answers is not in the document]
Seelengrab: A technically correct (but horrible UX) title could be

> PSA: Don't use non-threadsafe, volatile, shared state to attempt to coordinate concurrent task execution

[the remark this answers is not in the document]
MasonProtter: What I want is to catch the attention of people who are only dimly aware of this stuff and what it means.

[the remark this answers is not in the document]
vchuravy: "You are doing it wrong! Common misconceptions about threadid() and nthreads()"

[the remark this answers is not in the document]
IanButterworth: The last part of that is good

non-Jedi:
This blogpost should probably [acknowledge the previous blog post teaching this pattern](https://github.com/JuliaLang/www.julialang.org/blob/main/blog/2019/07/multithreading.md#thread-local-state). It would also be good to include a commit in this PR adding a disclaimer to that same blog post.

vchuravy:
> This blogpost should probably [acknowledge the previous blog post teaching this pattern](https://github.com/JuliaLang/www.julialang.org/blob/main/blog/2019/07/multithreading.md#thread-local-state). It would also be good to include a commit in this PR adding a disclaimer to that same blog post.

Oh my! That didn't age very well.

This gives a natural title: "We no longer recommend thread-local state;  Common misconceptions about threadid() and nthreads()"

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,176 @@
...
+@threads for x in some_data
+    tid = threadid()
+    old_val = states[tid]
+    new_val = some_operator(old_val, f(x))
+    states[tid] = new_val
+end
+do_something(states)
+```
+
+The above code is **incorrect** because the tasks spawned by `@threads` are  allowed to yield to other tasks during their execution.
```
KlausC, line 29: Wouldn't that be clearer?

```suggestion
The above code is **incorrect** because the tasks spawned by `@threads` are  neither guaranteed to have a threadid exclusively during their lifetime nor to keep the same threadid during their execution.
```

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,176 @@
...
+    @spawn begin
+        state = some_initial_value
+        for x in chunk
+            state = some_operator(state, f(x))
+        end
+        return state
+    end
+end
+states = fetch.(tasks) # get all the values returned by the individual tasks.
+# fetch is type unstable, so you may optionally want to assert a specific return value.
```
KlausC, line 95:
```suggestion
# fetch is type unstable, so you may optionally want to assert a specific return type.
```

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,176 @@
...
+# fetch is type unstable, so you may optionally want to assert a specific return value. 
+
+do_something(states)
+```
+
+This is a fully general replacement for the old, buggy pattern. However, for many applications this should be simplified down to a parallel version of `mapreduce`:
+```julia
+using Threads: nthreads, @spawn
+function tmapreduce(f, op, itr; init=some_initial_value, chunks_per_thread::Int = 2)
+    chunk_size = max(1, length(itr) ÷ chunks_per_thread)
```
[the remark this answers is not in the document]
Seelengrab: That again uses `nthreads`, which we explicitly want to get away from.

blog/2023/06/PSA-dont-use-threadid.md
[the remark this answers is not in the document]
IanButterworth: From @vchuravy's comment below, with a slight tweak:
```
PSA: Thread-local state is no longer recommended; Common misconceptions about threadid() and nthreads()
```

blog/2023/06/PSA-dont-use-threadid.md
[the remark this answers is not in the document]
MasonProtter: > In my experience in most cases an estimation for a reasonable chunk size is easier at hand than the number
of chunks

Typically, I find the opposite to be true, but that's why I provided both options. In the version of the code that I wrote, if the user specifies `chunks_per_thread` then `chunk_size` will be calculated from that, but if they specify `chunk_size` then it'll use that `chunk_size` and the `chunks_per_thread` argument will be ignored.



> That again uses `nthreads`, which we explicitly want to get away from. It's not a reliable number to query, hence the inversion of responsibility, thinking about the amount of tasks for that work, instead of the amount of work per task.

No, this is a valid usage of `nthreads`. The correctness of the program does not depend on `nthreads` being reliable, it's just used here as a heuristic for how to split up work, which is fine.

blog/2023/06/PSA-dont-use-threadid.md
MasonProtter: Hm, I think the "have a threadid exclusively during their lifetime" part is a bit easy to misunderstand. I think the sentence you're wanting to replace here should not be seen as an explanation, but as a definition at the start of an explanation, which is in the following couple of sentences.

I think it's useful to motivate or define here what "yielding" is

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,174 @@
...
+@threads for x in some_data
+    tid = threadid()
+    old_val = states[tid]
+    new_val = some_operator(old_val, f(x))
+    states[tid] = new_val
+end
+do_something(states)
+```
+
+The above code is **incorrect** because the tasks spawned by `@threads` are  allowed to yield to other tasks during their execution. This means that between reading `old_val` and storing `new_val` in the storage, the task could be paused and a new task running on the same thread with the same `threadid()` could concurrently write to `states[tid]`, causing a race condition and thus work being lost.
```
Seelengrab, line 29: Would it be a good idea to be more precise here, mentioning how adding an innocous looking debug printout in `f` or `some_operator` can induce the race conditions here, due to adding a yieldpoint? Or is it better not to mention what kind of stuff can make a task yield, since we're a cooperative runtime and not preemptive?

blog/2023/06/PSA-dont-use-threadid.md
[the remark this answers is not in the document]
MasonProtter: What's the difference?

blog/2023/06/PSA-dont-use-threadid.md
MasonProtter: Maybe that's a good use for a footnote or something?

blog/2023/06/PSA-dont-use-threadid.md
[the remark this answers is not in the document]
MasonProtter: okay I've changed the title-in-progress to that one. I'm still not really convinced it's any better since people don't really know what that means, but I also don't have strong feelings about it and the old title was also not perfect.

MasonProtter:
> This blogpost should probably [acknowledge the previous blog post teaching this pattern](https://github.com/JuliaLang/www.julialang.org/blob/main/blog/2019/07/multithreading.md#thread-local-state). It would also be good to include a commit in this PR adding a disclaimer to that same blog post.

Done in https://github.com/JuliaLang/www.julialang.org/pull/1904/commits/c69f3345848fc948db0fdad5e2b66e889d80f7f8

MasonProtter:
As I said in that thread, I'm not an expert on using channels, but my understanding is that that pattern is only useful if you

1) Need load balancing
2) The cost of spawning tasks is not negligible compared to your computation time
3) It's really important that you create as few intermediate states as possible

otherwise it's better and simpler and often faster to just spawn more tasks. 

If someone wants to propose a `Channel` based pattern to go into the blogpost as well as a discussion of when it should be used, then we can discuss it though. 

Alternatively, someone can propose a small note or even a footnote recommending people investigate channels as an additional tool, but without actually spelling out a specific pattern to use.

MasonProtter:
@vchuravy @vtjnash @IanButterworth @kpamnany, could I request some help getting this over the finish line? It'd be nice if we could put it out there even if its not perfect, I think having something we can point to whenever we see people do this pattern would be good.

vchuravy (approved):
From my side this looks great! Thank you Mason for leading the charge.

vtjnash (commented):
:+1:

IanButterworth:
Preview here https://julialang.netlify.app/previews/pr1914/blog/2023/06/psa-dont-use-threadid/ via #1914

timholy:
Formatting of top section seems messed up.

IanButterworth:
I'm working on some formatting fixes in #1914. I'll push them here when working

IanButterworth:
Formatting fixed and pushed here, thanks to @tlienart https://julialang.netlify.app/previews/pr1914/blog/2023/06/psa-dont-use-threadid/

IanButterworth:
Also the two docs PRs called out up top (https://github.com/JuliaLang/julia/pull/48542 https://github.com/JuliaLang/julia/pull/50047) are due to land in 1.9.2

blog/2023/06/PSA-dont-use-threadid.md
timholy: For developers who may be less familiar with race conditions, would it be easier to delay explaining what can go wrong until the `f(i) = (sleep(0.001); i)` example below? I'm imagining you could do something like this:
```julia
    tid = Threads.threadid()   # each task gets `tid = 1`
    old_var = state[tid]       # each task reads the current value, which for all is 0 (!) because... 
    new_var = old_var + f(i)   # ...the `sleep` in `f` causes all tasks to pause *simultaneously* here (all loop iterations start, but do not yet finish)
    state[tid] = new_var       # after being released from the `sleep`, each task sets `state[1]` to 1
```
My hope is that the comments walk readers through the logic more clearly than the short description above.

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,220 @@
...
+    tasks = map(Iterators.partition(itr, chunk_size)) do chunk
+        @spawn mapreduce(f, op, chunk; kwargs...)
+    end
+    mapreduce(fetch, op, tasks; kwargs...)
+end
+```
+
+In `tmapreduce(f, op, itr)`, the function `f` is applied to each element of `itr`, and then an *associative*[^assoc] two-argument function `op`.
+
+The above `tmampreduce` can hopefully be added to base Julia at some point in the near future. In the meantime however it's somewhat simple to write your own as above.
```
timholy, line 125:
```suggestion
The above `tmapreduce` can hopefully be added to base Julia at some point in the near future. In the meantime however it's somewhat simple to write your own as above.
```

timholy (commented):
An excellent and much-needed post, thanks!

blog/2023/06/PSA-dont-use-threadid.md
MasonProtter: hm yeah that probably is more clear than the prose. What would you think about having the comments *and* the prose?

MasonProtter: In addition to your comments Tim, @Seelengrab  I've added a new footnote trying to explain that while some code is currently okay because it happens to not yield, this is not a viable path forward and it could break at any time.

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -212,6 +212,17 @@ MultiThreadedCaches.jl on the other hand attempts to make the `states[threadid()
 
 \\
 
+[^yielding]: ~~~<h4>~~~Don't try to reason about yielding~~~</h4>~~~
+@@long-footnote
+   Many existing uses of thread local state happen to be relatively robust and give correct answers only because the functions they are calling during execution do not yield. One may then think "well, I can just avoid this problem by making sure my code doesn't yeield", but we think this is a bad and unsustainable idea, because whether or not a function call will yield is not stable, obvious, or easily inspectable.
```
IanButterworth, line 217: I can't diff for some reason `yeield -> yield`

MasonProtter: oops, thanks

tlienart (commented):
style note: for readability's sake it might be better to put # 1, # 2, # 3, ... and then put the comments in a numbered list below the code as otherwise this will y-overflow on devices and it's a bit awkward to read code having to scroll horizontally over it 🤷 

edit: hmm looks like I didn't do a good job at commenting the changes, I was referring to https://github.com/JuliaLang/www.julialang.org/pull/1904/commits/d3b24f259444d8d0757f27d1701d3a419562f3e4

IanButterworth:
1.9.2 is imminent and the docs are already live https://docs.julialang.org/en/v1/manual/multi-threading/#man-task-migration

MasonProtter:
So I tried a couple things to get the comments into a list like @tlienart suggested but I wasn't really happy with any of them and personally find it not so bad to scroll horizontally if the upside is that the comment is right beside the code. 

Like, if the comments aren't going to be in-line with the code then I'm not sure I really see the point because we already have a paragraph explaining what's going on anyways. 

Does anyone have suggestions on how to improve the comments in the code for more legibility? I'd really like to get this published.

MasonProtter:
Maybe lets get a new preview built so that we can see what it looks like?

IanButterworth:
https://julialang.netlify.app/previews/pr1914/blog/2023/06/psa-dont-use-threadid/

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,232 @@
...
+               Threads.@spawn begin
+                   tid = Threads.threadid()  # each task gets `tid = 1`
+                   old_var = state[tid]      # each task reads the current value, which for all is 0 (!) because... 
+                   new_var = old_var + f(i)  # ...the `sleep` in `f` causes all tasks to pause *simultaneously* here (all loop iterations start, but do not yet finish)
+                   state[tid] = new_var      # after being released from the `sleep`, each task sets `state[1]` to `i`
+               end
+           end
+           sum(state), sum(1:N)
+       end
+(100, 5050)
```
IanButterworth, line 59: This wrapping is based on what the preview shows without scrolling
```suggestion
julia> let state = [0], N=100
           @sync for i ∈ 1:N
               Threads.@spawn begin
                   tid = Threads.threadid()  # Each task gets `tid = 1`.
                   old_var = state[tid]      # Each task reads the current value, which for 
                                             # all is 0 (!) because...
                   new_var = old_var + f(i)  # ...the `sleep` in `f` causes all tasks to pause 
                                             # *simultaneously* here (all loop iterations start, 
                                             # but do not yet finish).
                   state[tid] = new_var      # After being released from the `sleep`, each task 
                                             # sets `state[1]` to `i`.
               end
           end
           sum(state), sum(1:N)
       end
(100, 5050)
```

MasonProtter: Isn't this going  to be device dependant?

IanButterworth: The suggestion above just fixes it on a regular computer browser view. Beyond that is too difficult to cater for IMO

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,232 @@
...
+@@long-footnote
+   Many existing uses of thread local state happen to be relatively robust and give correct answers only because the functions they are calling during execution do not yield. One may then think "well, I can just avoid this problem by making sure my code doesn't yield", but we think this is a bad and unsustainable idea, because whether or not a function call will yield is not stable, obvious, or easily inspectable.
+
+   For instance, if a function `f` is updated to include a background `@debug` statement or other forms of non-user-visible IO, it may change from being non-yielding to yielding. If during a call to `f`, the compiler encounters a dynamic dispatch where new code must be JIT compiled, a yield-point may be encountered, and any number of other internal changes could happen to code which can cause it to yielding.
+
+   Furthermore, future versions of julia may eventually move away from a [cooperative task model](https://en.wikipedia.org/wiki/Cooperative_multitasking) to a [preemptive task model](https://en.wikipedia.org/wiki/Preemption_(computing), in which case yield points would not be the only way that race conditions like this could be encountered.
```
giordano, line 222:
```suggestion
   Furthermore, future versions of julia may eventually move away from a [cooperative task model](https://en.wikipedia.org/wiki/Cooperative_multitasking) to a [preemptive task model](https://en.wikipedia.org/wiki/Preemption_(computing)), in which case yield points would not be the only way that race conditions like this could be encountered.
```

blog/2023/06/PSA-dont-use-threadid.md
IanButterworth: I don't think this should hold up publishing this. Whatever you think best @MasonProtter

MasonProtter:
Okay, I think this is ready to merge!

xwuupb:
I do not understand the logic in this post (https://julialang.org/blog/2023/07/PSA-dont-use-threadid/).

- in the first sample you are using the `@threads`-pattern for incorrect code, which is **hardly** reproducible.
- in the second sample you are using the `@sync`-`@spawn`-pattern to demonstrate the incorrectness, which is **surely** reproducible.

Could you please give an easily reproducible sample using the `@threads`-pattern? Thank!

Seelengrab:
`@threads` is more or less a thin wrapper around `@sync` `@spawn`. Neither is necessarily reproducible (in the sense that they end up with the same results every time) because of the potential presence of a race condition.

Could you clarify which part your thinking of when you say that the example is not reproducible?

xwuupb:
> Could you clarify which part your thinking of when you say that the example is not reproducible?

In the section "The buggy pattern" in this Blog (https://julialang.org/blog/2023/07/PSA-dont-use-threadid/) it reads

"The general template that this incorrect code follows is something like the following:" (then the `@threads`-pattern code).

The following Julia code using the general template (as mentioned in the Blog) is repeatedly tested via `julia -t 7 threads.jl` in a bash for-loop. However, the output does *not* change in different runs (at least in my tests).

```{julia}
using Base.Threads: nthreads, @threads, threadid

f(i) = (sleep(0.001); i)

states = [ 0 for _ in 1:nthreads() ]
N = 100
@threads for x in 1:N
  tid = threadid()
  old_val = states[tid]
  new_val = old_val + f(x)
  states[tid] = new_val
end
@show sum(states)
```

For the `@sync`-`@spawn`-pattern (the 2nd example in the Blog): the following Julia code is repeatedly tested via `julia -t 7 spawn.jl` in a bash for-loop. The output *does* change very frequently in different runs.

```{julia}
using Base.Threads: nthreads, @spawn, threadid

f(i) = (sleep(0.001); i)

let states = [ 0 for _ in 1:nthreads() ], N = 100
  @sync for i in 1:N
    @spawn begin
      tid = threadid()
      old_val = states[tid]
      new_val = old_val + f(i)
      states[tid] = new_val
    end
  end
  @show sum(states)
end
```

Thus, I mean *the general template* in the Blog is not reproducible. The `@threads`-pattern code above always gives correct results in my tests. More info about my tests:

- CPU:  AMD EPYC 7763 64-Core
- OS: Red Hat Enterprise Linux 8.8
- Julia: version 1.10.2
-

xwuupb:
> @xwuupb please, if you have any questions ask them on Discourse https://discourse.julialang.org/, a long merged pull request isn't a good place to have this discussions.

Now, the topic is moved here: https://discourse.julialang.org/t/psa-thread-local-state-is-no-longer-recommended-common-misconceptions-about-threadid-and-nthreads/101274/54

PR #1914, opened 2023-06-29
Preview #1904

IanButterworth:
Just a copy of #1904 as a repo branch given previews aren't generated for forks.

Do not merge


blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,175 @@
...
+                   tid = Threads.threadid()
+                   old_var = state[tid]
+                   new_var = old_var + f(i)
+                   state[tid] = new_var
+               end
+           end
+           sum(state), sum(1:N)
+       end
+(100, 5050)
+```
```
MasonProtter, line 51: This also isn't rendering correctly

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,175 @@
...
+maxtoclevel = 3
+title = "PSA: Thread-local state is no longer recommended; Common misconceptions about threadid() and nthreads()"
+authors = "Mason Protter, Valentin Churavy, Ian Butterworth, ..."
+published = "30 June 2023"
+rss_pubdate = Date(2023, 06, 30)
+rss = """PSA: Thread-local state is no longer recommended; Common misconceptions about threadid() and nthreads()"""
++++
+
+\toc
+
```
MasonProtter, line 12: the `toc` should probably have a listing for the first section. Maybe we can just do
```suggestion
\toc

# The buggy pattern
```
or something>

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,175 @@
++++
+mintoclevel = 2
+maxtoclevel = 3
+title = "PSA: Thread-local state is no longer recommended; Common misconceptions about threadid() and nthreads()"
```
MasonProtter, line 4: This is rendering pretty long and hard to read. Do we have a way to do sub-titles or something?

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,175 @@
...
+
+states = [some_initial_value for _ in 1:nthreads()]
+@threads for x in some_data
+    tid = threadid()
+    old_val = states[tid]
+    new_val = some_operator(old_val, f(x))
+    states[tid] = new_val
+end
+do_something(states)
+```
```
MasonProtter, line 28: For some reason this isn't rendering as code, I can't figure out why

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,175 @@
...
+This is a fully general replacement for the old, buggy pattern. However, for many applications this should be simplified down to a parallel version of `mapreduce`:
+```julia
+using Threads: nthreads, @spawn
+function tmapreduce(f, op, itr; init=some_initial_value, chunks_per_thread::Int = 2)
+    chunk_size = max(1, length(itr) ÷ chunks_per_thread)
+    tasks = map(Iterators.partition(itr, chunk_size)) do chunk
+        @spawn mapreduce(f, op, chunk; init=init)
+    end
+    mapreduce(fetch, op, tasks; init=init)
+end
```
MasonProtter, line 108: This'll make this more immediately runnable:
```suggestion
function tmapreduce(f, op, itr; chunks_per_thread::Int = 2, kwargs...)
    chunk_size = max(1, length(itr) ÷ chunks_per_thread)
    tasks = map(Iterators.partition(itr, chunk_size)) do chunk
        @spawn mapreduce(f, op, chunk; kwargs...)
    end
    mapreduce(fetch, op, tasks; kwargs...)
end
```

blog/2023/06/PSA-dont-use-threadid.md
```diff
@@ -0,0 +1,175 @@
...
+    3. Thread adoption (v1.9): Foreign threads can now be adopted (and later be removed) at any time during the execution of the program.
+    4. GC threads: The runtime can use additional threads to accelerate work like executing the Garbage Collector.
+
+    Any code that relies on a specific `threadid` staying constant, or on a constant number of threads during execution, is bound to be incorrect. As a rule of thumb, programmers should at most be querying the number of threads to motivate heuristics like how to distribute parallel work, but programs should generally **not** be written to depend on implementation details of threads for correctness. Rather, programmers should reason about *tasks*, i.e. pieces of work that may execute concurrently with other code, independently of the number of *threads* that are used for executing them.
+
+[^assoc]: ## Associativity
+    [Associativity](https://en.wikipedia.org/wiki/Associative_property) is an important property for parallel reducing functions, because it means that `op(a, op(b, c)) == op(op(a, b), c)`, and hence the result does not depend on the order in which the reduction is performed.
```
MasonProtter, line 173: more broken code formatting

blog/2023/06/PSA-dont-use-threadid.md
DilumAluthge: Maybe try putting an empty line after line 35 and before line 36?

IanButterworth: Already tried. This is outdated

tlienart:
Hello, the issues you're seeing are with the long multi-paragraph footnotes. The primary one is that there's a title in there (`# Uses of ...`) which Franklin doesn't expect and which messes everything up (in its present state Franklin is unable to tell you that it choked on such stuff which makes debugging hard, the pragmatic version to do it is to copy paste the content of a page, delete everything and add stuff in bits until you spot where the page starts to fail (here it was with the footnotes).

Secondary issue is with the styling of the footnotes; again Franklin is a bit limited with them because it uses a `<table>` for footnotes. For short ones this is ok, for big multi-paragraph ones, the content of the footnote is not placed appropriately. Long story short you  just had `Concurrency & Parallelism` actually part of the footnote and the rest was below it. This is not necessarily bad but the styling was a bit off. 

### Suggested replacement

Copy paste the following to replace everything after the paragraph on `MultiThreadedCaches.jl`. What it does is:

* some styling to have the footnote and the multi-paragraph content below it have a similar styling
* use `~~~<h4>~~~...~~~</h4>~~~` to get a title-like styling bypassing Franklin's management of section headings. 

<details>
```
~~~
<style>
.fndef {
    width: 100%;
    margin-bottom: 0;
    border: none;
}
.fndef-backref {
    width: 2em;
}
.long-footnote {
    width: 100%;
    background-color: var(--light);
    padding: 0.75rem;
}
</style>
~~~

[^historynote]: ~~~<h4>~~~Concurrency & Parallelism~~~</h4>~~~
@@long-footnote
   In Julia, tasks are the primitive mechanism to express concurrency. Concurrency is the ability to execute more than one program or task simultaneously.

   Tasks will be mapped onto any number of "worker-threads" that will lead them to be executed in parallel. This is often called `M:N` threading or green threads. Even if Julia is started with only one worker-thread, the programmer can express concurrent operations.

   In early versions of Julia, the `@async` macro was used to schedule concurrent tasks which were executed asynchronously on a single thread. Later, the `@threads` macro was developed for CPU-parallelism and allowed users to easily execute chunks of a loop in parallel, but at the time this was disjoint from the notions of tasks in the language. This lead to various composability problems and motivated language changes.

   The concurrency model in Julia has been evolving over minor versions. Julia v1.3 introduced the parallel scheduler for tasks and `Threads.@spawn`; v1.5 introduced `@threads :static` with the intention to change the default scheduling in future releases. Julia v1.7 enabled task migration, and Julia v1.8 changed the default for `@threads` to the dynamic scheduler.

   When the parallel scheduler was introduced, we decided that there was too much code in the wild using `@async` and expecting specific semantics, so `Threads.@spawn` was made available to access the new semantics. `@async` has various problems and we discourage its use for new code.

    ~~~<h4>~~~
   Uses of `threadid`/`nthreads`/`maxthreadid`
   ~~~</h4>~~~

   Over time, several features have been added that make relying on `threadid`, `nthreads` and even `maxthreadid` perilous:

   1. Task migration: A task can observe multiple `threadid`s during its execution.
   2. Interactive priority: `nthreads()` will report the number of non-interactive worker-threads, thus undercounting the number of active threads.
   3. Thread adoption (v1.9): Foreign threads can now be adopted (and later be removed) at any time during the execution of the program.
   4. GC threads: The runtime can use additional threads to accelerate work like executing the Garbage Collector.

   Any code that relies on a specific `threadid` staying constant, or on a constant number of threads during execution, is bound to be incorrect. As a rule of thumb, programmers should at most be querying the number of threads to motivate heuristics like how to distribute parallel work, but programs should generally **not** be written to depend on implementation details of threads for correctness. Rather, programmers should reason about *tasks*, i.e. pieces of work that may execute concurrently with other code, independently of the number of *threads* that are used for executing them.
@@

\\

[^assoc]: ~~~<h4>~~~Associativity~~~</h4>~~~
@@long-footnote
   [Associativity](https://en.wikipedia.org/wiki/Associative_property) is an important property for parallel reducing functions, because it means that `op(a, op(b, c)) == op(op(a, b), c)`, and hence the result does not depend on the order in which the reduction is performed.

   Note that associativity is not the same as commutivity, which is the property that `op(a, b) == op(b, a)`. This is *not* required for parallel reducing functions.
@@
```
</details>

PS: I've been working for a long time now on a rewrite of Franklin which fixes quite a few of these things, it's basically finished apart for docs adjustments. This is just to say that I'm well aware of existing Franklin warts, and even though the next version will likely also have its quirks, it should be generally easier to use and debug (at least that's what I hope).

PR #1923, opened 2023-07-09
chunks_per_thread to tasks_per_thread

themantra108:
use either tasks_per_thread or chunks_per_thread in function tmapreduce


LilithHafner:
The example above uses `tasks_per_thread`, so probably best to stick with that for consistency.

LilithHafner:
Thanks!

PR #1939, opened 2023-09-24
Over 10,000 packages

PallHaraldsson:


PallHaraldsson:
We are actually 3 packages from 10,000, but at least 6 of the 65 waiting in the queue should get registered (or not blocked) in next 3 days. Before you merge, can anyone look up how many downloads? I.e. new numbers, not just from juliahub.com? It can be a separare PR, or I can add here. I find likely we are now over 50 million.

We ARE over 10,000 (usable) packages, but not quite than many registered yet, so if you want to wait for that, fine, but it seems like a technicality, in case you want to merge as soon as you see this, then we might be passed the mark. [I intend to announce on Discourse.]

PallHaraldsson:
Bump.

No need to wait with merging for e.g. new downloads numbers, but they should also be updated, and "Over 10,000" at Juliahub.com.

Another thing, we have "JuliaCon 2022", and it's likely good to show one from 2023 or cycle through.

PallHaraldsson:
Bump.

PR #1942, opened 2023-10-02
Clarify compiled - Update index.html

PallHaraldsson:
[skip ci]

### DIFF
--- a/index.html
+++ b/index.html
@@ -65,13 +65,13 @@
       <!-- 1 -->
       <div class="col-lg-4 col-md-6 feature">
         <h3>Fast</h3>
-        <p>Julia was designed from the beginning for <a href="/benchmarks/">high performance</a>.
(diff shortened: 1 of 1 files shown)


PallHaraldsson:
I don't see how this look (I wanted short, still in three lines, why I skipped "from the beginning", which seems implied and of less value?). Compilation implies to many separately compiled, i.e. more work (which is now also possible).

ViralBShah:
Let's see how this looks.

PallHaraldsson:
I was aiming for 3 lines... so could do: "via LLVM, and support multiple platforms." -> "via LLVM; support multiple platforms."  Maybe even skip "via LLVM", it's an implementation detail, and I would argue either way about it.

"language, and has" -> "language; has"

And/or possibly smaller font? Otherwise I'm ok as is.

PR #1953, opened 2023-10-24
Update number of packages

ViralBShah:


PallHaraldsson:
Great! I missed changing there... and note, still at juliahub.com "9,500+ registered packages" and also likely outdated: "45 mn downloads JULY 2023". I suppose 50+ now, but do not want to quote that without confirmation.

PR #1963, opened 2023-11-15
Update organizations.md

arhik:
I am author of `JuliaWGPU` organization and would like to share it with community.
- added `JuliaWGPU` organization currently being maintained by two members.
- took liberty to add hyperlink to `organizations.md` file.


IanButterworth:
@arhik sounds good to me and thanks for the PR note.

As caught by the check, it seems `JuliaWGPU` currently only has one public member? 
Do you think your org colleague could also be public? 

The default is for org membership to be private on GitHub, so people often don't opt-in.

arhik:
Hi. Now we are two public members.

Issue #1922, opened 2023-07-08
Juliaup rollout


davidanthoff:
I'm starting to think through all the necessary changes we need to do to make Juliaup the official way to install Julia, and an important part is the homepage and documentation. 

Roughly, here is what I would suggest:
1. We [add](https://github.com/JuliaLang/julia/pull/50477) two new sections to the Julia manual: "Installation" and "Juliaup". The first one describes in detail all the installation options we have, the second one documents all the ins-and-outs of Juliaup.
2. We add a top level "Get Started" section to the Julia home page, https://julialang.org/get-started/.
3. We model that page after https://www.rust-lang.org/learn/get-started, i.e. the first part is a platform specific install instructions (short and sweet) with a link to the comprehensive installation instructions in the manual. So, this would display differently depending on which platform one opens that page. I think down the road we could also add suggestions about editors etc, but maybe for starters it could really just be "Do X to install" and then "Read the Getting Started section in the manual" with a link.
4. We replace the "Downloads" top level navigation element on the home page with a link to the new "Get Started" page.
5. On the homepage, we also replace the green "Download" link with one to the "Get Started" page.
6. We _keep_ the https://julialang.org/downloads/ page alive, though, essentially in its current form, but we just deemphasize it. I'm not entirely sure from which pages we would link to it, maybe still from the "Getting Started" page? But essentially at the top we would have text along the lines of "This page has an archive of downloads, to install Julia on a system please follow the instructions on the Get Started page".
7. We (well, I) write a blog post that announces this change.

Thoughts? Does that make sense?

If someone wants to help with making these changes it would be great!

ExpandingMan:
We've been having a long conversation about this on slack spurred by a new user who, entirely justifiably, got confused largely over broken and out of date Julia linux distro packages.  I think it would be very beneficial to do something along the lines of what @davidanthoff is suggesting.

In my experience, most of the relatively new programming languages are in the same boat as Julia, having patched dependencies that make the build process less than straightforward for package managers.  Rust, which has already been mentioned, also has [an install page](https://www.rust-lang.org/tools/install) which prominently mentions rustup.  Haskell, in which the recommended install method is `ghcup`, [is similar](https://www.haskell.org/downloads/).

hoegge:
[...]

PR #1969, opened 2023-11-20
Recommend juliaup

IanButterworth:
For when it is time to recommend juliaup over the downloads table.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -1,15 +1,37 @@
 # Download Julia
 
-~~~
-<a class="github-button" href="https://github.com/JuliaLang/julia" data-size="large" data-show-count="true" aria-label="Star JuliaLang/julia on GitHub">Star</a>
-~~~
-
 Please star us [on GitHub](https://github.com/JuliaLang/julia). If you use Julia in your research, please [cite us](https://julialang.org/research/). If possible, do consider [sponsoring](https://github.com/sponsors/julialang) us.
+
+
+## Juliaup: The recommended Julia installation manager
+
+To install Julia it is recommended to use the `juliaup` installation manager, which will automatically install
+and manage keeping julia up to date.
+
+Installation on Windows is either via the Windows store [here](https://www.microsoft.com/store/apps/9NJNWW8PVKMN)
+or via the command prompt:
+```
+winget install julia -s msstore
+```
+
+On Linux and MacOS:
+```
+curl -fsSL https://install.julialang.org | sh
+```
+
+Once installed `julia` and `juliaup` will be available via the command line interface.
+
+To install different versions see `juliaup --help` or the documentation at [github.com/JuliaLang/juliaup](https://github.com/JuliaLang/juliaup).
+
 
 ---
 
+## Julia Version History
+
+Note that all Julia versions are installable through Juliaup.
+
 ~~~
-<h2 id=current_stable_release><a href="#current_stable_release">Current stable release: v{{stable_release}} ({{stable_release_date}})</a></h2>
+<h3 id=current_stable_release><a href="#current_stable_release">Current stable release: v{{stable_release}} ({{stable_release_date}})</a></h3>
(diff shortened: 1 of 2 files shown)


LilithHafner:
See also: https://github.com/JuliaLang/juliaup/milestone/7

downloads/index.md
```diff
@@ -1,19 +1,37 @@
...
+```
+
+On Linux and MacOS:
+```
+curl -fsSL https://install.julialang.org | sh
+```
+
+Once installed `julia` and `juliaup` will be available via the command line interface.
+
+To install different versions see `juliaup --help` or the documentation at [github.com/JuliaLang/juliaup](https://github.com/JuliaLang/juliaup).
```
ericphanson, line 24: IMO it would be good to have a suggestion here, like

```sh
juliaup add 1.9
juliaup default 1.9
```

So folks can get to a working Julia session without visiting more documentation if they don't need anything fancy. (The hardcoded version could be updated each release, as this page is already updated each release).

IanButterworth: Doesn't juliaup install latest automatically?

davidanthoff: Yes, you automatically get the latest version.

ericphanson: ah ok!

downloads/index.md
```diff
@@ -1,15 +1,37 @@
...
+
+
 ---
 
+## Julia Version History
```
LilithHafner, line 29:
```suggestion
## Manual Download Links
```

downloads/index.md
```diff
@@ -1,15 +1,37 @@
...
+
+Note that all Julia versions are installable through Juliaup.
```
LilithHafner, line 31:
```suggestion
Note that all officially released Julia versions since 0.7.0 are installable through Juliaup.
```
https://github.com/JuliaLang/juliaup/issues/96

downloads/index.md
```diff
@@ -256,25 +278,21 @@ Checksums for this release are available in both, [MD5](https://julialang-s3.jul
...
 
-## Download verification
+### Download verification
 
 All Julia binary releases are cryptographically secured using the traditional methods on each platform. macOS and Windows releases are codesigned with certificates that are verified by the operating system during installation. Linux and source tarballs are signed with GPG using [this key](/assets/juliareleases.asc).
 
 ---
 
-## Supported platforms
+### Supported platforms
```
LilithHafner, line 295: Perhaps this should go above the manual download links

LilithHafner (approved):
Juliaup is not fully mature and there are unresolved issues on the release blocker milestone. But if a friend asked me how to install Julia, I would tell them to use juliaup because, despite its imperfections, it is already significantly better than the alternatives. Consequently, I want to merge this PR.

LilithHafner:
See also: #1922

IanButterworth:
@davidanthoff whats your current perspective on when juliaup is ready to get recommended like this?

I'm just thinking about the 1.10 release coming up

downloads/index.md
IanButterworth: I gave this a go but I think it's helpful for the version history to be within the viewpane so that it's still easy to see what the current version is.

StefanKarpinski (approved):

IanButterworth:
I think it's fair to summarize that julia devs feel that juliaup has basically already become the recommended method, so whilst there's some issues left to resolve, it seems to work well and it would be natural for it to be recommended here, so let's go ahead.

davidanthoff:
Fair enough :) I think in terms of things that really don't work right now, the only thing I can think of is depots on network file systems, i.e. clusters. Juliaup uses file system locks, and they don't work on network file systems, so that scenario just doesn't work.

I had a very busy semester and didn't have time for much Julia stuff at all, but I'm not teaching in the spring, so hopefully should have much more time to finish up the remaining open issues.

PR #1975, opened 2023-12-03
Update `greet_your_package_name` in developing_package.md to make tests pass 

Ian-Harreschou:
When `greet_your_package_name()` is defined, the `println("Hello YourPackageName!")` creates an object of type `::Nothing`. So the testset created later on never passes, because it is comparing an object of `::String` to one of `::Nothing`


LilithHafner:
Thanks!

PR #1977, opened 2023-12-04
Move protective comment closer to the content it refers to [nfc]

LilithHafner:
This disclaimer was moved to the top and comment added in #1432 cc @StefanKarpinski 
And then the note was moved to the bottom without the comment in #1456 cc @ViralBShah 

The disclaimer and comment should be together. If it's unlawful to have the note at the bottom, then the comment should say so. I'm not a lawyer this isn't legal advice.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -8,10 +8,6 @@ Please star us [on GitHub](https://github.com/JuliaLang/julia). If you use Julia
 
 ---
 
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
Yeah, was originally right above but, of course, got ignored when someone moved the notification about IP addresses 😂😭

PR #1978, opened 2023-12-04
Factor out "xx-bit" in downloads table (subset of #1770)

LilithHafner:
This is a hopefully less bikesheddable version of #1770. It has slightly more information, is slightly more concise/readable, and introduces no new terminology.


giordano:
Missed the nightlies page: https://julialang.org/downloads/nightlies/?

PR #1979, opened 2023-12-04
Update "Julia Computing" => "JuliaHub" in non-blog post content

LilithHafner:
The name Julia Computing is also used extensively in blog posts published before the name change. I do not think it is appropriate to edit those blog posts for this reason.


StefanKarpinski:
As long as it's indicated that they're edits, I think updating blog posts is fine, e.g. to fix links and such.

LilithHafner:
I agree. See https://github.com/JuliaLang/www.julialang.org/pull/1985

PR #1983, opened 2023-12-05
Fix typos

LilithHafner:


LilithHafner:
There's also a fair bit of removal of trailing whitespace—incidental but still beneficial.

PR #2001, opened 2023-12-31
Reorganize downloads

LilithHafner:
- Move latest stable release number higher on the page (on my default full-screen resolution this brings it from below the fold to above the fold)
- Stop recommending jill.py
- Move juliaup to the top of platform specific instructions instead of the bottom
- Place support tiers above manual instillation
- Minor reorderings
  - Put combined Linux & Mac juliaup instructions above Windows instead of below (they were below because, historically, JuliaUp was a windows-only tool.
  - Bring Linux to the top of the platform specific manual instillation instructions because Linux users are imo the most likely to choose manual instillation over the recommended juliaup 
  - Place SHA256 before MD5 because it's cryptographically secure


IanButterworth:
LGTM but can we merge this first? https://github.com/JuliaLang/www.julialang.org/pull/2005

IanButterworth (approved):

IanButterworth:
Let's address @KristofferC's comments here https://github.com/JuliaLang/www.julialang.org/pull/2005#issuecomment-1875043315

KristofferC:
I noticed now I actually made a mistake when reading the note about Juliaup... I thought it said "**The** recommended installation method is via the Juliaup"... but it actually says "**This** recommended installation method is via the Juliaup...". So it isn't as "bad" as I originally thought.

IanButterworth:
Yeah I was trying to get in "recommended" but I think that's implicit so doesn't need to be there.

IanButterworth:
As for the width, do you think it should go full width? Originally I had it narrower. It's supposed to stand out like a call-out. If it's full width the box padding will still mean the text is inset from the text below and on a mobile browser the edge of the box is up against the edge of the screen.

KristofferC:
Maybe it is fine the way it is :shrug:. Could possibly work better by having some delimiter between the box and the rest of the text or maybe just some extra vertical whitespace to make it feel a bit less connected to the text under it. But feel free to ignore.

IanButterworth:
All good suggestions. Thanks. Updated.

> and on a mobile browser the edge of the box is up against the edge of the screen.

This actually isn't true. The edge is inset.

IanButterworth:
This looks good to me now. @LilithHafner I'll let you final review and merge given I made changes

PR #2002, opened 2023-12-31
Fix some broken links

LilithHafner:
I have reached out to Jose Storopoli (@storopoli), Rik Huijzer (@rikhuijzer), and Lazaro Alonso on Zulip to check for a new link to their book.

I've also reached out to @logankilpatrick about another book 404, but have not those changes in this PR.

The twitter package feed gave a 403 error so I switched it to Zulip which is public & open source.

This PR can merge before hearing back from those folks.


rikhuijzer:
Hi @LilithHafner 

https://github.com/JuliaDataScience/JuliaDataScience/blob/080683b6c4878f9816aa825579b6e8365c867b77/README.md?plain=1#L17-L21

has more information about the current status. Thanks for double checking the links

storopoli (approved):
Thanks for the corrections.
We could also implement a GH Actions workflow like this: https://github.com/storopoli/storopoli.github.io/blob/main/.github/workflows/links.yml that has a cron scheduling that tests all links in the repo.

LilithHafner:
Yep! Though we should either perform a diff and only report new broken links or remove all the links to websites that track their users and refuse to load if the tracking services are not enabled (the broken link checker I'm using struggles to bypass those tracker-detectors)

LilithHafner:
This doesn't get everything, but I'm confident in the diff so far and I'm going to merge now and keep looking for more for another PR.

Issue #658, opened 2020-03-13
Simplify the download page so it's less intimidating 


musm:
- The download page is too intimidating for newcomers. Going to the download page you are hit with a complex wall of text and confusing matrices of downloads configurations.

![image](https://user-images.githubusercontent.com/4319522/76653703-98da2680-653f-11ea-9feb-e304bb9fa934.png)

----

- Let's simplify things for the first half of the page. We can keep the granular matrix of downloads further down the page, but users should be greeted with a simple download page that is not complicated.

- We should aspire to the following example

![image](https://user-images.githubusercontent.com/4319522/76653690-91b31880-653f-11ea-9dd9-02cdc5c14066.png)

@logankilpatrick

cormullion:
I just made this using a grid:

<img width="887" alt="Screenshot 2020-03-19 at 18 17 29" src="https://user-images.githubusercontent.com/52289/77100904-4cdf2400-6a0e-11ea-8475-88bec4a6f348.png">

Sort of thing? The [help] could be a question mark I suppose...

tlienart:
Nice, maybe a  one click button for the standard option would be good? (if we're going with something similar to VS Code). I guess that button could wrap the symbol so that if you click on it you get the most recent stable release that is most likely to  correspond to  your platform. 

Personal preference would also be for the icons to be smaller, maybe 70% of the size you show 😄

musm:
Yeah Looks good, but the icons are kinda massive and their not all the same size. Is there a better linux photo? It's weird seeing a B&W penguin :) 

For the windows one can we use a B&W or color version like this: 
![image](https://user-images.githubusercontent.com/4319522/77102256-eac2f680-69ee-11ea-89ed-db547ab20401.png)

fredrikekre:
Can we just detect the running system and present just a single button? I know that is done for other projects.

musm:
I don't think we should present a single button because there have been several times when I've wanted to download e.g. linux binaries for WSL. Having them all present makes it very user friendly to quickly grab and test the binaries you want.

fredrikekre:
I mean obviously the other ones shouldn't be hidden, but we can present a main one. The regular user probably wants that one.

Here is an example: https://ngrok.com/download

cormullion:
I avoided color just because it would be a bit inconsistent- with some in color and others not. (I didn't bother to size everything accurately at this stage... :))

musm:
Here's a mono linux logo that looks much better

 

[...]

PR #2005, opened 2024-01-02
Make juliaup instructions simpler

IanButterworth:
Simplifies the instructions and makes them automatically detect platform (with an override)

Closes https://github.com/JuliaLang/www.julialang.org/issues/658

PR
https://julialang.netlify.app/previews/pr2005/downloads/

Currently
Screenshot 2024-01-01 at 8 53 57 PM

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -687,3 +687,7 @@ pre {
 	outline: 0;
 	background-color: #eee9e6;
 }
+
+.install-platform-note {
+  font-size: 0.85rem;
+}
--- a/_css/app.css
+++ b/_css/app.css
@@ -689,11 +689,10 @@ pre {
 }
 
 .main-download-instructions {
-  height: 70vh;
-  min-height: 470px;
   display: flex;
   align-items: center;
   justify-content: center;
+  margin-bottom: 30px;
 }
 .main-download-instructions-inner {
   width: 80%;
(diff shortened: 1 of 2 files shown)


StefanKarpinski (approved):

LilithHafner:
I have not verified that the platform detection works on windows, but from Linux this GUI looks much better than before. Thanks!

giordano:
It'd be nice to see the instructions for another operating system without having the other operating system.  Like, having multiple tabs for the different OSes, with the current one open by default.

giordano:
One example: https://rustup.rs/ (but I'd prefer the tabs instead of this).

IanButterworth:
There's a link underneath the code block for that. Did you not see that?

IanButterworth:
Discussed with @giordano on slack and he had missed the existing text. Tabs seem like a good idea but that could be tried out in another PR.

As for windows, I checked in a cross-platform browser checker and it looks good
Screenshot 2024-01-02 at 4 55 36 PM

KristofferC:
The

> This recommended installation method is via the [Juliaup](https://github.com/JuliaLang/juliaup) installation manager, which will automatically install julia and help keep it up to date.

feels unclear to me when it comes after the installation instructions. Because it isn't clear that the above command is in fact installing and using Julia up. Maybe it could be something like:


> This will install the [Juliaup](https://github.com/JuliaLang/juliaup) installation manager, which will itself automatically install julia and help keep it up to date.

Issue #775, opened 2020-05-05
Respect dark mode preference


Keno:
I was looking at https://www.youtube.com/watch?v=VgZm53qgj9Q, which starts with the julia web-page but using some sort of plugin that turns it to dark mode. I'm wondering if we can do something to make the logo look better for people who do similar things.

tlienart:
That's a job for @cormullion 😂

cormullion:
Yes, I noticed some dark magic going on in that video - some kind of browser add-on that forces everything to dark mode. Is that @JeffBezanson 's computer?

cormullion:
I suppose the solution is to add a dark mode switch to julialang.org, then people won't need to a dark-mode plugin that doesn't invert images. That's a job for - well, I know it's tricky, since Slack took years to do it... :)

cormullion:
see also https://github.com/JuliaDocs/Documenter.jl/issues/1320

ViralBShah:
It's the browser settings of the person who made the video.

dillir07:
I believe its due to any addon/extension that changes the background. Ex: I use [DarkReader](https://darkreader.org/) in firefox.

![Darkreader in firefox](https://user-images.githubusercontent.com/47918989/95594159-3101a500-0a68-11eb-96c4-a408251f616d.png)

Keno:
I don't suppose there's a way to detect extensions like this and automatically provide a dark mode logo?

Keno:
Apparently there's now a CSS media query for dark mode preference: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme. That seems like something we could definitely support.

cormullion:
I'd added a similar issue to Documenter.jl - https://github.com/JuliaDocs/Documenter.jl/issues/1320 - in case someone wanted to make the switch automatic.

dillir07:
This may not be anywhere near the solution, but just out of curiosity I modified julia's black color in svg. It's same svg and looks okay in both light & dark background.
![julia light dark](https://user-images.githubusercontent.com/47918989/95648777-8042e680-0af7-11eb-8f16-d96c043ba91a.png)

ViralBShah:
The JuliaLogoGraphics repo has official logos for dark background.

https://github.com/JuliaLang/julia-logo-graphics

PR #2011, opened 2024-01-06
switch to dark julia logos if darkreader dark mode detected

IanButterworth:
Fixes https://github.com/JuliaLang/www.julialang.org/issues/775

With darkreader on
Screenshot 2024-01-05 at 9 03 32 PM

Without
Screenshot 2024-01-05 at 9 03 39 PM


IanButterworth:
This works and in the absence of darkreader shouldn't have any noticeable effect. So I'll give this a go

Issue #2006, opened 2024-01-02
Homepage does not serve valid html


LilithHafner:
`curl https://www.julialang.org | grep "</html>"`, for example, reveals that the html tag is never closed.

LilithHafner:
Other problems: https://validator.w3.org/nu/?doc=https%3A%2F%2Fjulialang.org%2F (maybe these are Franklin.jl bugs?)

StefanKarpinski:
cc @tlienart

tlienart:
the landing page is actually not really handled by Franklin that much, the source of that page is mostly plain HTML  https://github.com/JuliaLang/www.julialang.org/blob/main/index.html 

there are things that are off though, that `</html>` is in the source and gets eaten up, I suspect it's the minifier that's not working well here (not a Franklin thing per se and can be disabled). 

First thing that can be done would be to disable the minifier by changing

https://github.com/JuliaLang/www.julialang.org/blob/c35d1b42090ce631a46a221996446d3ecc9c6700/.github/workflows/deploy.yml#L61

to 

```
Franklin.optimize(prerender=true, minify=false);
```

this will have no impact other than un-garbling the HTML (which won't serve any slower as GH pages optimises things for us). https://github.com/JuliaLang/www.julialang.org/pull/2013

I'll take a look at the rest when I can, some of these errors are simple to handle like the `meta` tag with name duplication (can be fixed https://github.com/JuliaLang/www.julialang.org/blob/main/_layout/meta.html)

fredrikekre:
The closing html tag being removed is most likely on purpose since it is an optional tag, see https://www.w3.org/TR/2014/REC-html5-20141028/syntax.html#optional-tags

LilithHafner:
Yes. Dropping the closing tag is valid. However, also noting that the tag is present on other pages (e.g. https://www.julialang.org/downloads), that omitting the tag saves very little bandwidth, and that there are other issues that make the html invalid.

LilithHafner:
Also, I think it's nice to serve valid XML if it's not too much trouble (and xml requires matched tags).

fonsp:
I don't think we need a minifier, since the content is already being gzipped by fastly. 

[...]

PR #2013, opened 2024-01-11
Update deploy.yml to remove problematic minifier step

tlienart:
The minifier that Franklin leans on is buggy and, in any case, github pages optimises stuff for us so it's not needed.

In the context of #2006, it will help to disable it to see the generated HTML more clearly.

impact can be assessed here: https://github.com/JuliaLang/www.julialang.org/tree/gh-preview/previews/PR2013 where everything is generated properly and e.g. the `index.html` is not garbled anymore (and does actually have a closing html tag) https://github.com/JuliaLang/www.julialang.org/blob/gh-preview/previews/PR2013/index.html

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -58,7 +58,7 @@ jobs:
               using NodeJS; run(`$(npm_cmd()) install highlight.js`);
               using Franklin;
               Franklin.HIGHLIGHTJS[] = abspath(joinpath("_libs", "highlight", "highlight.min.js"));
-              Franklin.optimize(prerender=true);
+              Franklin.optimize(prerender=true, minify=false);
               cp(joinpath("__site", "feed.xml"), joinpath("__site", "index.xml"))' > build.log
           cat build.log
 


LilithHafner (approved):
Skipping minification does have a performance impact, even with gzip. However, I find the justification here and in https://github.com/tlienart/Franklin.jl/issues/836 and https://github.com/JuliaLang/www.julialang.org/issues/2006#issuecomment-1895604913 to be worth that performance cost.

`curl https://julialang.org/ | wc -c`
39252 characters
3083 bytes sent
47404 bytes received

`curl https://julialang.netlify.app/previews/pr2013/ | wc -c`
47558 characters (+21%)
3773 bytes sent (+22%)
56056 bytes received (+18%)

When using the `--compressed` flag,

`curl --compressed https://julialang.org/ | wc -c`
39252 characters
2238 bytes sent
16771 bytes received


`curl --compressed https://julialang.netlify.app/previews/pr2013/ | wc -c`
47558 characters (+21%)
2584 bytes sent (+15%)
16886 bytes received (+0.7%)

PR #2044, opened 2024-02-03
Add a new book by A. Varga to books.md

andreasvarga:
New book by A. Varga


andreasvarga:
I am not sure I did all right (just klicked following the indications).

PR #2048, opened 2024-02-07
Update Trixi.jl GSoC projects

ranocha:
We have already someone working on the GPU project outside of GSoC. Thus, it's better to remove this project. The other project is still up to date.

CC @sloede


sloede (approved):
LGTM, thanks!

PR #2053, opened 2024-02-14
Set stable release to 1.10.1

ararslan:
Also comment out PowerPC for stable release binaries because they're not currently available for PPC.

I have no idea why the old releases script dumped 1.9.0-rc1, will investigate (✅)

### DIFF
--- a/config.md
+++ b/config.md
@@ -13,9 +13,9 @@ author = ""
 # Templating of the Downloads
 # NOTE: When updating for a new release, make sure to also rerun
 # `downloads/oldreleases.jl`
-stable_release = "1.10.0"
+stable_release = "1.10.1"
 stable_release_short = "1.10"
-stable_release_date = "December 25, 2023"
+stable_release_date = "February 13, 2024"
 lts_release = "1.6.7"
 lts_release_short = "1.6"
 lts_release_date = "July 19, 2022"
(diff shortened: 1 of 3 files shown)


ararslan:
The apparent removal of v1.9.0-rc1 in oldreleases.md is actually because it is currently duplicated for whatever reason, and regenerating the file removed the duplicate. So all good.

KristofferC:
I think there was a bad tag of 1.9.0-rc1 (without the `v` prefix). That tag has been deleted so probably because of that.

ararslan:
Something weird is going on with the build and deploy CI step: when generating the preview, it said "no space left on device" and hung so I canceled and retried, and it's now (seemingly) hung again.

ararslan:
I believe the `gh-preview` branch is keeping around a bunch of copies of the repo, one per PR: https://github.com/JuliaLang/www.julialang.org/tree/gh-preview/previews. That seems... suboptimal.

KristofferC:
Reset it to master and force push to it to clean it up maybe?

ararslan:
Kristoffer IRL

brain

giordano:
I guess you want to adapt https://github.com/JuliaCon/www.juliacon.org/blob/8f3d89329ee9b0b592cb73ad813ff3b5796476c5/.github/workflows/cleanup.yml

PR #2057, opened 2024-02-21
Taija gsoc

pat-alt:
This PR:

- Adds two project proposals for GSoC for [Taija.jl](https://github.com/JuliaTrustworthyAI).
- Adds a project proposal for JSoD for Taija.jl. (I have left the existing proposal by @kdpsingh for 2023 in there for now. Should I remove that?)

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -1,18 +1,31 @@
 # This file is machine-generated - editing it directly is not advised
 
+[[ArgTools]]
+uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
+version = "1.1.1"
+
 [[Artifacts]]
-deps = ["Pkg"]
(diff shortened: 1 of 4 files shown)


kdpsingh:
Thanks for asking. I don't have bandwidth to oversee someone this year so would suggest taking it down. Before making a final call, let me ask @rdboyes if he would like to repurpose this proposal to help with TidierPlots documentation, contingent on him being able to supervise if the proposal is selected.

rdboyes:
Thanks for thinking of me here, but I a. don't think I have the bandwidth either, and b. think that TidierPlots is about to undergo a large rewrite so it might not be the best time to document everything. Next year maybe?

pat-alt:
Thanks both, I've updated accordingly. 

@TheCedarPrince I believe I should be tagging you here? 😃

PR #2072, opened 2024-03-11
Update classes.md with NUS EE5311

mchitre:
Added NUS course EE5311 to list of courses using Julia


TheCedarPrince:
Merged -- thanks @mchitre !

PR #2089, opened 2024-04-18
Tell people not to use their package manager's "Julia"

LilithHafner:
I just got off the phone with someone running into issues with Julia 1.7, recently installed courtesy of their package manager.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -52,6 +52,10 @@ Please star us [on GitHub](https://github.com/JuliaLang/julia). If you use Julia
 
 ---
 
+### Please do not use the version of "Julia" shipped by unix package managers
+
+Many unix package managers ship broken and/or significantly out of date versions of Julia. Please use juliaup or download the official binaries instead.
+
 ### Supported platforms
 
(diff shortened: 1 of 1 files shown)


StefanKarpinski (approved):
Might word it to more strongly emphasize that juliaup is recommended and also mentions that using your distro's juliaup package and then letting that install official Julia binaries is a fine way to go.

LilithHafner:
We already place Juliaup above the official binaries, and I don't want to do too much more than that given the state of https://github.com/JuliaLang/juliaup/milestone/7. If someone is deciding whether or not to use their distro's broken 1.7.1, official binaries and juliaup are both a bit step up.

As for juliaup via package managers/distros, I'm hesitant to recommend that given

> **Important note:** As of now, we strongly recommend to install Juliaup via the Windows Store or `curl` command above rather than through OS-specific software repositories (see below) as the Juliaup variants provided by the latter currently have some drawbacks (that we hope to lift in the future).

https://github.com/JuliaLang/juliaup?tab=readme-ov-file#software-repositories

StefanKarpinski:
Ok, carry on then!

PR #2090, opened 2024-04-22
Update organizations.md

NeroBlackstone:
Add JuliaMusic organization


NeroBlackstone:
Check for any listed orgs with fewer than 2 public members failed.

Do I need to continue editing the file to remove these organizations?

ViralBShah:
@pkofod Is it possible to have a second public member in the JuliaNLSolvers org?

PR #2098, opened 2024-05-07
Change from GoogleAnalytics to Plausible

aviks:
For a more privacy friendly approach. Will obviate the need to use cookie consent dialogs.


StefanKarpinski:
Nice!

vchuravy (approved):

PR #2102, opened 2024-05-29
"Official Domains": mention additional Juliaup domains

DilumAluthge:
The first time that the user installs Juliaup, the Juliaup binary is downloaded from a URL of the form `https://github.com/JuliaLang/juliaup/releases/latest/download/*`. [^1]

[^1]: See [line 84 of this file](https://github.com/JuliaLang/juliaup/blob/13fa656f43c2feb9d85882def899eeda968537b9/deploy/shellscript/juliaup-init.cmd#L84) for the place where this URL is defined.

Therefore, the first time that a user installs Juliaup, they need to be able to make network requests to the `github.com` domain.

This PR documents this requirement in the "Official Domains" section of the "Download" page.


downloads/index.md
```diff
@@ -485,3 +485,15 @@ The following domains are official and used by open source Julia infrastructure
 - `julialang.net` and all subdomains
 
 If you are using Julia behind a firewall that blocks access to these, you may have trouble downloading and installing Julia packages. If this is the case, please ask your sysadmin to add these domains to the firewall allow list. Traffic can be limited to HTTPS (TCP port 443).
+
+#### Additional domains used only by Juliaup
+
+If you are using Juliaup, then you need access to the following domains when you install Juliaup for the first time:
```
StefanKarpinski, line 491:
```suggestion
If you are using `juliaup`, then you need access to the following domains when you install `juliaup` for the first time:
```

downloads/index.md
```diff
@@ -485,3 +485,15 @@ The following domains are official and used by open source Julia infrastructure
...
+
+- `github.com`
+
+Traffic can be limited to HTTPS (TCP port 443).
+
+This is only required if you are using Juliaup.
```
StefanKarpinski, line 497:
```suggestion
This is only required if you are using `juliaup` (recommended but not required).
```

downloads/index.md
```diff
@@ -485,3 +485,15 @@ The following domains are official and used by open source Julia infrastructure
...
+
+If you are not using Juliaup, then you do not need to have access to the `github.com` domain.
```
StefanKarpinski, line 499:
```suggestion
If you are not using `juliaup`, then you do not need to have access to the `github.com` domain.
```

LilithHafner (approved):

DilumAluthge:
I'm now [told](https://discourse.julialang.org/t/changes-to-pkgservers-storage-backend/112378/22?u=dilumaluthge) on [Discourse](https://discourse.julialang.org/t/changes-to-pkgservers-storage-backend/112378/25?u=dilumaluthge) that the above-referenced line is dead code that is never actually run.

Thus we don't actually need the note added in this PR.

PR #2109, opened 2024-06-09
added neovim to Tools list

mostafatouny:
I am looking forward to create a more accessible tutorial for configuring neovim for Julia. For now, This is a good start.


IanButterworth (approved):
I'm not that familiar with this but given it links to an established wiki on a core package it seems good to me.

PR #2137, opened 2024-08-04
Add stewards to footer

LilithHafner:
Lots of folks at JuliaCon didn't know the stewards are a thing. It is important for folks to know how to seek help in case of adverse behaviors on Julia discussion forums.

Adding directly below community standards because that is the primary domain of the stewards.


TheCedarPrince (approved):
This looks great @LilithHafner , thanks for adding this! Merging in!

> **Thought for Future:** Someday in the future it would be great to somehow draw more attention to Code of Conduct and Stewards and make it more prominent somehow but I am not sure how to make that look.

Issue #2136, opened 2024-08-04
Homepage last modified date 0001-01-01


LilithHafner:
**Describe the bug**
The homepage shows "Last modified: 0001-01-01. This site is powered by [Netlify](https://www.netlify.com/), [Franklin.jl](https://franklinjl.org/)..."

Other pages are fine.

**To Reproduce**
1. Go to 'https://julialang.org/
3. Scroll down to footer

ViralBShah:
Perhaps relevant? https://github.com/tlienart/Franklin.jl/issues/700

LilithHafner:
I don't think the solution in that issue fixes this because other pages which use the same `Last modified: {{ fill fd_mtime }}.` invocation are fine. I'm guessing it has to do with special handling we give the homepage for faster load time.

IanButterworth:
I think we just need to edit the homepage?

IanButterworth:
I edited index.html in #2157 but the home page date is still bad on the preview and now deployed version.

@tlienart any idea?

ViralBShah:
Maybe in the meanwhile, we can just remove the "Last modified" bit?

ViralBShah:
I think we should remove the "Last Modified" until we have a fix.

LilithHafner:
Are you okay with removing it on all pages? Right now it is in footer, not the homepage specifically.

PR #2155, opened 2024-10-04
Make trivial change to index.html

LilithHafner:
Tests https://github.com/JuliaLang/www.julialang.org/issues/2136#issuecomment-2394504400


IanButterworth:
I did the same in https://github.com/JuliaLang/www.julialang.org/pull/2157 lets see if it works there

Issue #2144, opened 2024-09-13
Recent blog posts shows "nothing" for the blog post content


ViralBShah:
<img width="1154" alt="Screenshot 2024-09-13 at 10 37 30 AM" src="https://github.com/user-attachments/assets/cb68e998-c8fc-4deb-8be8-ac474dc1134f">

cormullion:
I think this error was introduced in the update of July 23rd: 

https://github.com/JuliaLang/www.julialang.org/commit/9abba82e824f69dc2e231fb157c292d94deb0007#diff-77f6106b330f94b72767e939c4331f770846d06af4f97415d941b404087155daR65

where this code was added.

```julia

     # Franklin.pagevar appears to be an internal function that has a guard against
      # recursive overflow so we need to reset the counter to 0, otherwise we cannot read
      # pagevars for all pages.
      Franklin.PAGEVAR_DEPTH[] = 0
      title = pagevar(surl, :title; default="Untitled")
```

However my Franklin's a bit too rusty/basic for fixing this - perhaps we could get Thibaut to have a look...?

Also, there's an error in the footer:

![Image](https://github.com/user-attachments/assets/5495497f-2c86-42e4-80a7-262316e031ee)

which looks to have been added [here]
(https://github.com/JuliaLang/www.julialang.org/commit/c703e344a08fd55409ada99ff07f1cc9f4da04c6#diff-9fbced659a9c3a69fa276f1cc07a1f91d229c07f69980c786b5c1a2d9b5be3f5R47)

 (by @IanButterworth I think). I don't know where this is defined though.

IanButterworth:
I'll look into this.

cormullion:
> However my Franklin's a bit too rusty/basic for fixing this - perhaps we could get Thibaut to have a look...?

Perhaps this is where we’d insert `@delay` as mentioned here:

https://github.com/tlienart/Franklin.jl/issues/924#issuecomment-965736656

cormullion:
@IanButterworth Thanks! (And sorry for finding more work for you 😀)

cormullion:
![Image](https://github.com/user-attachments/assets/2147c09a-a503-49fe-88a9-386122453055)

giordano:
That's #2136

PR #2157, opened 2024-10-05
reset Franklin.PAGEVAR_DEPTH before each internal pagevar call

IanButterworth:
Investigating fixing https://github.com/JuliaLang/www.julialang.org/issues/2144

### DIFF
--- a/utils.jl
+++ b/utils.jl
@@ -67,6 +67,7 @@ function hfun_blogposts()
                 # pagevars for all pages.
                 Franklin.PAGEVAR_DEPTH[] = 0
                 title = pagevar(surl, :title; default="Untitled")
+                Franklin.PAGEVAR_DEPTH[] = 0
                 pubdate = pagevar(surl, :published)
                 if isnothing(pubdate)
                     date    = "$ys-$ms-01"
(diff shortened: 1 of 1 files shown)


IanButterworth:
I'm fixing a few issues here, but it seems that the reason that the descriptions aren't being found is a bug in Franklin where it's supposed to, but no longer checking for a `rss` field, instead only `rss_description`
https://github.com/tlienart/Franklin.jl/blob/c06db25f51fa25e97db88c5b5d3a5dff657687e9/src/manager/write_page.jl#L221-L224

Also it seems that `pagevar` is an internal expects top level calls to have reset `Franklin.PAGEVAR_DEPTH[] = 0` to guard against recursion. I don't believe that was the error here, but as noted before it was in another case, so I've set it to always be reset before a `pagevar` call. Seems adding a proper `pagevar` function to the API would be good?

fyi @tlienart

IanButterworth:
I guess this was also not generating RSS feeds.. but I have no idea if that's a thing or how it works.

IanButterworth:
I propose we merge this to fix the website, then investigate upstream Franklin fixes, then consider reverting unneeded parts of this once fixes are in.

PR #2159, opened 2024-10-07
Update index.html

dfghjkjhgr:
GTK.jl was deprecated a few months ago, I think that we ought to direct people to the up-to-date version.


ViralBShah:
Thank you. We should update much of that section as I noted in a recent issue - but good to fix this one  right away.

PR #2172, opened 2024-10-25
Add freebsd aarch64 to the supported platforms and remove macOS 32-bit

ViralBShah:
@ararslan Looks good?


ararslan:
Yep looks great! Thanks!

PR #2177, opened 2024-10-27
Fix link to package naming rules

han-ol:
Since the heading of the subsection changed from guidelines to rules, the link needed updating.


ViralBShah:
cc @DilumAluthge

PR #2179, opened 2024-10-28
Bypass aka.ms/terminal terminal link that doesn't always work

PallHaraldsson:
I get Service Unavailable, and assume the link redirected to is stable.

I see people are having issues still on Windows with terminal setup see: https://github.com/JuliaLang/juliaup/issues/528

Might juliaup recommend something on startup (for Windows users), or even download for users?

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -46,7 +46,7 @@ Julia installs all its files in a single directory. Deleting the directory where
 
 ## Windows
 
-Julia works best on 64-bit, and latest Windows 11 (you can expect terminal problems on default Windows 10), and is available for Windows 7 and later for both 32- and 64-bit versions.
+Julia is available for Windows 7 and later for both 32- and 64-bit versions.
 
 **We highly recommend running Julia using a modern terminal, such as installing the [Windows Terminal from the Microsoft Store](https://apps.microsoft.com/detail/9n0dx20hk701).**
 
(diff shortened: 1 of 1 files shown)


downloads/platform.md
```diff
@@ -46,9 +46,9 @@ Julia installs all its files in a single directory. Deleting the directory where
 
 ## Windows
 
-Julia is available for Windows 7 and later for both 32 bit and 64 bit versions.
+Julia works best on 64-bit, and latest Windows 11 (you can expect terminal problems on default Windows 10), and is available for Windows 7 and later for both 32- and 64-bit versions.
```
KristofferC, line 49: Please don't "sneak in" unrelated changes into a PR does a single thing (updates a link).

ufechner7:
I do not understand this suggestion. The problem discussed on discourse was NOT that someone used Windows 10 or the old windows command prompt, but that the variable TERM was set.

PallHaraldsson:
Feel free to copy the _first_ commit and amend as you see fit in a new PR. This PR is no longer, since second commit, about clarifying anything, just fix the link itself, since after I got (not outdated) feedback.

It should be very uncontroversial to merge as is. I don't want to spend much more time on Windows I don't even use. It's also probably better if other, that use Windows [terminal], can clarify docs related to it if needed.

PallHaraldsson:
Bump. Still getting:

Service Unavailable

HTTP Error 503.

and not with the fixed, link, now the only change so it should be uncontroversial.

PR #2189, opened 2024-11-21
"Release process" blogpost: Explain that we no longer adhere to a fixed four-month schedule for feature freeze

DilumAluthge:
Since this is a blogpost, it felt kind of weird to delete the content wholesale, so for posterity I struckthrough the old content, and then I added the new content, preceded by the phrase `Edit (November 2024)` to make clear the new nature of the content.


DilumAluthge:
Hat-tip to the Discourse user (https://discourse.julialang.org/t/julia-1-12-feature-freeze-wednesday-january-8-2025/122902/2?u=dilumaluthge) that noticed this outdated content.

DilumAluthge:
cc: @felixcremer

DilumAluthge:
Okay, I think I've found and fixed all the relevant locations.

blog/2019/08/release-process.md
```diff
@@ -37,7 +37,8 @@ This information is collected from a small set of posts on [discourse](https://d
...
 
-- Minor releases are branched every **four months**, which means that there are three minor releases per year. The rate is controlled by doing timed features freezes for minor releases: every four months, we announce on discourse that the current development version is about to feature freeze (with about two weeks notice); then on the freeze date, we branch a `release-1.3` branch for the minor release and after that no more feature development is allowed on that branch, from which the release will be tagged. More on that process below.
+- ~~Minor releases are branched every **four months**, which means that there are three minor releases per year. The rate is controlled by doing timed features freezes for minor releases: every four months, we announce on discourse that the current development version is about to feature freeze (with about two weeks notice); then on the freeze date, we branch a `release-1.3` branch for the minor release and after that no more feature development is allowed on that branch, from which the release will be tagged. More on that process below.~~
```
giordano, line 40: Is it necessary to keep the old text that was probably never applied anyway?

DilumAluthge: My thinking was that because this is a blogpost, it would be weird to delete content. So my original plan was to strikethrough the original content.

Now, I'm actually wondering if we should just preserve the original blogpost as-is, and just make a new page (either in the Julia devdocs, or somewhere in the Julia website) with an updated version of this content. And then, at the top of this blog post, just add a note that says:

> This is a historical blog post that is now outdated. The updated version of this content is maintained here: [link]

ViralBShah: I think that would be better.

ViralBShah:
I am closing this one for now.

PR #2196, opened 2024-11-25
add note about macOS x64 via rosetta not being recommended

IanButterworth:
See https://github.com/JuliaLang/julia/issues/55878#issuecomment-2498007062

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -72,10 +72,15 @@ Different OSes and architectures have varying [tiers of support](/downloads/#sup
       <th> Support Tier </th>
     </tr>
     <tr>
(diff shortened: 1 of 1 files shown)


ViralBShah (approved):

LilithHafner:
Would "tier 3" be appropriate?

ViralBShah:
More likely to be the case on windows where we do not have win arm64, and hence x86 will get run in emulation.

LilithHafner (approved):

ViralBShah:
I just realized that I have been running BinaryBuilder locally through x64 binaries on Rosetta on my mac all along (since BB requires Julia 1.7).

Issue #2054, opened 2024-02-16
Download Page: Prioritize 'Official Binaries for Manual Download' Over 'Supported Platforms'


diegozea:
**Problem:**
The current placement of the "Supported Platforms" section before "Official Binaries for Manual Download" in the Julia installation guide complicates installation for users in environments that restrict access to the Windows Store or similar services. This layout necessitates unnecessary navigation, hindering accessibility.

**Solution:**
Reorganize the installation guide to list "Official Binaries for Manual Download" before the "Supported Platforms" section. This change will make manual download options more visible and accessible, streamlining the installation process for users facing institutional restrictions.

**Additional Context:**
In many academic and corporate settings, security policies limit the use of platform-specific stores for software installation. Prioritizing manual download options can significantly improve accessibility and user experience for those in such environments.

KristofferC:
I think it is enough to also link to the MSIX installer on the Windows download page (which AFAIU should work even in this settings)?

diegozea:
I think that should also work; I could test it if you provide me a link to such an installer. For the moment, I am using the `.exe` installer: [julia-1.10.1-win64.exe](https://julialang-s3.julialang.org/bin/winnt/x64/1.10/julia-1.10.1-win64.exe)

KristofferC:
https://github.com/JuliaLang/juliaup?tab=readme-ov-file#windows


-----------

If the Windows Store is blocked on a system, we have an alternative [MSIX App Installer](https://learn.microsoft.com/en-us/windows/msix/app-installer/app-installer-file-overview) based setup. Note that this is currently experimental, please report back successes and failures https://github.com/JuliaLang/juliaup/issues/343. To use the App Installer version, download [this](https://install.julialang.org/Julia.appinstaller) file and open it by double clicking on it.

If neither the Windows Store nor the App Installer version work on your Windows system, you can also use a MSI based installer. Note that this installation methods comes with serious limitations and is generally not recommended unless no other method works. For example, there is no automatic update mechanism for Juliaup with this installation method. The 64 bit version of the MSI installer can be downloaded from [here](https://install.julialang.org/Julia-x64.msi) and the 32 bit version from [here](https://install.julialang.org/Julia-x86.msi). By default the install will be a per-user install that does not require elevation. You can also do a system install by running the following command from a shell:

-------------

diegozea:
The `.appinstaller` there fails with the following error:

```
[...]

PR #2205, opened 2024-12-04
Remove highlight bubble from downloads page

Keno:
I watched several new users attempt to install julia today. The results were very painful. First the vast majority of users completely ignored the grey bubble at the top that contains `juliaup` information. This is a well documented phenomenon (https://en.wikipedia.org/wiki/Banner_blindness), but applies here quite strikingly. Next people would usually scroll down to the support table, pattern match their platform and try clicking on things, getting frustrated when nothing was happening. Very few people scrolled down to the actual list of binaries. This is a significant usability regression over the previous downloads page that just had the list of binaries.

Try to fix things by:
1. Removing the styling of bubble at the top to signal that it is body content, not a banner.
2. Moving the manual install table up, so that people see it before the support table.

This is a quick fix to hopefully restore usability for new users. Further improvements could surely be made, but hopefully this'll help for now.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -2,8 +2,8 @@
 @def title = "Download Julia"
 
 ~~~
-  <div class="main-download-instructions">
-   <div class="main-download-instructions-inner">
+  <div>
+   <div>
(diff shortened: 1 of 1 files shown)


KristofferC:
I think this fixed https://github.com/JuliaLang/www.julialang.org/issues/2054 by reverting the reordering introduced in https://github.com/JuliaLang/www.julialang.org/pull/2001.

PR #2208, opened 2024-12-13
newsletter for 2024-11 and 2024-12

Krastanov:
getting further and further behind schedule... will try to reset for January

Will upload to discourse tomorrow


ViralBShah:
Should we merge these? @Krastanov Happy to also give you commit here to be able to merge these quickly. Please let me know.

Krastanov:
Commit privileges would be convenient, but not a high priority. Thanks for offering.

This is indeed good to merge on my end.

PR #2217, opened 2025-01-10
Add a new proposal for GSoC25 in Trixi.jl

huiyuxie:
I found a good quadtree example in the CUDA examples and have included the link here (instead of writing my own samples for guidance). However, I still plan to write a corresponding example in Julia to provide better support.

Also, I am unsure whether CUDA.jl offers robust support for handling complex tasks involving dynamic parallelism, as CUDA does. Moreover, the actual speedup depends heavily on the problem size. For example, 3D problems generally achieve greater speedup compared to 1D, and deeper levels of refinement typically result in better acceleration.

But this proposal is interesting to me; if no one chooses it, I'm happy to work on it myself.

### DIFF
--- a/jsoc/gsoc/trixi.md
+++ b/jsoc/gsoc/trixi.md
@@ -98,14 +98,15 @@ This project is good for both software engineers interested in the fields of sci
 The possible subtasks in this project include:
 
 - Implementing the abstract tree initialization process on GPUs.  
-- Exploring the `TreeMesh` and `P4estMesh` initialization processes on GPUs based on the implementation of the first task and combining them.  
+- Exploring the [`TreeMesh`](https://trixi-framework.github.io/Trixi.jl/stable/meshes/tree_mesh/) initialization processes on GPUs based on the implementation of the first task and combining them.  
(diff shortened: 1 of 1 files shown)


huiyuxie:
CC @ranocha @jlchan @sloede

huiyuxie:
Let me know if you find any problems or have any questions. Thanks!

jlchan:
It reads well overall, but I'm not sure how feasible migrating the P4est tree management code to GPU would be.

huiyuxie:
> but I'm not sure how feasible migrating the P4est tree management code to GPU would be.

Which thing are you most concerned about? I also think `P4estMesh` is more challenging than `TreeMesh`, so perhaps its order could be moved toward the end of the subtask list.

jlchan:
> Which thing are you most concerned about? I also think `P4estMesh` is more challenging than `TreeMesh`, so perhaps its order could be moved toward the end of the subtask list.

That would address my only concern!

huiyuxie:
Please review again and approve it if there is no problem. Thanks

jlchan (approved):

huiyuxie:
Could you @ViralBShah please help merge it if you are available thanks

inkydragon:
Thanks for your proposal.

Assuming `ViralBShah` is busy, I'll merge this.

huiyuxie:
Thanks 😊

PR #2218, opened 2025-01-16
Update Tiers: rosetta t3, musl t3, ppc t3, armv7 t4. Add: wsl t3

ViralBShah:


downloads/index.md
```diff
@@ -379,6 +379,10 @@ Different OSes and architectures have varying [tiers of support](/downloads/#sup
       <td> i686 (32-bit) </td>
       <td> <font color="green">Tier 1</font> </td>
     </tr>
+    <tr>
+      <td> WSL 2 </td>
```
Keno, line 383: Needs a rowspan increment above. There's also the complication that it also depends on the support tier of the linux distribution that WSL2 is running. WSL2 is just a VM + some bits.

inkydragon: > the linux distribution that WSL2 is running

Perhaps it could be clarified that we only support the default installed Ubuntu LTS.
(I think it's Ubuntu 24.04 for now)

```
PS C:\Users\inkyd> wsl --list --online

NAME                            FRIENDLY NAME
Ubuntu                          Ubuntu
Debian                          Debian GNU/Linux
kali-linux                      Kali Linux Rolling
Ubuntu-18.04                    Ubuntu 18.04 LTS
Ubuntu-20.04                    Ubuntu 20.04 LTS
Ubuntu-22.04                    Ubuntu 22.04 LTS
Ubuntu-24.04                    Ubuntu 24.04 LTS
OracleLinux_7_9                 Oracle Linux 7.9
OracleLinux_8_7                 Oracle Linux 8.7
OracleLinux_9_1                 Oracle Linux 9.1
openSUSE-Leap-15.6              openSUSE Leap 15.6
SUSE-Linux-Enterprise-15-SP5    SUSE Linux Enterprise 15 SP5
SUSE-Linux-Enterprise-15-SP6    SUSE Linux Enterprise 15 SP6
openSUSE-Tumbleweed             openSUSE Tumbleweed
```

There are many other distributions available in the windows store, for example: Arch, Alpine, ...

giordano:
While you're at it, julia on musl often fails to build, which means it should be demoted to tier 3

ViralBShah:
I see PPC binaries in both 1.10 and 1.11, so I'm making it Tier 3. I believe ARMv7 should be Tier 4 since I don't think it has been built since 1.7

downloads/index.md
```diff
@@ -395,11 +399,11 @@ Different OSes and architectures have varying [tiers of support](/downloads/#sup
...
     <tr>
       <td> ARMv7 (32-bit) </td>
-      <td> <font color="crimson">Tier 3</font> </td>
+      <td> <font color="black">Tier 4</font> </td>
     </tr>
     <tr>
-      <td> PowerPC (64-bit) </td>
-      <td> Tier 4 </td>
+      <td> PowerPC LE (64-bit) </td>
+      <td> <font color="crimson">Tier 3/font> </td>
```
inkydragon, line 406:
```suggestion
      <td> <font color="crimson">Tier 3</font> </td>
```

PR #2225, opened 2025-01-21
[gsoc] Remove tabular data GSoC projects

xlxs4:
cc @bkamins


xlxs4:
also cc @tanmaykm

ViralBShah:
@bkamins Perhaps I closed too soon before you could chime in. The original page is still there, and we could leave the DataFrames.jl GSOC open as a placeholder (doesn't probably have to be joins).

bkamins:
We discussed this on Slack. It can be closed. I do not expect much interest in this GSOC from contributors.
How I see the state of DataFrames.jl now is that it is mostly in maintenance mode. We do not receive significant requests for new features/performance improvements from the users recently.

If I were to point the need for GSOC contributions is to provide support/development of decent lightweight connectors to DB engines.

PR #2228, opened 2025-01-22
[GSoC] Project list updates

xlxs4:
I'll make a single PR to update them all in one go. Changes are as indicated through prior discussion with the respective mentors

cc:
- JSServe + Makie: @SimonDanisch (commit msg is wrong, oops)
- Taija: @pat-alt 
- QLM + Makie, language interop: @barche


xlxs4:
I've deleted a couple of old projects because we were unable to contact the mentors. If you're the mentor of one of these projects, feel free to ping me here or on Slack to get them listed again. These projects are:
- JuliaStats (@ParadaCarleton @ayushpatnaikgit and more)
- Constraint programming (@Azzaare)
- Juliareach (@mforets)
- PDMP, Kalman (@mschauer)
- MLJ (@vollmersj (?))

mforets:
[xlxs4](https://github.com/xlxs4), @TheCedarPrince  Thanks for pinging me!

Perhaps you tried contacting via Julialang's Slack, which i open very often. We have chosen to use Juilalang's Zulip for our org (see #juliareach channel). Together with cc @schillic , we have weekly dev calls openly announced in our julialang's Zulip channel.

Unless otherwise requested, we would like the projects to be added again (can do tomorrow, and also review / update all links and descriptions) 🚀

xlxs4:
@mforets I've opened a PR (https://github.com/JuliaLang/www.julialang.org/pull/2257), feel free to change it as you see fit

ParadaCarleton:
> JuliaStats (@ParadaCarleton @ayushpatnaikgit and more)

Sorry for the late response—just wanted to mention that if at all possible, I'd really suggest finding someone who'd be interested in mentoring and/or contributing. Distributions.jl is badly in need of a fork and rewrite, to the point where it's holding back the Julia ecosystem as a whole right now. My frustrations with bugs in StatsBase.jl and Distributions.jl (which can't be fixed without changes some contributors have asserted would qualify as breaking) are what ultimately led me to leave Julia and go back to Python, and they're responsible for the infamous Yuri post—11 of the 15 issues he described are from JuliaStats, and as far as I know the overall quality of the StatsBase.jl code has not improved substantially since then (or since I started tuning out of Julia about a year and a half ago).

I'm personally aware of at least 2 commercial projects I was part of that considered using Julia to build their products, but rejected it and went back to Python after running into Distributions.jl bugs.

xlxs4:
Hi @ParadaCarleton, thanks for letting us know. Unfortunately the proposals for this year have already been submitted. I am not aware of the maintainer status of JuliaStats, so I won't comment on that. I wasn't able to reach out to any of the JuliaStats previously listed mentors. It would be great if you (or any other JuliaStats maintainer) can sketch out a rough summary of what the current status is and what needs to be done moving forward. We can then take a look at how we can help get this running, e.g., by making some open calls, including some projects in the next GSoC, etc.

ajinkya-k:
@ParadaCarleton id be willing to help with this outside of GSoC. Can you point me to the right set of issues to get started?

xlxs4:
(note that we very much can, and should, make sure that we include some project ideas for next GSoC too!)

ParadaCarleton:
> @ParadaCarleton id be willing to help with this outside of GSoC. Can you point me to the right set of issues to get started?

Generally I'd suggest just looking at the directly. The longer-term structural issue with Distributions.jl is the code is all written exclusively to be compatible with `Array{Float64}`, rather than working with arbitrary types. Some maintainers argue fixing this would be a breaking change, since it leads to an unexpected change in output types (even though I'd argue this is incorrect behavior, so _not_ fixing it is "breaking").

The test suite probably needs an overhaul, given how many of these bugs have made their way into the code. I'd suggest: 
1. Using EvilArrays.jl to check for out-of-bounds access;
2. Explicitly test against as many implementations of "lots of numbers grouped together" as possible, including with autodiff—DimensionalData.jl, YAXArrays.jl, and Tables.jl might cover the biggest use cases;
3. Searching for type errors with JET.jl;
4. Property-based testing with Supposition.jl;
5. Adding LLM-written tests.

PR #2230, opened 2025-01-23
Remove JuliaGenAI GSoC projects from last year

svilupp:
Could you please review @xlxs4 ?

cc @cpfiffer for visibility


cpfiffer:
Works for me!

PR #2238, opened 2025-02-07
GSOC proposal for FastDifferentiation.jl

brianguenter:
@xlxs4 This is my proposal for GSOC.


xlxs4:
Thanks for the writeup @brianguenter! A few comments:
- Maybe add a high-level description of `FastDifferentiation`?
- As stated in the [GSoC docs](https://google.github.io/gsocguides/mentor/defining-a-project-ideas-list), it's good to not provide a fleshed-out plan on how to see the project through. I was wondering, however, if the `Expected Outcome` sections could be slightly rephrased to be a bit more specific?
- Similarly, perhaps consider rephrasing some of the text in `Recommended Skills` -- I can understand how, e.g., the first listed project requires mostly experience with the Julia programming language, but doesn't the second project depend more on the theoretical background? Reading off of "[...] as well as some theoretical work to determine which higher order terms can be reused."

LGTM other than these minor comments

brianguenter:
@xlxs4 thanks for the suggestions. I've made changes 1 and 3 you suggested. 

Not sure how to do 2 without being redundant, i.e., repeating in the Expected Outcome section what was already stated in the description of work. Do you think this is necessary? Can you point me to an example to help me understand what Google is expecting in this section? I looked at a couple of the other proposals and their Expected Outcome sections look very similar to mine, or they don't have this section at all.

xlxs4:
It's great as it is. LGTM @ViralBShah

PR #2243, opened 2025-02-09
Change Star and Support buttons to make all match

nimthenimble:
# Before 
Screenshot 2025-02-09 at 23 19 03

---

# After 
Screenshot 2025-02-09 at 23 19 54

### DIFF
--- a/_layout/navbar.html
+++ b/_layout/navbar.html
@@ -32,7 +32,7 @@
           </ul>
           <div class="navbar-action-buttons d-flex gap-4">
               <a class="btn btn-outline-dark btn-sm btn-star" href="https://github.com/JuliaLang/julia" target="_blank">⭐ Star</a>
-              <a class="btn btn-danger btn-sm btn-sponsor" href="https://github.com/sponsors/julialang" target="_blank">❤️ Sponsor</a>
+              <a class="github-button" href="https://github.com/sponsors/julialang" data-icon="octicon-heart" data-size="large" aria-label="Sponsor @julialang on GitHub">Sponsor</a>
           </div>
       </div>
   </div>
--- a/_layout/navbar.html
+++ b/_layout/navbar.html
@@ -31,7 +31,7 @@
               </li>
           </ul>
           <div class="navbar-action-buttons d-flex gap-4">
-              <a class="btn btn-outline-dark btn-sm btn-star" href="https://github.com/JuliaLang/julia" target="_blank">⭐ Star</a>
+              <a class="github-button" href="https://github.com/JuliaLang/julia" data-icon="octicon-star" data-size="large" data-show-count="true" aria-label="Star JuliaLang/julia on GitHub">Star</a>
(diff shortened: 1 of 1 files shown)


_layout/navbar.html
```diff
@@ -31,12 +31,12 @@
...
+              <a class="github-button" href="https://github.com/sponsors/julialang" data-icon="octicon-heart" data-size="large" aria-label="Sponsor @julialang on GitHub">Sponsor</a>
           </div>
       </div>
   </div>
 </nav>
 <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/dist/css/bootstrap.min.css" rel="stylesheet">
 
-<script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/dist/js/bootstrap.bundle.min.js"></script>
\ No newline at end of file
+<script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/dist/js/bootstrap.bundle.min.js"></script>
```
nsajko, line 42: Best to avoid unrelated whitespace changes.

nimthenimble: Understood

nimthenimble: Understood. What do you think about the PR?

giordano, line 42:
```suggestion
<script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.0/dist/js/bootstrap.bundle.min.js"></script>

```

cormullion:
There are now two star counters? 🤔

xlxs4:
If we remove the bottom right star button, I prefer the new look

cormullion:
I think the github logo (octocat) would be better.

_layout/navbar.html
nsajko: @nimthenimble please accept this suggestion, to avoid unrelated whitespace changes, as discussed above.

nsajko: oops now i see this suggestion actually adds a newline?

nsajko: IMO the PR diff should change lines just for changing whitespace, in any case (unless that's the only thing the PR does)

nsajko:
I can't really help with reviewing, sorry, Web isn't usually my thing.

nimthenimble:
@nsajko Who is the user who specialises the most in reviewing web PR's?

nsajko:
IMO it doesn't make sense to ping people for review right away after pushing. Give potential reviewers some time before pinging :smile:. 24 hours for example. Plenty people already commented here, so they're probably still subscribed anyway, and will get notified on your every push. Point being, have patience.

However, if there's no new review activity in a couple of days, then definitely start being annoying :wink:.

ViralBShah:
Is this ready? I have approved the workflow for CI

nimthenimble:
So is this getting approved?

cormullion:
Is it much of an improvement? 🤔

ViralBShah:
Let's merge and see how it is.

PR #2250, opened 2025-02-21
Adding Herb.jl JSoC proposal

THinnerichs:
- adding a JSoC proposal for our library Herb.jl, including two projects


xlxs4:
@TheCedarPrince merge?

PR #2255, opened 2025-02-25
Correct headings for FastDifferentiation.jl GSoC

mortenpi:
Currently, it shows up weirdly on the merged projects page -- it looks like it's part of the docs one:

image

cc @brianguenter


xlxs4:
Nice catch, thanks @mortenpi

PR #2269, opened 2025-03-18
Add juliageo GSoC projects

asinghvi17:
Not sure if this will get much visibillity this year but might as well leave it in...


xlxs4 (approved):

asinghvi17:
Definitely - I was planning on updating around November when we see how the ecosystem has evolved.  There may be other projects then as well around type stability for file IO, et cetera...

xlxs4:
@TheCedarPrince

PR #2272, opened 2025-03-21
Add tooling and graphics to all_projects

vtjnash:


vtjnash:
Preview seemed fine. Notably though, we should consider whether to delete the graphics page or add a link to it: @davidanthoff who has the only project listed there

PR #2275, opened 2025-04-01
Use juliaup ID rather than name

IanButterworth:
See https://github.com/JuliaLang/juliaup/issues/1191

cc. @davidanthoff

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -7,7 +7,7 @@
     <div class="container pt-sm-2">
       <div class="row" id="windows-instructions" style="display: none;">
         Install the latest Julia version (<a href="#current_stable_release">v{{stable_release}}</a>{{stable_release_date}}) from the <a href="https://www.microsoft.com/store/apps/9NJNWW8PVKMN">Microsoft Store</a> by running this in the command prompt:
-        <pre><code class="language-plaintext cmdprompt-block">winget install julia -s msstore</code></pre>
+        <pre><code class="language-plaintext cmdprompt-block">winget install 9NJNWW8PVKMN -s msstore</code></pre>
         <div class="install-platform-note"><span id="platform-subnote-windows">It looks like you're using Windows. </span>For Linux and MacOS instructions <a onclick="showOther()" href="javascript:void(0);">click here</a></div>
       </div>
       <div class="row" id="other-platforms-instructions" style="display: none;">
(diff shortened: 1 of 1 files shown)


Krastanov:
Without the context of https://github.com/JuliaLang/juliaup/issues/1191 , this command looks really sketchy. I would immediately distrust someone that asks me to run something that looks like this.

ajinkya-k:
Can we change the name to JuliaLang to disambiguate? Or is that not possible now?

downloads/index.md
```diff
@@ -7,7 +7,7 @@
     <div class="container pt-sm-2">
       <div class="row" id="windows-instructions" style="display: none;">
         Install the latest Julia version (<a href="#current_stable_release">v{{stable_release}}</a>{{stable_release_date}}) from the <a href="https://www.microsoft.com/store/apps/9NJNWW8PVKMN">Microsoft Store</a> by running this in the command prompt:
-        <pre><code class="language-plaintext cmdprompt-block">winget install julia -s msstore</code></pre>
+        <pre><code class="language-plaintext cmdprompt-block">winget install 9NJNWW8PVKMN -s msstore</code></pre>
```
ajinkya-k, line 10:
```suggestion
        <pre><code class="language-plaintext cmdprompt-block">winget install --name Julia --id 9NJNWW8PVKMN -e -s msstore</code></pre>
```

downloads/index.md
```diff
@@ -7,7 +7,7 @@
     <div class="container pt-sm-2">
       <div class="row" id="windows-instructions" style="display: none;">
         Install the latest Julia version (<a href="#current_stable_release">v{{stable_release}}</a>{{stable_release_date}}) from the <a href="https://www.microsoft.com/store/apps/9NJNWW8PVKMN">Microsoft Store</a> by running this in the command prompt:
-        <pre><code class="language-plaintext cmdprompt-block">winget install julia -s msstore</code></pre>
+        <pre><code class="language-plaintext cmdprompt-block">winget install --name Julia --id 9NJNWW8PVKMN -e -s msstore</code></pre>
```
giordano, line 10: Are there long version also for the -e -s options?

ajinkya-k:
Oh boy we have a bigger issue! Julia with that ID does not even show up in the german market (I changed my laptop region to germany):
image

downloads/index.md
ajinkya-k: yes `--exact` and `--source`

IanButterworth:
Ok. Better to error than install the wrong thing so we can leave this like it is.

PR #2282, opened 2025-04-13
More explicitely recommend juliaup as "the" way to install Julia

Eben60:
See discourse [here](https://discourse.julialang.org/t/about-julias-development-policy-regarding-windows/128011/17) and [here](https://discourse.julialang.org/t/julia-wont-open-after-updating-macos-julia-can-t-be-opened-because-julia-is-not-allowed/126199) . It was also my experience, a colleague would first go to manual download despite my explicit recommendation to use juliaup.


Eben60:
Mostly dealt with by https://github.com/JuliaLang/www.julialang.org/pull/2283 

Would open separate issues for Win and Mac.

PR #2300, opened 2025-04-29
Remove empty orgs

inkydragon:
Keep:
- [JuliaNLSolvers](https://github.com/JuliaNLSolvers) 1 members: Last Update 4 days ago
- [JuliaNonconvex](https://github.com/JuliaNonconvex) 1 members: Updated on Nov 22, 2024
- [juliahci](https://github.com/juliahci) 1 members: Updated on Dec 28, 2024
- [JuliaQuantum](https://github.com/JuliaQuantum) 1 members: Last Update 2023

Remove:
- [JuliaPraxis](https://github.com/JuliaPraxis) 0 members: Last Update 2019
- [JuliaActors](https://github.com/JuliaActors) 0 members: Last Update 2021


DilumAluthge:
I don't know if we necessarily need to remove those orgs.

inkydragon:
> I don't know if we necessarily need to remove those orgs.

I don't know, too.

> But I don't think we need to necessarily remove orgs with fewer than 2 public members.

So let's keep these organizations.
It would be nice if Github Action could show warnings instead of errors.

PR #2235, opened 2025-02-05
Refactor Navbar for Enhanced User Experience

jainampatel22:
### **Design Updates:** 
• A cleaner and more modern layout, featuring better alignment , improved visual hierarchy and added slight animations.

### **Changes:**
• Updated the CSS for the navbar to align with new design standards.
• Updated a hamburger menu for mobile view.
• Implemented smooth scroll transitions between sections.
• Updated color scheme for better visibility and consistency.

### **Impact:**
• Improved overall navigation experience, especially on mobile.
• Enhanced website accessibility.
• A more modern, streamlined look and feel.

### **Before:**
• **Desktop View** 
julia-pc

• **Mobile View**
juliaorg-mobile

### **After** 
•  **Desktop View**
local-pc

• **Mobile View**
local-mobile

### **Testing:**
• Verified the navbar responsiveness across different screen sizes (desktop, tablet, mobile).

PR #2307, opened 2025-05-14
Fix dark logo

IanButterworth:
Fixes another breakage from https://github.com/JuliaLang/www.julialang.org/pull/2235


IanButterworth:
confirmed to be fixed in the preview

StefanKarpinski:
Should we just revert that change? It's causing so much trouble.

IanButterworth:
🎣  https://github.com/JuliaLang/www.julialang.org/issues/2308#issuecomment-2881487625

PR #2310, opened 2025-05-15
Fix logo disappearing

### LINKED ISSUE #2308: julia logo only appears at some page width ranges
I think this also broke in https://github.com/JuliaLang/www.julialang.org/pull/2235

@ViralBShah 

https://github.com/user-attachments/assets/e16b582a-b88c-4583-8322-bc21b5ee8313

IanButterworth:
Fixes #2308 

Or that's the idea. Will check on the preview.

### DIFF
--- a/_css/app.css
+++ b/_css/app.css
@@ -25,6 +25,7 @@
   color: #ca3c32;
 }
 .navbar-brand {
+  min-width: 80px;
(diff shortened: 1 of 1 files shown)


IanButterworth:
The nav bar overhangs on the right still but this fixes the important issue

PR #2311, opened 2025-05-18
Clarify supported Windows versions

GHTaarn:
The page makes it seem like Windows 7 is supported.
This PR makes it more clear which Windows versions are supported.

https://discourse.julialang.org/t/windows-7-support/129116/3

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -46,7 +46,7 @@ Julia installs all its files in a single directory. Deleting the directory where
 
 ## Windows
 
-The following lists instructions for supported and unsupported versions of Windows. See [supported platforms](/downloads/#supported_platforms) for supported Windows versions.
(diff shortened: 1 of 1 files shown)


downloads/platform.md
```diff
@@ -46,7 +46,7 @@ Julia installs all its files in a single directory. Deleting the directory where
 
 ## Windows
 
-Julia is available for Windows 7 and later for both 32-bit and 64-bit versions.
+The following lists instructions for supported and unsupported versions of Windows. See [supported platforms](/downloads/#supported_platforms) for supported Windows versions.
```
giordano, line 49: I'm not a native English speaker, but "The following lists instructions..." sounds odd to me.  Maybe you can reword it in a less ambiguous way?

giordano:
Yes, thanks!  I feel like the instructions for Windows 7/8 could just be removed altogether, unless they somewhat apply to Windows 10 too (I have no idea!) but that can be done in a follow up PR.

PR #2319, opened 2025-06-18
Update index.md

ViralBShah:


_assets/2024-julia-user-developer-survey.pdf
giordano, line 1: @ViralBShah would it be possible to have numbers on the bars, like in previous years? As it is now this survey is fairly useless, I can't quote literally any number.

ViralBShah: Ok. I do have that - but I did the pdf export in keynote and didn't realized it got rid of the numbers. Can you add it?

PR #2330, opened 2025-07-17
Add GSoC 2025 projects announcement blog post

xlxs4:
I originally posted this on Slack. Post adapted from https://blog.rust-lang.org/2025/05/08/gsoc-2025-selected-projects/


xlxs4:
My idea was we can do an announcement post (this one), and then another one after the program has concluded. I can add a sentence or two about the midterm evaluation

PR #2332, opened 2025-07-21
link  that open store

xgdgsc:
After reading https://discourse.julialang.org/t/a-graphical-julia-installer-for-windows/128413/48?u=xgdgsc I found the old link shows a "Not found" for me. So I found this new link that opens the store app.


StefanKarpinski:
Does this link seem good, @davidanthoff?

ViralBShah:
The original link works fine for me.

ViralBShah:
It looks like the original link forwards to this new link. I think it is ok to keep the original URL as is, but if more people run into problems, we can reopen this PR.

PR #2339, opened 2025-08-12
Add month/year to this month in Julia world blog titles

ViralBShah:
cc @Krastanov


Krastanov:
sounds good! I will follow this style in the future

PR #2352, opened 2025-09-15
2025 August newsletter

Krastanov:
I will merge this pretty quickly after tests pass due to it being late.

These posts can generally be expected to be late. Let me know if you prefer that there is nonetheless some vetting period with me waiting before merging.


Krastanov:
The github action failed with an "out of disk space" error. This does not seem related to this PR. Closing and reopening to rerun.

Krastanov:
The build is still crashing. I will not attempt to investigate this today.

giordano:
> This does not seem related to this PR.

There's no failure in https://github.com/JuliaLang/www.julialang.org/actions/runs/17738059274/job/50404812240?pr=2348 (just triggered), so it does feel like this is related to this PR (although an Out of diskspace sounds a bit weird, the build shouldn't use that much storage to fill 10 GB).  ***Edit***: ah, the other PR didn't trigger the deploy step because it's from a fork, so the comparison isn't 1:1.

Krastanov:
Thanks for looking into this! I will try to rerun this from a fork and also run it locally to attempt to debug. But it would probably take about a day to get around to it.

ViralBShah:
@Krastanov You can merge these right away (once we are back to working deployments). No vetting period necessary. If there are edits required, we can always do another PR.

DilumAluthge:
@Krastanov @ViralBShah CI is green here now.

PR #2354, opened 2025-09-16
Update deploy workflow with cleanup steps

ViralBShah:
Added cleanup steps and updated checkout action version.


DilumAluthge:
https://github.com/JuliaLang/www.julialang.org/pull/2359 will do the cleanup we need

PR #2361, opened 2025-09-16
Community: Remove JuliaPraxis from the organizations list

DilumAluthge:
The `JuliaPraxis` org (https://github.com/JuliaPraxis) hasn't been active since 2019. I propose that we remove it from the org list. We can re-add it if the org becomes active again in the future.


DilumAluthge:
@JeffreySarnoff Is the JuliaPraxis org currently active?

DilumAluthge:
I'll also CC some other folks that have contributed in the past to repos in the JuliaPraxis org.

@miguelraz @mgkuhn @YongHee-Kim

JeffreySarnoff:
You can remove it.

DilumAluthge:
Thank you for confirming @JeffreySarnoff!

miguelraz:
Thanks for the ping, but yes. 

Delete it.

PR #2365, opened 2025-10-01
Animate the home page banner

IanButterworth:


giordano:
Looks cut out on mobile

Screenshot_20251001-135030

Screenshot_20251001-135021~2

IanButterworth:
Yeah, noticed that too. I guess I'm just collecting feedback currently on whether people want this.

giordano:
At least in the preview on mobile I can't quite tell what's supposed to be animated, nothing moves.

DilumAluthge:
For me on mobile, it is animated. I will note that the shapes move so slowly, so it's hard to notice at first (not saying that's a bad thing! just that you might not notice at first).

I tried on mobile on two different browsers, and it works in both.

IanButterworth:
If you're interested, try playing with these values https://github.com/JuliaLang/www.julialang.org/blob/1bc701500fe61311f96a253a15d271c9e4ec27e3/_libs/animated-banner.js#L17-L20

Then host it locally `julia --project -e "using Franklin; serve()"`

For instance this is a different aesthetic
```
  triangleSize: 50,      // Average size of triangles
  density: 100.0           // Triangles per 100x100 area
```

giordano:
If I wait longer, I start seeing triangles move slowly. I'm in the Eurostar station and the signal isn't great, maybe this was latency in downloading whatever script is needed to make things move around?

adienes:
I'm no UX expert or anything, but subjectively the background blue is a bit too blue & hazy imo. what would it look like knocked a few ticks closer to black?

DilumAluthge:
BTW, the text of the phrase "The Julia Programming Language" seems smaller in this PR (see below) compared to master. Would it be possible to go back to the larger font size?

Master:

master


This PR:

pr

IanButterworth:
I've held off doing anything here because I'm yet to hear anyone say they like this... 😂

IanButterworth:
Closing due to lack of enthusiasm

PR #2368, opened 2025-10-08
rm Intel and NASA sponsorships

Keno:
As far as I know both of these have ended.


ViralBShah:
Thinking about it, perhaps these should be moved to past sponsors?

PR #2382, opened 2025-11-08
make the download warning much more aggressive

Krastanov:
I have students that still go to this page instead of the install page. I am brainstorming how to do the warning better.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -2,7 +2,7 @@
 
 # Manual Downloads (Not recommended for typical installations)
 
-\note{`juliaup` is the proper way to install julia, not these downloads}{See the [installation instructions](/install/) on using `juliaup` to install Julia. If the official binaries do not work for you, please [file an issue in the Julia project](https://github.com/JuliaLang/julia/issues).}
+\note{juliaup is the proper way to install julia, not these downloads}{See the [installation instructions](/install/) on using `juliaup` to install Julia. If the official binaries do not work for you, please [file an issue in the Julia project](https://github.com/JuliaLang/julia/issues).}
 
 ~~~
 <h4 id=current_stable_release><a href="#current_stable_release">Current stable release: v{{stable_release}} ({{stable_release_date}})</a></h4>


ViralBShah:
Ok to have, but visually looks a bit odd. Probably needs some more vertical spacing.

Calling it a warning feels a bit strong - perhaps a note or something?

Zentrik:
The note feels rather difficult to read with the blue text on dark blue background, 
image

Krastanov:
This is the default css for the note command. I will look around for where else the note command is used, but I would like to avoid being responsible for modifying css used elsewhere on this website.

Krastanov:
A quick code search seems to show `\note` admonitions are not used anywhere else in this codebase, so I feel a bit more comfortable making modifications to how it is configured. I changed the color so that it has the same intensity as the warning admonition, but a blue hue.
image

Krastanov:
This should be ready for review now.

The warning is on purpose also added to the header. I believe this type of repetition is necessary, given how often newbie students of mine use this page even after I tell them "use juliaup to install julia"

ViralBShah:
I can see that it is just easier to download Julia directly rather than get an installer and learn how to use it.

ViralBShah:
I don't like adding it in the header though. It looks a bit ugly.

Krastanov:
In my experience a syllabus needs to have this type of repetition to actually get the attention of a distracted reader. I think of docs the same way.

Krastanov:
I used `|` instead of `:` because otherwise the whitespacing is wrong (wrong for English text, but I think the French do use whitespace on both sides)

Currently with `:` it looks wrong to me
image

ViralBShah:
Yes the whitespace is awkward but | looked more confusing. Looks like I or l.

I agree about repetition but maybe that can be achieved in a different way that is nicer looking?

If you google download Julia, this is where you are sent. And it is still a perfectly fine way to get Julia.

ViralBShah:
Maybe we don’t add it to the header for now and see how that works out.

For Rust they make it harder to find any other way to get rust that is not rustup. That would need a larger redesign.

Krastanov:
I fixed the note colon issue by reworking the way the note and warn commands work.

I changed the wording to avoid the word-for-word repetition.

Krastanov:
image

ararslan:
Saying that manual downloads are not the "proper" way to install Julia feels unnecessarily harsh/negative, implying that if you're installing Julia manually then you're doing it wrong, which isn't true. Otherwise, why show these download links at all? I use manual downloads, and I wouldn't fault anybody else for doing so as long as they know what they're doing and acknowledge that it's an unusual/"advanced" workflow.

ViralBShah:
Yes the language should perhaps be a bit better.

PR #2386, opened 2025-11-16
Add MakieOrg to organizations list

SimonDanisch:
Since it was missing.
If people are up for it, I'd also make a separate PR to update the plotting section to mention Makie more prominently and use a more up to date visualization made with Makie.

CC: @jkrumbiegel


ViralBShah:
Please do!

Also Makie should be more prominent for sure!

ViralBShah:
@SimonDanisch I would be completely fine with overhauling the visualization section of ecosystem to with a Makie example and description. Just make sure there is a Plots.jl reference in there as well. I don't think we need much more beyond Makie and Plots.

PR #2387, opened 2025-11-20
Change Windows WSL2 tier from 3 to 2

PallHaraldsson:
WSL 2 got tier 2 support recently. Does the GPU package inherit it? Oversight or intentionally tier 3? Should it also state Ubuntu? I think official support is conservative, and probably most distros just work (assuming, and not just meaning with GPUs).


ViralBShah:
There is very little testing for the Intel GPU backend, no CI. So Tier 3 is correct.

If CI support is added to oneAPI.jl and it passes, then Tier 2 would make sense.

PR #2391, opened 2025-11-22
Amend the language around manual downloads and make it more prominent

ararslan:
#2382 made the language around manually downloading the official Julia binaries fairly aggressive, but perhaps unnecessarily so. The concern that motivated that PR was that newer users were manually downloading these binaries rather than managing their installations via JuliaUp. There is nothing inherently wrong with that method, however, though it is much less convenient for the user.

To make the this warning friendlier to users who know what they're doing and have specific requirements that motivate this workflow (or otherwise just prefer it), we can soften the language a bit and make it a bit more specific, but also make the warning itself more prominent on the page by putting it above the "Manual Downloads" heading.

@Krastanov, I would love your thoughts on this approach and whether you think it strikes a reasonable balance with your needs.

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -1,8 +1,8 @@
 @def title = "Download Julia"
 
-# Manual Downloads (Not recommended for typical installations)
+\note{Prefer juliaup for typical installations}{See the [installation instructions](/install/) on using `juliaup` to install the official Julia binaries. Users who require a specific, atypical setup may manually download and install the official binaries from this page. If the official binaries do not work for you, please [file an issue in the Julia project](https://github.com/JuliaLang/julia/issues).}
 
-\note{juliaup is the proper way to install julia, not these downloads}{See the [installation instructions](/install/) on using `juliaup` to install Julia. If the official binaries do not work for you, please [file an issue in the Julia project](https://github.com/JuliaLang/julia/issues).}
+# Manual Downloads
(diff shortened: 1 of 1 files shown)


ararslan:
This is what it looks like:

Screenshot 2025-11-22 at 2 34 12 PM

Krastanov:
I do prefer the stronger / aggressive / paternalistic approach that is in the current master branch. My logic is that the type of users you describe will know how to interpret the aggressive warning, but inattentive students that do not have the best reading comprehension/attention would not notice a single warning.

However, this is more of a gut feeling -- I do not have the data to back it up. I do not believe I have the standing to unilaterally block this change and have less experience managing a large public software product, so I am happy to deffer to you or to a vote.

ViralBShah:
The problem is not with the wording, but with the fact that a `download julia` google search leads you to the downloads page. Students who do not pay attention are not going to pay attention to either wording. To be fair, I usually download first and read instructions later.

The real solution would be to rearrange the pages so that a google search for download julia leads you to the install page. 

julialang.org/downloads should have the `juliaup` installation instructions (currently install)
And the current downloads should become julialang.org/manual-downloads.

Krastanov:
Viral's argument sounds very convincing to me. I will not be able to contribute the suggested changes in the next few weeks.

PR #2398, opened 2025-11-23
Reorganize the downloads page

ViralBShah:
Reorganizing:

install -> downloads.
The old downloads -> manual downloads.

Addresses the discussion and suggested language in #2391. 

Also, try to get users to download and install without having to read any background on juliaup, which comes later on the page now.

### DIFF
--- a/downloads/platform.md
+++ b/downloads/platform.md
@@ -8,21 +8,12 @@ The instructions here refer to the latest stable release of Julia. If you are
 using the LTS version or an older version, please substitute the appropriate
 version number in the instructions.
 
-## Cross-platform installer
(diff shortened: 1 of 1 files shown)


Krastanov:
~~A minor suggestion: The wider internet might have links to /install. Is there a way to do a redirect so that it does not cause a 404?~~

My bad, the comment is already outdated.

ViralBShah:
Yes it is redirected. Also /install is relatively new. The traditional link has been /downloads.

Let’s see how this works out.

ViralBShah:
I'm merging this and we can fine tune things in newer PRs.

Krastanov:
Thank you!

Issue #2407, opened 2025-11-26
Blog post title is formatted strangely with a lot of whitespace before it


ViralBShah:
<img width="1150" height="559" alt="Image" src="https://github.com/user-attachments/assets/656b8c21-13a6-4855-b37f-07309282d2ce" />

mdrafi28:
Hey I can take this up and improve the placement of the headings to make the structure clearer. I will open a PR once its ready.

mdrafi28:
Hi @ViralBShah this change doesn’t fully resolve the issue, but it makes a small structural adjustment that improves how long headings behave without changing the existing style.
If you have a specific direction in mind for how the heading should ideally be structured, im happy to follow that. Otherwise I can propose a couple of minimal alternatives that stay within the current design language and community standards.

PR #2409, opened 2025-11-28
Fix RSS icon alignment for long blog titles

mdrafi28:
#2407 

### What this does
This adjusts the styling of blog post titles so the RSS icon remains visually aligned when titles wrap across multiple lines.

### Why
Longer blog titles previously caused the RSS icon to shift awkwardly in the layout, making the header feel unbalanced. This change improves readability without altering content or structure.

- CSS-only change
- Affects blog pages only
- No functional or behavioral impact

### DIFF
--- a/downloads/index.md
+++ b/downloads/index.md
@@ -4,16 +4,12 @@
 
 ~~~
 <div id="windows-instructions" style="display: none;">
-  Install Julia using the <a href="https://install.julialang.org/Julia.appinstaller">MSIX App Installer</a>. Alternatively, if you have access to the <a href="https://www.microsoft.com/store/apps/9NJNWW8PVKMN">Microsoft Store</a>, you can install Julia by running the following in the command prompt:<br><br>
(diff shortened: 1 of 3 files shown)


ViralBShah:
When Franklin regenerates the site, will anything in here get overwritten?

mdrafi28:
@ViralBShah the changes are limited to head.html and franklin.css which are source files. No generated output files were modified so nothing here would be overwritten when Franklin regenerates the site.

ViralBShah:
@mdrafi28 I don't think this fixed it. The title needs and date/author should be left aligned. 

Screenshot 2025-12-11 at 8 29 25 AM

mdrafi28:
Got it @ViralBShah ,this PR only addressed the RSS icon alignment. I was planning to handle the title, date alignment in a follow-up, but I wasn’t sure about the exact design direction, and the issue was closed before I could clarify. Now that I understand what’s needed, I’ll open a new PR shortly to fix the title and metadata alignment as well.

mdrafi28:
#2416  @ViralBShah could you please review this and let me know if any changes are needed?

PR #2406, opened 2025-11-26
Add code examples to the home page

IanButterworth:
Fixes https://github.com/JuliaLang/www.julialang.org/issues/2289

Just some starter code ideas from me and Claude.

https://julialang.netlify.app/previews/pr2406/

PR #2413, opened 2025-12-12
Fix julia mono downloads for code examples

IanButterworth:
And make it lazy load so it doesn't block page render.


IanButterworth:
@cormullion I think this addresses your review https://github.com/JuliaLang/www.julialang.org/pull/2406#issuecomment-3647425590 so I'll go ahead.

cormullion:
awesome - thanks!

PR #2415, opened 2025-12-13
Code examples: Tweak the first example ("Functions"), and add a new "Multiple Dispatch" example

DilumAluthge:
Preview URL: https://julialang.netlify.app/previews/PR2415/

Inspired by [this Discourse thread](https://discourse.julialang.org/t/multiple-dispatch-example-at-for-julialang-org/134537?u=dilumaluthge).

For Slide 1:

1. Change the slide title from "Multiple Dispatch" to "Functions"
2. Put the two invocations of `greet()` on separate lines (inspired by [this Discourse comment](https://discourse.julialang.org/t/multiple-dispatch-example-at-for-julialang-org/134537/2?u=dilumaluthge)).
3. Make the two methods of `greet()` more different from each other, to make it easier to quickly notice the visual difference.

For Slide 2:

1. Create Slide 2, with the title "Multiple Dispatch"
2. Add the "cats and dogs" example (from https://www.youtube.com/watch?v=kc9HwsxE1OY, timestamp 7:47)

### DIFF
--- a/_layout/carousel.html
+++ b/_layout/carousel.html
@@ -19,7 +19,7 @@
 <span class="hljs-string">"Hello, world!"</span>
 
 <span class="repl-prompt">julia> </span>greet(<span class="hljs-number">42</span>)
-<span class="hljs-string">"Nice to mee you, #42!"</span></code></pre>
+<span class="hljs-string">"Nice to meet you, #42!"</span></code></pre>
         </div>
       </div>
       <!-- Slide 2: Display Customization -->
--- a/_layout/carousel.html
+++ b/_layout/carousel.html
@@ -2,14 +2,15 @@
 <div class="code-carousel-container">
   <button class="carousel-arrow carousel-arrow-left" aria-label="Previous slide">&#8249;</button>
   <div class="code-carousel">
-      <!-- Slide 1: Functions -->
+      <!-- Slide 1: Straightforward syntax -->
+      <!-- Specifically, shows how to define and call two methods for one generic function -->
       <div class="carousel-slide active">
         <div class="code-showcase julia-blue">
           <div class="code-header">
             <span class="code-dot red"></span>
             <span class="code-dot yellow"></span>
             <span class="code-dot green"></span>
-            <span class="code-title">Functions</span>
+            <span class="code-title">Straightforward syntax</span>
           </div>
           <pre class="code-content"><code class="language-julia"><span class="repl-prompt">julia> </span>greet(x) = <span class="hljs-string">"Hello, $(x)!"</span>
 
--- a/_layout/carousel.html
+++ b/_layout/carousel.html
@@ -13,13 +13,13 @@
           </div>
           <pre class="code-content"><code class="language-julia"><span class="repl-prompt">julia> </span>greet(x) = <span class="hljs-string">"Hello, $(x)!"</span>
 
-<span class="repl-prompt">julia> </span>greet(x::<span class="hljs-built_in">Number</span>) = <span class="hljs-string">"Nice to meet you, #$(x)!"</span>
+<span class="repl-prompt">julia> </span>greet(x::<span class="hljs-built_in">Number</span>) = <span class="hljs-string">"Nice to meet you, number $(x)!"</span>
 
 <span class="repl-prompt">julia> </span>greet(<span class="hljs-string">"world"</span>)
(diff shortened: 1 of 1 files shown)


DilumAluthge:
~~Also, it was pointed out in the Discourse thread that technically this example is only single dispatch. Maybe we rename this example to simply "Dispatch", and then add another example that shows multiple dispatch?~~

EDIT: I've implemented this. Slide 1 is now titled "Functions", and I've added a Slide 2 titled "Multiple Dispatch".

IanButterworth:
Needs vertical space after the prompt returns.

DilumAluthge:
> Needs vertical space after the prompt returns.

Done

IanButterworth:
It didn't change here, but for the first one it strikes me now that `#$(x)!` could look really weird to those unfamiliar with julia syntax. Maybe change it to `"Nice to meet you, number $x"` ?

`Functions` is a bit of a plain title.. like the first thing we advertise is that julia allows you to have functions? Maybe something about the function syntax? `Straightforward syntax`?

And I'd add this to the end of the multiple dispatch one
```
julia> greet(Dog(), Cat())
"chases"
```

DilumAluthge:
> It didn't change here, but for the first one it strikes me now that `#$(x)!` could look really weird to those unfamiliar with julia syntax. Maybe change it to `"Nice to meet you, number $x"` ?

Done!

DilumAluthge:
> And I'd add this to the end of the multiple dispatch one

Done!

DilumAluthge:
> `Functions` is a bit of a plain title.. like the first thing we advertise is that julia allows you to have functions? Maybe something about the function syntax? `Straightforward syntax`?

Done. I've changed the title to "Straightforward syntax", but I'm open to bikeshedding it.

digital-carver:
"Straightforward syntax" sounds like a bit of tongue twister to me, and maybe sounds (subjectively) a bit defensive as the first thing one sees in the carousel. "Clear & Readable Syntax", "Easy & Natural Syntax", or some other combination of those adjectives, would sound better IMO. (I'm using the ampersand to match the other example titles so that they all feel coherent.)

DilumAluthge:
What if we do the following:

1. Rename the current Slide 1 from "Straightforward syntax" to "String interpolation".
2. Move the "String interpolation" slide from being Slide 1 to somewhere else (maybe even at the end of the deck).
3. Make the "Multiple Dispatch" slide (the cats and dogs example) the first slide

Zentrik:
to nitpick/bikeshed, it seems unfortunate that the default example requires scrolling down on it (not even just scrolling down on the page) to see the whole thing (or at least it does on my pc).
image
And scrolled down a bit on the page
image

DilumAluthge:
> to nitpick/bikeshed, it seems unfortunate that the default example requires scrolling down on it (not even just scrolling down on the page) to see the whole thing (or at least it does on my pc).

Yeah, it also requires scrolling for me.

Hmmm. Should we make the "terminal" taller?

IanButterworth:
Whatever the first one is called should be defendable as "the first thing we want to show off is our great X"

"The first thing we want to show off is our great string interpolation" doesn't make sense IMO.

Regarding the layout, I saw it as a pro that it's a little off screen. Promptes scrolling. Design by committee is hard.

DilumAluthge:
I think "multiple dispatch" is a good choice for "first thing we show off", but I think the dog/cat example is better for showing that.

jakobnissen:
Bikeshed, but I don't think multiple dispatch is a great selling point for Julia. I'm myself of the opinion that it's wildly overrated, but even to those who love it, its advantages are not visible in a small example. People who haven't already bought in don't know why they should like MD.

Instead, I'd have a cool unicode + math + broadcasting example to attract the Matlab crowd, or else a multithreading or a "performance" example (low-level optimisation or a performance-critical for loop)

ViralBShah:
I would generally agree that people come to Julia for things like performance, ease of use, package ecosystem etc. Multiple dispatch is what they discover and learn (and perhaps even love), but is not why they come. 

I think it is good to have an example for multiple dispatch in the rotation, but we could definitely have more end-user exciting features in the first 2-3 examples in the carousel.

DilumAluthge:
After #2507, the first example on the home page ("Multiple Dispatch") does indeed show multiple dispatch (as opposed to single dispatch).

So I think the main goal of this PR has landed.

PR #2421, opened 2026-01-09
This Month in Julia World Dec 2025 (and fixes for Nov)

Krastanov:
I have not proofread this yet. I will try to do so tomorrow and merge if everything looks good.


ViralBShah:
I really like the formatting in discourse, where after each bullet, there is some vertical space. The blog post feels like a wall of post.

Krastanov:
The markdown is the same in the two locations, so presumably it is a difference in the parser/renderer (which is probably not easy to change) and in the CSS (which is probably what can be edited easily).

It seems that deleting  `.container ul li p {margin-bottom: 0;}` is enough to fix this difference.

Krastanov:
right here: https://github.com/JuliaLang/www.julialang.org/blob/8c61401402b06d85f980032b4ae1e0ded470a928/_layout/head.html#L19

But I do not know why this was added in the first place.

ViralBShah:
Do you or someone here know how to fix that?

Krastanov:
Deleting that particular line of css should be enough. But someone presumably added it 6 years ago for a reason. I would not be able to take care of reviewing (manually or with a script) all the rest of the website for location that might be broken after such a change, at least not in the near term

ViralBShah:
I'll give it a shot when I find some time.

PR #2430, opened 2026-01-20
Update recommended installation method to include download as well as `juliaup`

aplavin:
**upd**: this issue was renamed, twisting its original intent; read the thread for the full picture.

---

During a recent [discussion](https://discourse.julialang.org/t/why-is-juliaup-emphasized-as-the-installation-mechanism), I crafted an arguably cleaner installation page design – thanks to LLMs for quick proof of concept. Seems like it got quite some support in that discussion thread, so let's try here :)

image

It covers both "Julia code user" and "Julia code writer" usecases with a single page. IMO it is easier to grasp for a new user than either of the existing installation pages ([1](https://julialang.org/downloads/) or [2](https://julialang.org/downloads/manual-downloads/)).

This is a very rough draft, and the exact styling can be tweaked of course.


DilumAluthge:
Let me see if I can get some people to chime in here.

IanButterworth:
Why is the complication necessary/beneficial? It would be good to motivate and explain the changes here in the PR itself.

vchuravy:
From an educator perspective the move to `juliaup` has vastly improved the experience of students and simplified teaching. It used to be the case that during workshops or when teaching classes I kept running into old versions of Julia that required significant support and distracted from the actual teaching. So I strongly disagree with the characterization of `juliaup` being for developer, for me the use case is exactly the opposite.

adienes:
I also strongly disagree with this characterization of juliaup not being intended for beginners. I would recommend raising any issues with juliaup in its repo rather than remove its recommended use from julialang.org

fingolfin:
Before `juliaup` I had lots of people whom i had to help with trouble installing Julia. And afterwards they never updated and were using two year old Julia versions etc.

Everything got a lot better for me and the people I've helped to get going after I switched to teaching them to use `juliaup`. This is also what we tell people on our "install" page, were "step 1" is to install Julia. We got complaints before, but we've got zero since the switch to Juliaup.

So I am also not convinced this is a move in the right direction. I understand some people are unhappy with `juliaup`. But I also know a lot are *very* happy.

If there was actual data instead of anecdotal evidence (a few people discussing on a discourse thread), that would be different. Perhaps a user survey could be held? Though it is a bit difficult to get meaningful data that way, as many people will just install it once and be done, no matter which method is used.

aplavin:
Just think of the situation from the PoV of a user of a Julia code (somebody who doesn't write Julia nor intends to).
They:
- need a specific Julia version (eg, 1.10) now
   - and only need to update it if (i) the relevant codebase gets updated and requires a more recent Julia *and* (ii) they continue using the code for an extended period of time
- prefer familiar installation GUI, at least on windows/macos

I think (also evidenced by others in that thread) that the current standalone installer fits this usecase perfectly well – it's exactly what one gets with the standalone installer! Download the Julia version you are told, do the typical installation process for your OS, and that's it.

Compare to Juliaup:
- open terminal, paste some command line incantation, answer some questions in the terminal environment as well
- install the Julia version you actually need (also needs terminal)
- remove whatever Julia version happens to be installed by default with Juliaup so that not to waste space (also needs terminal)

How's that a benefit? I've always directed potential users of my codes to standalone installers, and don't really see any reason to stop doing that. Actually was surprised when that thread started and I looked at the installation page – to see the standalone installers so hard to find for a newbie (who needs them the most).

aplavin:
And I'm speaking as a juliaup user myself. It's similar to rustup, and is indeed quite convenient for those who _write_ Julia code or intend to.

aplavin:
To address some potential confusions from comments above:

> From an educator perspective the move to juliaup has vastly improved the experience of students and simplified teaching.

Sounds like your students do _write_ Julia code? Then the installation page from this PR continues to recommend juliaup for them, no change whatsoever.

> I also strongly disagree with this characterization of juliaup not being intended for beginners. 

Same for me! That's why the proposed install page doesn't mention "beginners" at all. The distinction is not "beginners" vs "experienced", but "write Julia" vs "only use existing Julia code".

IanButterworth:
Maybe it just needs this? https://github.com/JuliaLang/www.julialang.org/pull/2433 i.e. https://julialang.netlify.app/previews/pr2433/downloads/

ViralBShah:
I can't find the discussion now, but we went through a whole discussion about how students kept downloading the direct binaries and running into trouble, and `juliaup` would be the preferred installation method for teaching.

I do see the value of a standalone installer for people who want a standard install experience on Windows/mac. I am not sure I agree with the separation of the use cases, since it can be quite subjective.

I like @IanButterworth's suggestion to make manual downloads a bit more visible, and it could even be updated to say something about a traditional installer experience.

aplavin:
I don't care about specific framing, just please consider covering this usecase as well – one that arguably should become more common with general Julia adoption:
- Person wants to run a specific Julia codebase, that naturally requires specific Julia 1.X
- That person goes to the Julia website, clicks "download" as they are used to everywhere else...
- ... and there's no link to actually download anything! To even get to a download link is quite challenging, unless one is directed by someone more experienced.

Is there any other language that (i) needs to be installed by users, not just developers and (ii) makes downloading a specific version so complicated?

---

I'm very surprised to see this question of actually providing all-familiar download links at the download page so controversial. I guess if this doesn't happen I'll have to provide direct links to Julia installers in instructions, instead of simply saying "install Julia, version 1.10"... Would clearly be suboptimal, but still better than expecting everyone to find the actual downloads page themselves despite all the effort julialang website does to actively discourage this.

adienes:
> * Person wants to run a specific Julia codebase, that naturally requires specific Julia 1.X

I disagree with this assertion.

> * Is there any other language that (i) needs to be installed by users, not just developers 

all of them? I guess maybe I still fail to understand the use case you have in mind? There aren't exactly many video games written in Julia.

> * (ii) makes downloading a specific version so complicated?
```
curl -fsSL https://install.julialang.org | sh

juliaup add 1.11
```

couldn't possibly be simpler imo

giordano:
> ```
> curl -fsSL https://install.julialang.org | sh
> juliaup add 1.11
> ```
> 
> couldn't possibly be simpler imo

Yes, it can
```
curl -fsSL https://install.julialang.org | sh -s -- --default-channel=1.11
```
if you want to install v1.11 _only_ and have that as a default

IanButterworth:
I think this might boil down to serving people who aren't comfortable using the command line.

I thought about making my PR "if you're not comfortable using the command line" but that seemed patronizing.

adienes:
but how can you use julia, or anything written in julia, without the command line? are there really any popular pure GUI apps that require julia to be installed, but don't interact with command line at all?

IanButterworth:
The vscode extension + notebooks integration is pretty terminal-light and handles julia installation for you. (I think that's a new feature)

If you come from matlab that kind of terminal-less usage is roughly most familiar.

tecosaur:
I do think there's a worthwhile argument to be had that `uv` is the model we should emulate for "start using Julia" rather than Rustup, but I suspect this discussion is best had in another forum.

StefanKarpinski:
I'm against this change. Would be in favor of adding graphical methods to install juliaup. For every person who has complained about juliaup, there have been dozens who have said that it works great for them and is much easier. The way to improve this situation is to address problems that people have encountered with juliaup, not to fracture the recommended installation mechanism.

aplavin:
Idk where it comes from exactly, but there's clearly a big disconnect between different crowds. A bunch of the arguments given here are already discussed at the linked discourse thread, don't want to rehash them yet again.

I really don't understand how putting actual links to download julia installer to the downloads page at the official website is controversial.
The gui installers already exist, they behave the same way as other software installers on windows/macos, and they are even linked from the website – but these links are super hard to find for a person new to julia who just wants to run an existing code. The website seems to actively discourage using all-familiar installation mechanism that every other piece of software uses, and people are familiar with!

Also, don't appreciate forced rename that twists the intended purpose of this PR.

ViralBShah:
@IanButterworth 's suggestion in https://github.com/JuliaLang/www.julialang.org/pull/2433 to make manual downloads more prominent solves the immediate problem. I made some small edits to it and will merge it soon.

@aplavin I do like your suggestion here. I do not agree with the classification of users - but just presenting the two methods without trying to say who should use which is something I'll experiment with.

rayegun:
If possible the graphical installers should just use juliaup behind the scenes. I second Valentin's observation that the use of `juliaup` has *dramatically* simplified the installation process for beginners, notebook users, students, etc. I have spent quite some time repeatedly helping very senior members of the community figure out how to deal with versioning issues with the graphical installer versions of Julia on macOS.

Making the graphical installers more visible like in #2433 is important, but recommending them as the default for "new users" is a huge step down. This is *not* just an issue for version upgrades. When we're teaching using something like Enzyme, CUDA, etc the working Julia version is (unfortunately) a moving target. Dealing with these compatibility issues and upgrading and downgrading Julia is a huge headache if "new user" students go to the installation page and it recommends the graphical installers.

`juliaup` could also add icons to the applications folder perhaps.

aplavin:
> suggestion in https://github.com/JuliaLang/www.julialang.org/pull/2433 to make manual downloads more prominent solves the immediate problem

Does it help much?.. Just a few things noticed from a quick look:
- Downloads page still doesn't have any download links
- says "If you prefer to use manual installation <...>" at the download page – but how is a familiar-style GUI installer more manual than julia-specific command line installer?
- the so-called "manual downloads" page remains with the same title (again, how it's more manual?), and the language used at the top reads as heavily discouraging – new users can read that and end up digging into the terminal two-step installation (juliaup -> julia) instead of the standalone installation typical for other software.

Also yes, I think mine also looks nicer :) And why have two separate downloads pages at all, if one can show all the key info without being overcrowded?

> I do not agree with the classification of users 

Interesting :) In which direction? As someone who writes Julia a lot, I do enjoy using juliaup. But at the same time, cannot imagine juliaup being superior for someone who fits all three conditions on the left (doesn't write julia; needs single version; not familiar with cli).

aplavin:
At least one piece of confusion persisting over multiple people in this thread (eg @rayegun @adienes @vchuravy @StefanKarpinski) seems to be the impression that this PR changes the recommended installation method for beginners in Julia.
While that is 100% _not_ the case (and that's why I don't appreciate force-renaming this PR to twist its purpose). I agree that for those _writing_ Julia code (be it packages or notebooks or whatever), juliaup is generally better. But that's only a part of Julia users – probably, a large part currently, but still a part.

While in fact the main change is to streamline the downloads page so that existing gui installers are possible to find at all for new users (new to Julia, not intending to _write_ any code).

ViralBShah:
cc @Krastanov

StefanKarpinski:
Perhaps it's not what you intended, but putting the standalone installers first and labeling them "for users" very much sends the message that those are the recommended way to install Julia. Many Julia users who _write programs_ do not describe themselves as "developers" or "programmers"—the number of times I've had people who write programs on a nearly daily basis tell me "I'm not a programmer" is substantial. The term "developer" has even more connotation of being a professional programmer. Without reading the fine print, I would assume that "for users" means for people who want to write Julia and "for developers" is for people who want to develop Julia itself.

The first thing on the page should be the installation option that is appropriate for the most people who go to that page. Do you really think that the majority of people going to julialang.org and navigating to the install page are not going to be writing Julia code? Sure, there are some people who just want to run a Julia program that someone else has sent them, but I really doubt that it's a majority. If the two panes were swapped and titled differently, I could maybe get on board with this, but as it stands, it _does_ effectively change the recommended installation method, and I cannot support it.

StefanKarpinski:
Also the page is called "Install" not "Downloads" so that whole angle of argument seems somewhat vacuous to me. The URL is still /download/ so as not to break old links, but I will happily change it around so that /download/ redirects to /install/ instead of the other way around.

ViralBShah:
The /install and /downloads thing is my creation and has been on my mind to swap at some point. It was to try and make sure Google picked the right page, since downloads had much more pagerank back then and would automatically send people to the old downloads page.

If we were to do the panes, I would definitely want to swap them.

tpapp:
From the discussion, I think that the main gripe is not about `juliaup` *per se*, but that it is not a GUI.

"Users" in the proposed draft really refers to "people who don't want to use the command line for some reason". I don't think there are a lot of people who set up Julia and then don't write a single line of code, so I am not sure that distinction is very useful.

I think that the long-run solution is improving `juliaup`, eg by providing some sort of GUI on Windows at least. It should remain the recommended installer.

Directing the inexperienced users to binaries does them a disservice, making it more difficult for them to get updates, or even bugfixes.

ViralBShah:
The GUI installer is now much more prominently mentioned.

If someone wants to vibecode a juliaup GUI installer, that will be a fun project!

DilumAluthge:
Yes, I do think a GUI installer for Juliaup would go a long way.

ViralBShah:
Other discussion aside, the one thing I do not like about this PR is that it is not easy to edit (except for text replacements). There's a lot of CSS/html to make it look nice. Maybe it will meaningfully only be editable through claude.

aplavin:
> The GUI installer is now much more prominently mentioned.

Now it's indeed easier to find the link to actual installers on the "Downloads" page!

Still, the path for a new user (who just wants a specific Julia version to run someones code) looks like:
- go to julia website, click "Download" on top
- hmm, no links to actually download anything, but at least after careful reading there's a link to "Manual Downloads" (do I really want "manual"? but there are no download links on this page anyway, so let's click there)
- now I finally see a bunch of download links! (would be best if they focused on the ones for my system, but ok); I guess I'll just ignore the prominent warning "Users who are unable to use juliaup or require a specific, atypical setup may manually download and install the binaries from this page" because this is the only place with download links anyways

Doesn't seem very friendly, to be honest...

ViralBShah:
No disagreement.

aplavin:
And making that path smoother was exactly the sole point of this PR – have main installation options on one page, with short clear instructions which one to use in different scenarios. Unfortunately, this was interpreted differently (and even renamed to twist the original purpose) – but maybe someday Julia installation flow would become smooth for a non-programmer user, I'm not losing hope...

StefanKarpinski:
> Yes, I do think a GUI installer for Juliaup would go a long way.

Just to be clear, I think that just a GUI installer for juliaup is not quite what we should have: there should be simple but usable JuliaUp apps for Windows and MacOS that install and manage juliaup and provide an interface to some of the more common commands, e.g. "self update", "update", "list", "status". And probably start a (configurable) terminal app with a specific version of Julia running inside. That seems like the basic functionality to me. Ideally, we would also ship this app in the Windows Store and on the MacOS Store and those would keep them up-to-date. The install instructions could then be: "Click here to install the JuliaUp GUI app. If you prefer a pure command-line experience, run this command: ..."

IanButterworth:
Sounds good. Existing discussion: https://github.com/JuliaLang/juliaup/issues/54

davidanthoff:
I think for the scenario of a user who just wants to run some Julia code (i.e. the original intent of this PR on the left side in the top screenshot), the best workflow would actually to double down on https://github.com/JuliaLang/juliaup/pull/1315 and leave the current install recommendation intact, maybe with some minor tweak that Juliaup does delay installing the `release` channel until it is explicitly requested for the first time. And probably also with a tweak to find a secure way to make `JULIA_PROJECT=@.` the default.

So in practice then things would work like this:
- User wants to run some Julia code they got from someone, they just follow the instructions to install Julia on the website, done.
- User gets some code from someone that they want to run, so they extract that into a local folder.
- Presumably users would still run that from the command line? So they go to that folder, and they run `julia scriptname.jl` (or maybe they launch via a double click on the filename)
- The Juliaup launcher finds the project/manifest combo in that folder, and then it presumably should show a security opt-in dialog ("The code you are trying to run requires download of Julia version X and a number of packages that you currently have not installed. Do you want to download and install all these requirements? Make sure you trust everything. (y/n)"
- If the user opts in, Juliaup installs the required Julia version, and then instantiates the project, and then launches `scriptname.jl`

In my mind that is as user-friendly as it gets for someone who just wants to run some code. And we wouldn't have a bifurcation of how Julia ends up installed on a user system.

I think pointing users that just want to run code to the version-specific installers is really not a good idea, as that entire model completely breaks as soon as a user wants to use even just two pieces of Julia code that require different Julia versions, and resolving that situation is really tricky, _especially_ for occasional users.

aplavin:
I understand that tweaking the website is a small local change, and probably this rephrase is not optimal long-term (better would probably be making juliaup to use user-familiar full-GUI installation mechanisms).
However, minimality is really the reason I proposed this tweak, with the intent to get _some_ improvement for users quickly. Another half a year later – the situation is still confusing for users who just want to run julia code given to them:
- go to https://julialang.org
- click "download" – still no download links anywhere
- one needs to find "If you prefer to use manual installation <...>" (hmmm do I really prefer manual installation?.. is that even more "manual" than running some commands in the terminal as highlighted in the download page?..)
- minor, but the "manual downloads" (actually, this should just be "download" page, because only it has actual download links) could use some streamlining for the common case (see left half of the screenshot in first post)

ViralBShah:
@IanButterworth Is the JuliaUp GUI close to being ready for release? If so, I feel that is the simplest solution.

If not, I don't mind exploring this PR a bit more.

ViralBShah:
I have reworded this PR slightly and finished some of the TODO items. I am comfortable merging this until we have a GUI version of `juliaup`. Check the PR preview in the CI to check out the preview/

adienes:
I don't think any of the original concerns have actually changed. https://github.com/JuliaLang/www.julialang.org/pull/2430#issuecomment-3829438226 https://github.com/JuliaLang/www.julialang.org/pull/2430#issuecomment-3829485012
https://github.com/JuliaLang/www.julialang.org/pull/2430#issuecomment-3848832115

this PR would still frame the GUI installer as "for users" and "the simple case," and frames `juliaup` as "for developers" and "the complicated case."

It's not my understanding that this is the consensus desired summary of the two options, so I remain opposed to merging this PR.

ViralBShah:
Yes, I was considering that as I edited it. What if I remove the "users" and "developers" - the rest is just facts.

adienes:
personally speaking, I would phrase the two as something more like

> For users who prefer a GUI and don't intend to frequently update their version of Julia

&

> For users who prefer command line tools or want more flexibility to update their version of Julia

and then swap the two boxes so the juliaup one is on the left and GUI installer on the right.

davidanthoff:
I also don't think this should be merged, and I think at some level the right strategy for this PR is to ack that most people that chimed in seemed to not be in favor, and so might be best to just not go forward with this instead of trying to find some compromise.

I _do_ very much think we should think hard about specific scenarios of users and how we can make their experience better around Juliaup, so I'm not saying that the concerns @aplavin brough up are not there. I just think we should address them by improving Juliaup and keep steering folks towards it.

Here are some random thoughts around that:
- On Windows it actually doesn't have to be a command line prompt to install Julia. We could put a link to the Windows Store listing in there instead.
- We could maybe also offer small installers (scripts, really) that one can download that are version specific. For example there could be a download link "Download Julia 1.11", and it downloads a small script that launches the installer for Juliaup with the default channel selector for 1.11, or something like that. I think that in particular would probably entirely address the scenario that @aplavin had at the beginning, without fracturing our setup story?
- I also still think it would be useful to get a sense of what the scenario is where using command line stuff is not in the picture?

ViralBShah:
> * I also still think it would be useful to get a sense of what the scenario is where using command line stuff is not in the picture?

Just repeating what was discussed above, I believe there are plenty of non-sophisticated users as the OP pointed out that are simply not familiar or comfortable with the command line and terminal (usually on Windows, but also Mac). So having a GUI install option for such folks is valuable (and the best case would naturally be a GUI that is a frontend to juliaup).

tpapp:
Personally I am puzzled by the claims about the large intersection of 

1. people who want to use a complex, high-performance, AOT-compiled, multimethods language, designed primarily for scientific computing (like Julia), that **has a REPL**, which is conceptually like a command line, and is used for various tasks (including package management),

2. people who are not comfortable enough with the command line to open one and copy-paste a single command.

davidanthoff:
> not familiar or comfortable with the command line and terminal (usually on Windows, but also Mac)

So, on Windows that is _very_ easily solved, we can just replace the instructions to install via `winget` with a direct link that opens `Julia` in the Windows Store, and that is as non-commandliny as it gets. Once we merge @IanButterworth's PR with the GUI that would give a fully non-command line option on Windows. We probably should also at some point at tighter integration between Juliaup and the VS Code extension, so that there would be another non command line way to install Julia versions via Juliaup.

On Mac, it is just not entirely clear to me what a good non-command line setup method is. Download one of these files that one then drag-and-drops into the app folder? We could of course package Juliaup like that (in addition to all the other options we already have), but it is not clear to me how Juliaup itself then would update, as MacOS doesn't seem to have any mechanism for that? Of course, in an ideal world we would get Juliaup into the Mac app store (listed as just `Julia`), but I think that is too

davidanthoff:
Just asked the models what options there are on MacOS: https://github.com/JuliaLang/juliaup/issues/1540.

aplavin:
> Personally I am puzzled by the claims about the large intersection of <...>

I tried to explain it multiple times already in this thread, maybe I'm not using the right words?..
Say I have developed an app in Julia, and want to share with colleagues. They couldn't care less what it is written in, and whether that language has a REPL.

The perfect solution for them would be:
- download a familiar installer (regular "installation wizard" for windows, .app/.dmg on macos) for "Julia 1.10" or whatver version I tell them
- install in a familiar way, like every other software is installed (& have it deletable in the same manner)
- double-click my app (.jl script) and have it opened

This PR's tweak to the "downloads" page was intended to solve the first two of these bullet points, that's it. VSCode and other developer-focused stuff dragged into this thread is completely unrelated here.
File association would also be nice, but that's beyond simple website tweaks.

---

Here I describe one specific usecase, just one example – hoping to make it clear to others.

adienes:
I think that is still not so specific

> Say I have developed an app in Julia, 

what app? again, there are really few (are there even any..?) GUI-only apps written in Julia. and why can't these apps bundle their own Julia in that case.

aplavin:
> people who are not comfortable enough with the command line to open one and copy-paste a single command.

It's quite a bit more than that actually!
- single command just installs juliaup; then you need to dig into it and understand that you need "juliaup install 1.10" and set as default for convenience
- after you no longer need julia: no obvious way to uninstall everything, need to manually find and remove ~/.julia/

aplavin:
> what app? again, there are really few (are there even any..?) GUI-only apps written in Julia

There is no GUI-only way to install Julia, of course there are barely any GUI-only apps written in it 😁
I have a bunch of specialized apps that can easily be fully-GUI (https://plav.in/astrotools), but I never bothered for the sole reason that there's no user-convenient way to install Julia in the first place. Currently, they are started from the REPL, as there is no double-click or any other convenience features in Julia.

> and why can't these apps bundle their own Julia in that case.

Instead of lightweight source-code-only distributions/downloads, I'm expected ship a whole Julia installation manually? That's a whole new level of hassle.

aplavin:
Anyway, this discussion seems to be going in the same circles as half a year ago.
Hopefully, some day we'll have a user-friendly way to install Julia runtime, and "downloads" page would actually have a download link...

davidanthoff:
@aplavin, I think we should try very hard to support the scenario you describe, but I think we can support that with Juliaup, and I think we can support it in a way that is way more robust and easier for users than pointing them to the legacy installers. So I just want to make clear that I'm not disagreeing with your goal or scenario at all, I think it is important!

> - download a familiar installer (regular "installation wizard" for windows, .app/.dmg on macos) for "Julia 1.10" or whatver version I tell them
> - install in a familiar way, like every other software is installed (& have it deletable in the same manner)

I think on Windows we are there already (at least if we replace the command line command instructions with a link to the store), there is nothing more "normal" than installing something from the Windows Store, and then there is also a regular uninstall that removes the start menu entry and removes things from the `PATH`. It does _not_ remove the `.julia` folder, but neither does the legacy installer.

I think on MacOS https://github.com/JuliaLang/juliaup/issues/1540 would get us to the same place?

> double-click my app (.jl script) and have it opened

I think this scenario _especially_ speaks against the legacy installers. You can only have one version of Julia associated with .jl files or on the `PATH` with the legacy installers. So now if someone wants to say run your apps, they have to _global_ configure their system to only run Julia 1.11 properly. What if the app inspires them to want to use Julia for something else (in which case they ideally would be using the current version or LTS, not something unsupported like 1.11), or if they want to install another app that requires a different Julia version? Now everything gets really messy and complicated for them. And at least in a teaching scenario it was always these kind of students that were a major problem: they had at some point tried something Julia, and now their system was in a state where it was really complicated to get it into a workable, up-to-date situation.

On the flipside, I think with Juliaup (and specifically the launcher) the scenario could work really well especially once we make the Julia version selection via manifest the default: you double click the .jl file, the launcher looks at the manifest to figure out which exact Julia versions is needed, a UI prompt says "To run this script you need to install some additional stuff, do you agree?", you say yes, the correct Julia versions gets auto-installed, then the project gets auto-instantiated, and everything just works. If you later install another app that requires a different Julia version, everything still works (unlike with the legacy installers, where even something like that would break everything). In my mind the mental model here is that instantiating a model now includes downloading the right packages _and_ Julia version. From a user's point of view they are just some stuff that is needed for this specific app.

aplavin:
I'm not against other longer-term solutions, just want to get some improvements already – quoting my comment above:

> I understand that tweaking the website is a small local change, and probably this rephrase is not optimal long-term (better would probably be making juliaup to use user-familiar full-GUI installation mechanisms).
However, minimality is really the reason I proposed this tweak, with the intent to get some improvement for users quickly. Another half a year later – the situation is still confusing for users who just want to run julia code given to them

---

> I think this scenario especially speaks against the legacy installers. You can only have one version of Julia associated with .jl files

I see literally zero issues here. The approach many people are familiar with is right-click -> open with -> julia 1.10 or whatever other julia version they have. I remember python (some popular distribution of it?..) working like that back when I was learning it on windows.

PR #2541, opened 2026-06-15
make a Julia 1.10 release notes section more idiomatic

d-netto:
This PR enhances a Julia 1.10 release note that I wrote a while back by making it a bit more idiomatic.


d-netto:
@KristofferC: is it OK to retroactively fix release notes like this?

d-netto:
CC: @vchuravy, who might also know this.

d-netto:
Bump. Will merge tomorrow if there are no objections.

PR #2567, opened 2026-08-12
downloads: publish the new KMS release signing key

KristofferC:
Julia release binaries are now GPG-signed in CI by a key held in AWS KMS (fingerprint `64B779A570972FFF7BFC2B54EAD471E1A1F2C10A`) instead of the legacy juliareleases key — starting with v1.13.0-rc2, whose binaries are already live with signatures that do not verify against the currently published key.

Since the legacy key was compromised, this **replaces** the published key at `/assets/juliareleases.asc` with the new public key only (rather than appending), and the downloads page now notes that the old key (fingerprint `3673DF529D9049477F76B37566E3C7DC03D6E495`) should no longer be trusted.

The key material matches `signing-pubkeys/tarball_signing.pub.asc` in JuliaCI/julia-buildkite (see JuliaCI/julia-buildkite#592 for the release-pipeline context). Verified locally: the new keyring validates the v1.13.0-rc2 Linux tarball signature.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


KristofferC:
@Keno boop

Keno:
I don't think we should append. The old key was compromised, so we should revoke it.

KristofferC:
Updated

Note:

> a true GPG revocation requires a revocation certificate generated with the old private key, and none is published anywhere I could find (including JuliaCI/julia-buildkite). Removing the key from the website stops distributing it, but anyone who already imported it will still trust it. If the key really was compromised, whoever holds (or held) the old private key should generate a revocation certificate and publish it to keyservers

downloads/manual-downloads.md
```diff
@@ -369,7 +369,7 @@ Please do not use Julia shipped by Linux or BSD package managers. Many Linux/BSD
...
 
-**Download verification** — All Julia binary releases are cryptographically secured using the traditional methods on each platform. macOS and Windows releases are codesigned with certificates that are verified by the operating system during installation. Linux and source tarballs are signed with GPG using [this key](/assets/juliareleases.asc).
+**Download verification** — All Julia binary releases are cryptographically secured using the traditional methods on each platform. macOS and Windows releases are codesigned with certificates that are verified by the operating system during installation. Linux and source tarballs are signed with GPG using [this key](/assets/juliareleases.asc). As of August 2026 (starting with v1.13.0-rc2), releases are signed in CI by a key held in a cloud KMS, fingerprint `64B779A570972FFF7BFC2B54EAD471E1A1F2C10A`. The previous signing key (fingerprint `3673DF529D9049477F76B37566E3C7DC03D6E495`) has been retired and should no longer be trusted; signatures made with it should not be relied upon.
```
DilumAluthge:
```suggestion
**Download verification** — All Julia binary releases are cryptographically secured using the traditional methods on each platform. macOS and Windows releases are codesigned with certificates that are verified by the operating system during installation. Linux and source tarballs are signed with GPG using [this key](/assets/juliareleases.asc). As of August 2026 (starting with v1.13.0-rc2), releases are signed in the key with fingerprint `64B779A570972FFF7BFC2B54EAD471E1A1F2C10A`. The previous signing key (fingerprint `3673DF529D9049477F76B37566E3C7DC03D6E495`) has been retired and should no longer be trusted; signatures made with it should not be relied upon.
```

KristofferC: still "signed by the key", or?

KristofferC: Applied

PR #2594, opened 2026-09-11
Add new courses to active university courses section

LUISSMR:
I have added three courses in which I have actively using Julia. Soon I will create the GitHub space to upload all contents.


LUISSMR:
I'd like to add those courses and university (Academy) to the list of sites where Julia is used. I look forward to hearing from you.

